Issue #28, opened 2020-03-21
Add Helm source


olblak:
It would be nice to update Helm version when needed

PR #47, opened 2020-05-27
Add helmchart source&condition

olblak:
This pull request allows to specify helm chart that needs to be automatically updated when a version is available
Solve #28


jetersen:
This is not released yet 😢

PR #54, opened 2020-06-29
make title and message pretty

jetersen:
This works out great.
Here you got the updated commit message:
```
[updatecli] Update external-dns version to 2.20.4

Key 'versions.externalDns', from file 'system-services/helm/values.yaml', was updated to '2.20.4'
```

Not sure making name required is desired or not but for my use-case it works.

I was thinking of taking the name from source if none is supplied but it became cumbersome changeset.
for maven you would use `artifactId` and github release you would use `repository`
for docker you would use `imageName` and helm chart you would use chart `name`

WDYT?

image

### DIFF
--- a/pkg/engine/target/main.go
+++ b/pkg/engine/target/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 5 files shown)


jetersen:
@olblak would be great if you could review it :)

somewhat blocked by #55

olblak:
That's really great, thanks for the PR

PR #56, opened 2020-07-01
Add support for array in Target yaml key

### LINKED ISSUE #51: support updating partial values in cases of kustomize manifests
In kustomize you can have remote bases with versions like how argo-cd publish their manifest:
```yaml
bases:
- https://github.com/argoproj/argo-cd/manifests/cluster-install?ref=v1.5.7
```

Not sure the current "target" supports targeting a array index and replacing the value.
The source in argo-cd case would be github releases.

Something like this would work.

```tpl
source:
  kind: githubRelease
  spec:
    owner: "argoproj"
    repository: "argo-cd"
    version: "latest"
targets:
  argocd:
    name: "ArgoCD"
    kind: yaml
    spec:
      file: "kustomize/argo-cd/kustomization.yaml"
      key: bases[0]
      value: https://github.com/argoproj/argo-cd/manifests/cluster-install?ref={{ .version }}
```

olblak:
This pull request allows updating YAML values specified with an array position

resolve #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,10 @@ targets:
       # or
       #git:
 ```
+
+NOTE: A key can either be string like 'key' or a position in an array like `array[0]` where 0 means the first element of `array`.
+Keys and arrays can also be grouped with dot like `key.array[3].key`.
+
 #### scm
 Yaml accept two kind of scm, github and git.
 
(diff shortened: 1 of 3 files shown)


jetersen:
You absolute hero 💪

PR #106, opened 2020-10-24
Only clone git repository once

### LINKED ISSUE #105: Only clone git repository once
In the current state we run git clone then we did delete the directory after each operation in order to be sure that we don't have corrupted repositories. 

I want to clean git repositories only once and delete the temporary directory at the end unless we explicitly use a cli parameter to disable the clean behavior

olblak:
Fix #105 

This PR improve several things:
* Add a clean parameter for the diff subcommand like `updatecli diff --clean=true`
* By default delete the temporary directory content at the end of a full run instead of per strategy run, so we only clone git repositories once
* Change --clean parameter to keep or delete the main working directory instead of per repository 

This PR drastically drastically improve performances, I tested on [jenkins-infra/charts](https://github.com/jenkins-infra/chart)

Before:
`updatecli apply --config updateCli/updateCli.d --values updateCli/values.yaml  39,12s user 6,91s system 43% cpu 1:45,92 total`
After:
`updatecli apply --config updateCli/updateCli.d  --values updateCli/values.yaml   30,29s user 3,60s system 66% cpu 51,092 total
`

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -14,7 +14,7 @@ var (
 
 	applyCmd = &cobra.Command{
 		Use:   "apply",
-		Short: "apply checks if an updated is needed then apply the changes",
+		Short: "apply checks if an update is needed then apply the changes",
 		Run: func(cmd *cobra.Command, args []string) {
 			fmt.Printf("\n%s\n\n", strings.ToTitle("Apply"))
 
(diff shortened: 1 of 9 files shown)


jetersen:
Hmm the cleaning of the directory is one speed improvement.

I however think another place is to perform `clone` only once perhaps by determinering each repository that would need a clone and than just use `git checkout $branch` to reset the working directory before each operation.

olblak:
> Hmm the cleaning of the directory is one speed improvement.
> 
> I however think another place is to perform `clone` only once perhaps by determinering each repository that would need a clone and than just use `git checkout $branch` to reset the working directory before each operation.
> 
> https://github.com/olblak/updatecli/blob/4f51ca1e20eca618823f1ff207d67cf6dccd0ba5/pkg/engine/target/main.go#L88

That's a good point, thanks for the suggestion. It won't be too hard to do that and it will be useful for later stages.

olblak:
Cloning the repository once requires some refactoring so I am going to merge this one and then a new one when ready

PR #107, opened 2020-10-26
use latest download link in github action

jetersen:
this removes the pinning in the docs. Not sure if you want to have comment there suggesting they can pin it.


olblak (approved):
Good point

PR #109, opened 2020-10-29
Only clone git repository once at the beginning on the run

olblak:
This PR refactor the main engine by 
- Adding a new stage named "prepare" where we can clone needed git repositories only once.
- Exports the GitHub parameter `Directory` in order to define custom directories from updatecli.
- Simplifies the condition spec interface by moving the logic to the package which satisfies the interface. Now the condition interface requires two interface method condition(version) and conditionFromSCM(version, scm)
- Simplifies the target spec interface by moving the logic to the package which satisfies the interface. Now the target interface requires two interface method target and targetFromSCM
- Reorganise code under `core` and `plugins` directory
- Allow git/github to add several files in once

Time:
./bin/updatecli apply --config  --values   21,07s user 1,95s system 66% cpu 34,422 total


jetersen:
🚀 Great job on this major refactoring. Nice to see those performance gains!

PR #123, opened 2020-11-18
tidy up publish-assets

jetersen:


Makefile
```diff
@@ -4,10 +4,6 @@ BUILD_DATE=`date -R`
 VERSION=$(shell git describe --tags)
 GOVERSION=$(shell go version)
 
-ENV=updateCli.env
```
jetersen, line 7: As mentioned in https://github.com/olblak/updatecli/pull/117#issuecomment-724665899 this part is not needed.

Makefile
```diff
@@ -4,10 +4,6 @@ BUILD_DATE=`date -R`
...
-
-include $(ENV)
```
jetersen, line 9: As mentioned in https://github.com/olblak/updatecli/pull/117#issuecomment-724665899 this part is not needed.

olblak (approved):

olblak (approved):
For some reason the test is not passing because it doesn't' have access to a dockerhub secret

jetersen:
> For some reason the test is not passing because it doesn't' have access to a dockerhub secret

You need to pass the env through I think :)

PR #146, opened 2021-01-25
chore: handle the error after creating the file in the yaml target

garethjevans:


olblak (approved):

olblak:
The test is failing because you don't have access to the dockerhub credential

PR #147, opened 2021-01-25
chore: add error handling around reported areas from golangci-lint

garethjevans:


olblak:
\o/ Thanks for this

pkg/core/engine/main.go
```diff
@@ -30,13 +28,12 @@ type Engine struct {
 
 // Clean remove every traces from an updatecli run
 func (e *Engine) Clean() (err error) {
-	tmp.Clean()
-	return err
+	err = tmp.Clean()
+	return
```
olblak, line 32: That's interesting that you use the implicit return, without specifying `return err`

pkg/plugins/github/changelog.go
```diff
@@ -42,13 +42,6 @@ func (g *Github) Changelog(name string) (string, error) {
 
 	client := g.NewClient()
 
-	type release struct {
```
olblak, line 45: I have to double check because I think it's needed for the gaphql calll

olblak:
I have to slightly modify your PR, initially I wasn't correctly handling Dockerhub response and with your PR, it correctly fail :p

PR #162, opened 2021-01-28
chore: convert fmt.Print* to logrus

garethjevans:


olblak:
@garethjevans  When I evaluated a logging framework, Logrus came up first but I was afraid of the big warning on the main page. 

> Logrus is in maintenance-mode. We will not be introducing new features. It's simply too hard to do in a way that won't break many people's projects, which is the last thing you want from your Logging library (again...).

Which is why I delayed using it, any opinions?

garethjevans:
@olblak I took that to mean, they won't be introducing new features that will break the API

> This does not mean Logrus is dead. Logrus will continue to be maintained for security, (backwards compatible) bug fixes, and performance (where we are limited by the interface).

This should mean its fairly stable.

Issue #160, opened 2021-01-28
Add transformers parameter


olblak:
**Is your feature request related to a problem? Please describe.**

A major benefits of updatecli is to retrieve information from a third location then apply string manipulation to it by 
adding prefix, postfix, or replacer. More ways of modifying string is needed and we'll quickly need to specify the order of string manipulation 

**Describe the solution you'd like**

It would be simpler and better to add a new key for every stage that would accept a list of transformer, so we could control the order  of transformations

```
targets:
    targetID:
       kind: yaml
       transformers:
           - prefix: 
           - postfix:
           - replacer:
           - ...
```

**Describe alternatives you've considered**
We could use "modifiers" instead of "transformers"

**Additional context**
The current way of specifying prefix/postfix should remain but with a deprecation warning

PR #167, opened 2021-01-30
[File] add line matcher to File resource

olblak:
Add parameter "line" to identify a specific line.
It behaves differently depending on the stage

**source** 
Return the first list that contain the string

**condition**
Return true for the first line that match the string

**target**
Update file based on the first line returned from string

### DIFF
--- a/examples/updateCli.generic/file/checksum.tpl
+++ b/examples/updateCli.generic/file/checksum.tpl
@@ -3,31 +3,35 @@ source:
   kind: file
   spec:
     file: https://releases.hashicorp.com/terraform/0.14.5/terraform_0.14.5_SHA256SUMS
-    line: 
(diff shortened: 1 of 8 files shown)


olblak:
I am starting to wonder if it's the proper place to put this. 
Maybe it's better to add transformer rule to #160  to include a line, includeOnly, and exclude lines

Like

```
transformers:
    - line:
        includeOnly: "arm64" # Only takes the line containing the string "arm64"
```

```
transformers:
    - line:
        exclude: 
           - "arm64" # Exclude the line containing the string "arm64"
           - "amd64" # Also exclude the line containing the string "amd64"
```

olblak:
Rename:
* IncludesOnly by Include 
* Include by "HasIncludes"

PR #168, opened 2021-01-31
Add transformer definition

### LINKED ISSUE #152: Add string matcher
I am looking for a way to extract sub information from a string manipulated by a stage where
* source would extract sub information based on a string
* a condition would test that the match works
* a target would only use the value returned by the regex

I am looking to provide a regex expression to the key match, as in the following example.
- > https://golang.org/pkg/regexp/#Match
```
source:
  name: "Retrieve latest jenkins weekly version"
  kind: githubRelease
  match: <regex> 
  spec:
    owner: "jenkinsci"
    repository: "jenkins"
    version: "latest"
```

/

**Additional context**
A typical scenario would be to retrieve package checksum for a specific version to run some validation like from 
https://releases.hashicorp.com/terraform/0.14.5/terraform_0.14.5_SHA256SUMS

### LINKED ISSUE #160: Add transformers parameter

olblak:
Fix #160 
Fix #152 

Add the parameter `transformers` for every stage definitions, "source", "condition", and "target".
A "transformers" contains a list of transformer rule which are applied successively to the source value.
Each transformer applies a different rule

We also deprecated string manipulation like prefix/postfix/replacer defined outside a transformers definition

## transformer

* `prefix`, add a prefix to the source value
* `suffix`, add a suffix to the source value
* `trimPrefix`, remove the source value prefix
* `trimSuffix`, remove the source value suffix
* `replacer`, replace a substring by another one in the source value
* `find`, return the first occurrence that matches the regex

```
source:
  name: "Get latest jenkins weekly version"
  kind: jenkins
  transformers:
    - prefix: "alpha-"
    - suffix: "-jdk11"
    - trimSuffix: "-jdk11"
    - suffix: "-jdk11"
    - replacer:
        from: "-jdk11"
        to: "-jdk15"
    - replacers:
        - from: "-jdk15"
          to: "-jdk17"
  spec:
    release: weekly
targets:
  targetID:
    name: "Update file file"
    kind: file
    spec:
      file: TODO
```

This rule should generate the value "alpha-2.277-jdk11"

### DIFF
--- a/examples/updateCli.generic/transformers/jenkins.tpl
+++ b/examples/updateCli.generic/transformers/jenkins.tpl
@@ -2,10 +2,10 @@ source:
   name: "Get latest jenkins weekly version"
   kind: jenkins
   transformers:
-    - prefix: "alpha-"
-    - suffix: "-jdk11"
+    - addPrefix: "alpha-"
+    - addSuffix: "-jdk11"
     - trimSuffix: "-jdk11"
-    - suffix: "-jdk11"
+    - addSuffix: "-jdk11"
     - replacer:
         from: "-jdk11"
         to: "-jdk15"
--- a/examples/updateCli.generic/transformers/jenkins.tpl
+++ b/examples/updateCli.generic/transformers/jenkins.tpl
@@ -1,0 +1,22 @@
+source:
+  name: "Get latest jenkins weekly version"
+  kind: jenkins
+  transformers:
+    - prefix: "alpha-"
+    - suffix: "-jdk11"
+    - trimSuffix: "-jdk11"
+    - suffix: "-jdk11"
+    - replacer:
+        from: "-jdk11"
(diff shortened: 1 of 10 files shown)


olblak:
Add "add" to prefix and suffix

Issue #154, opened 2021-01-28
Improve distribution of updateCli


dduportal:
**Is your feature request related to a problem? Please describe.**

* I would like to verify the checksum of the `updateCli` binary I'm downloading
* I would like to be able to use `updateCli` as a Docker image sometimes (instead of a binary)
* As a macOS user, I would like to be able to install and update `updateCli` with homebrew, the package manager

**Describe the solution you'd like**

The goreleaser tool (https://goreleaser.com/) could be a great help for solving all of this requests at once (and maybe more)

**Describe alternatives you've considered**

* Scripting the generation of checksum can be done quite easily
* It looks like there is already a Dockerfile at the repository's root: adding a piece of doc could be helpful

**Additional context**

N.A.

olblak:
Almost everything as been implemented in https://github.com/updatecli/updatecli/releases/tag/v0.3.4
I am still missing the brew part but as I a non macOS user, I am not planning to spend more time on this

PR #180, opened 2021-02-18
Switch build process to goreleaser

dduportal:
This PR aims at switching the build process from tedious shell scripting to [GoReleaser](https://goreleaser.com/quick-start/).
It is a foundational work for the futures PR about #154.

The scope of the PR is to provide a replacement "as it", without enabling all the powerful feature of goreleaser yet, which means:

- Only build the binaries with the same name as before
- No release, no binary checksum, no changelog, no homebrew support managed by goreleaser

Please note that command `make build` is now building all of the binaries:

- The rationale is to use build matrix instead of 1 build per couple `GOOS` + `GOARCH` as in https://github.com/garethjevans/inspect/blob/main/.goreleaser.yml#L9 (ldflags are not duplicated, etc.).
- The overhead is ~ 0.4s on my machine (MBP Core i7 2,6 Ghz, 16 Go, NVMe): Running 15 times the build has a median build time of 4.2s (max 4.8, min 3.9) while it was 3.8s before goreleaser (max 4.4 min 3.6).
- The difference with `make build.all` **for now** is that `make build` is aimed at development, so it uses the `--snapshot` flag for goreleaser.


olblak (approved):
Thanks for this PR, I can't wait to use more goreleaser

PR #181, opened 2021-02-18
[WIP] Allows reinjecting configuration value into a different places

### LINKED ISSUE #158: Allow to define and use multiple sources

In the current state, we can only define one source per configuration file. 
Sometimes to update a target, the information need to come from multiple sources, as example is 
to fetch both the latest version of a package and its checksum from a different file. 
Combined with #156 this would also drastically simplify pull request as every changes related to one thing would be grouped in the same pull request.

Obviously having multiple sources would also mean that every condition/target should be able to either specify the source or fetch a default source value

The sources would accept a map source identified by a sourceID
Conditions and and targets would accept the key "from" to define which source value to use by default

```
sources:
  source1:
    name: Fetch value from 
    kind: file
    spec:
      file: "http://example.com/checksum.txt"
  source2:
    name: Fetch package version
    kind: githubRelease
    spec:
      owner: owner
      repository: repositry
      username: username
      email: email
      token: token
conditions:
  condition1:
    from: source2
    name: Test Docker image publish
    kind: dockerImage
    spec:
     image: docker_image 
  condition2:
    from: source1
    name: Test content
    kind: file
    spec:
     file: "http://example.com/checksum.txt"
targets:
  targetID1:
    from: source1
    name: Test Docker image publish
    kind: dockerfile
    spec:
     file: Dockerfile
     instruction: ARG[0][0]
  targetID2:
    from: source2
    name: Test Docker image publish
    kind: dockerfile
    spec:
     file: Dockerfile
     instruction: ARG[1][0]

```

/

**Additional context**
/

olblak:
This PR is an attempt to partially fix #158

It allows reinjecting configuration value into a different places.
The function "pipeline" accept one parameter where the value correspond to an updatecli configuration key, value where the key must start with a capital letter. 

which can be defined like: 
```
title: {{ pipeline "Source.Output" }}
source:
  name: Get Latest helm release version
  kind: githubRelease
  spec:
    owner: "helm"
    repository: "helm"
    token: {{ requiredEnv .github.token }}
    username: olblak
    version: latest
```
The configuration is updated after each stage, if it finds a function configuration value set to `pipeline "updatecli_config"`, then it tries to replace with a value but only if the value exists otherwise it keeps `pipeline "updatecli_config"` for the next run .

This pull request also remove the constraint to only use go template with files extension .tpl

Once the multiple sources is implemented, I'll add a new template function named Source which will accept as a parameter a source id. The purpose will be to have a simpler way to retrieve a source output

```
title: {{ source "Id" }}
sources:
  id:
    name: Get Latest helm release version
    kind: githubRelease
    spec:
      owner: "helm"
      repository: "helm"
      token: {{ requiredEnv .github.token }}
      username: olblak
      version: latest
```

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -1,9 +1,14 @@
 package config
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
+	"text/template"
 
 	"github.com/mitchellh/hashstructure"
 	"github.com/sirupsen/logrus"
@@ -11,8 +16,21 @@ import (
 	"github.com/olblak/updateCli/pkg/core/engine/condition"
 	"github.com/olblak/updateCli/pkg/core/engine/source"
 	"github.com/olblak/updateCli/pkg/core/engine/target"
-	"github.com/spf13/viper"
 	"gopkg.in/yaml.v3"
+)
+
+var (
+	// ErrConfigFileTypeNotSupported is returned when updatecli try to read
+	// an unsupported file type.
+	ErrConfigFileTypeNotSupported = errors.New("file extension not supported")
+
+	// ErrNoEnvironmentVariableSet is returned when during the templating process,
+	// it tries to access en environment variable not set.
+	ErrNoEnvironmentVariableSet = errors.New("Environment variable doesn't exist")
+
+	// ErrNoKeyDefined is returned when during the templating process, it tries to
+	// retrieve a key value which is not defined in the configuration
+	ErrNoKeyDefined = errors.New("key not defined in configuration")
 )
 
 // Config contains cli configuration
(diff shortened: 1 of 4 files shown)


olblak:
More I think about this PR and less I am convinced that it's the right approach

dduportal:
@olblak WDYM?

olblak:
@dduportal They are many side cases that I have to deal with.

First, because of an initial design, we can't predict the order of execution for sources/conditions/targets as they are using maps instead of lists. 

The reason to use a map instead of a list was driven by viper as describe on this [PR](https://github.com/updatecli/updatecli/pull/195). viper can set value based on environment variable as long as it respected the format UPDATECLI_CONDITION_ID_KEY
It didn't support lists and I had to define one environment variable per key/value so I introduced Go templates.

I am now investigating with HCL as it could build a graph of execution for source,condition and targets.

One of the requirements will be to switch from maps to lists.

dduportal:
In order to stay with maps, WDYT about the following:

* Maps of sources/conditions are by default sorted in the lexicographic order 
* You might had a custom attribute to the map which would be the "priority"

olblak:
I am closing this PR as it has been superseded by #242

PR #183, opened 2021-02-20
Add multi sources support

olblak:
This PR comes hand in hand with #181  to solve #158 
While keeping backward compatibility, I am switching from one source to a map of sources.
Now for each condition/target, we must specify the `sourceID` parameter to specify which source a condition/target depends on, 
it mainly used to get default values for those stages.  

An example of this change
```
---
title: Bump jenkinsciinfra/helmfile image
sources:
  helm:
    name: Get Latest helm release version
    kind: githubRelease
    spec:
      owner: "helm"
      repository: "helm"
      token: {{ requiredEnv .github.token }}
      username: olblak
      version: latest
  helm2:
    name: Get Latest helm release version
    kind: githubRelease
    spec:
      owner: "helm"
      repository: "helm"
      token: {{ requiredEnv .github.token }}
      username: olblak
      version: latest
conditions:
  isENVSet:

```

### DIFF
--- a/examples/updateCli.d/sources/dockerfile_sources.tpl
+++ b/examples/updateCli.d/sources/dockerfile_sources.tpl
@@ -1,0 +1,71 @@
+---
+title: Bump jenkinsciinfra/helmfile image
+sources:
+  helm:
+    name: Get Latest helm release version
+    kind: githubRelease
+    spec:
+      owner: "helm"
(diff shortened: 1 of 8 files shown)


olblak:
I modified this PR to only focus on multi sources and remove any mention of more advanced templating.

PR #199, opened 2021-03-25
fix: make tests pass by mocking out auth'd http calls

garethjevans:


olblak:
Thanks for the PR, since you are transforming some integration test into unit tests, I have to carefully review it

garethjevans:
Probably shouldn't be running integration tests as part of the build, we should tag them as "integration" tests and exclude them from the default build to exclude any external calls.

olblak:
As discussed I would be interested to keep integration tests as well but running them less often.

So unit tests would be executed using `go test -v -short` and long-running test with `go test -run Integration`

`

`

dduportal:
I'm closing this PR, as the httpclient is implemented since #418 .

I wasn't aware of this PR but it seems that it's the same idea (I'm not totally crazy ;) ).

Thanks a lot!

Issue #159, opened 2021-01-28
Allow to configuration changelog


olblak:
**Is your feature request related to a problem? Please describe.**
Right now the way to retrieve a changelog depends on the source resource. 
It would be nice to allow to parameterized it.

**Describe the solution you'd like**

```
sources:
  source1:
    name: Fetch value from 
    kind: file
    spec:
      file: "http://example.com/checksum.txt"
    changelog:
        kind:
        spec:
            github:
                owner:
                ....
```

**Describe alternatives you've considered**

**Additional context**

olblak:
While the need is still there the configuration need to be think more

PR #208, opened 2021-04-02
First get changelog information before applying transformers to the source value

olblak:
In the default behavior, updatecli tries to guess changelog information using the value returned from a source. 
There is a bigger chance that this initial value can be used to retrieve changelog information than once it has been transformed for later stages.

Thanks, @dduportal  for identifying this issue

Btw I still have some plan to define a changelog configuration to customize the way we retrieve changelog information as described in this issue #159 


dduportal (approved):
LGTM 👍

PR #220, opened 2021-04-21
Use Goreleaser to build and publish assets

olblak:
Use Goreleaser to build and publish assets (including docker images)


dduportal (approved):
🚀

PR #221, opened 2021-04-21
BREAKING: Add git push --force parameter

olblak:
In the current stage, we always push commits/tags using the --force parameter.
I propose the parameter `force` which can be specified at the Github/git level to enable or disable that behavior with the default behavior set to no force push which is a breaking change compared to the current situation.


dduportal (approved):
LGTM 👍 

Do'nt forget to prefix the PR's title with `BREAKING:` because it's a breaking change ;)

oleg-nenashev (approved):

PR #224, opened 2021-04-22
fix: Re-install ca-certificates in Dockerfile

olblak:
This is regression introduced by #220 


dduportal (approved):
🚀

PR #246, opened 2021-07-03
Add username/password parameter for Docker Registry authentication

### LINKED ISSUE #245: unauthorized ghcr.io access
**Describe the bug**
It appears that tests are now failing as we can't authenticate on ghcr.io API anymore
-> https://github.com/updatecli/updatecli-action/runs/2980227475?check_suite_focus=true

**To Reproduce**

cp pkg/plugins/docker/registry/ghcr
go test 

**Expected behavior**
It used to be working

 - OS: linux
 - Version: v0.4.0

**Additional context**

olblak:
Fix #245 

For some reason, the previous authentication mechanism using Token doesn't work anymore and changing to bearer like other registry works

This pull request refactor docker registry authentication, we should now specific a username and password parameter when we need to authenticate with a docker registry such as 
 
```
dockerImage:
    kind: dockerImage
    postfix: "-jdk11"
    spec:
        image: jenkins/jenkins
        username: bob
        password: xxx
```

**Remark**
Credentials required for the new test are not available from Github Action triggered by pull request from fork repository so I am introducing short test and long test to only execute short tests in such cases


olblak:
GITHUB_TOKEN doesn't work with bearer token which is a serious regression, I need to investigate other options

olblak:
It seems that secrets are not available from the runner

olblak:
Credentials required for the test are not available from github action running for pull request so I am introduce short test and long test to only execute short tests from PR runner

olblak:
https://docs.docker.com/registry/spec/api/#content-digests

PR #247, opened 2021-07-04
Create CODE_OF_CONDUCT.md

olblak:


oleg-nenashev:
I was just about doing the same for
https://github.com/oleg-nenashev/FaaScinator
Now I feel like a copycat :)

PR #254, opened 2021-07-13
Add Runtime templating

### LINKED ISSUE #158: Allow to define and use multiple sources

olblak:
# Add runtime templating

Fix #158

This pull request is a third attempt to introduce runtime templating.
Some refactoring was needed to avoid templating unwanted information such as changelogs containing mustache syntax.

On top of https://github.com/updatecli/updatecli/pull/242 
I modified the `pipeline` template function to only fetch information from context such as stage Results. I keep it undocumented for now
I introduce the `source` template to quickly retrieve a source result 

This pr also fix minor issues such as:
- Wrong message `Since version 1.2.0...` become `Since version 0.2.0`
- Run every condition even if one of them is failing, so we directly have visibility on what needs to be fixed
- Remove duplicate error message

The following snippet is an example of how to use the source function in a template to reuse source output. 
Just insert this syntax {{ source "<< source id >>" }}

```
sources:
  stable:
    kind: jenkins
conditions:
  stabledockerImage:
    name: 'Is docker image jenkins/jenkins:{{ source "stable" }} published?'
    kind: dockerImage
    sourceID: stable
    spec:
      image: jenkins/jenkins
      tag: '{{ source "stable" }}-jdk11'
  weeklydockerImage:
    name: 'Is docker image tag{{ pipeline "Sources.weekly.Output" }} published?'
    kind: dockerImage
    sourceID: weekly
    spec:
      image: jenkins/jenkins
      tag: '{{ pipeline "Sources.weekly.Output" }}-jdk11'

```

## Test

```
  make test-short
  export GITHUB_TOKEN=xxx
  go build -o bin/updatecli
  ./bin/updatecli diff --config examples/updateCli.generic/runtimeTemplating.tpl --values  examples/values.yaml
```

### Tradeoff

I am missing a lot of UT in this specific area but I also lack of time. 😞 

### Potential improvement

Additional refactoring is needed for the engine package
Additional tests are needed for the engine package


olblak:
After back and forth thinking, I am removing the go template function pipeline.
-  pipeline is a bit too dangerous as it creates a strong binding between the data structure context used inside updatecli and upatecli configuration.
- pipeline is now redundant with the basic templating feature as we don't anymore modify the configuration at runtime. 

For now, I'll just document the source function template  as it solves the initial need which was being able to reuse source output in other configuration

PR #264, opened 2021-08-19
Add a new resource of type "shell"

### LINKED ISSUE #263: Add a new resource of type "shell"

Sometime, I would want to execute a custom shell script to retrieve a "source" for updatecli when it does not support it (for instance: until #222 is implemented, I would use a `get_latest_ami.sh` script instead to retrieve the string I expec, or any unsupported system.

A new provider `shell`that would:

- for sources, uses the stdout of the command as value for the source, if and only if the exit code is 0, otherwise updatecli fails, and the stderr of the command would be shown as the updatecli CLI output

- for conditions, pass the provided source as argument to the command, and uses the exit code to validate the condition

- for target, pass the provided source as argument to the command, uses the exit code to see if it changes something, or not, or failed (code 0 = success with nothing changed, code 2 means success with the change printed on the stdout, other ocdes means failure)

```yaml
title: Do something
sources:
  default:
    kind: shell
    spec:
      command: "bash -x get_latest_ami.sh {{ .aws.account_name }}"
      name: get latest AMI
conditions:
  exist:
    name: Check that the AMI has the tag "cb-ready" present
    kind: shell
    spec:
      sourceID: default
      # Script check_tag_on_ami expectes $1 to be the provided template value, and $2 to be the value from "default"
      command: "bash -x check_tag_on_ami.sh {{ .aws.account_name }}"
      name: cluster-autoscaler
targets:
  copyAmi:
    name: Copy the AMI from us-east-1 to the region eu-west-2
    kind: shell
    spec:
      sourceID: default
      # the command expects a $1 argument set to the value of the source
      command: "aws ec2 copy-image --region-from=us-east-1 --region-to=eu-west-2"
```

N.A.

**Additional context**

This "generic" provider:

- Could allow supporting non native existing providers, hence limiting the amount of work
- Should be able to (nice to have) customize the command prompt: `sh`, `zsh` or even... `powershell` on windows :)

dduportal:
# Add a new resource of type "shell"

Closes #263 

## What this PR does?

It introduces a new resource of type `shell` to allow executing a custom command.

* For sources, this resource uses the stdout of the command to report the value to updatecli, if and only if the exit code is 0, otherwise it fails, and the stderr of the command is reported by updatecli to the end user.

* For conditions, this resource executes a command with the associated source appended as last argument. If the exit code is 0 then the condition succeed, otherwise it fails.

* For targets, this resource executes a command with the associated source appended as last argument. Then:
  * If the exit code is zero AND the stdout reports a non empty string, then updatecli reports that there was a change, and the stdout reports what has changed.
  * If the exit code is zero and stdout is empty, then updatecli report that there was no change (but still success)
  * Another exit code means that the target failed.
  * The environment variable `DRY_RUN` is set either to `false` or `true` in the target's command environment

There is an generic example in `examples/updateCli.generic/shell/shell.yaml` (along with shell scripts).

## Test

```
go test ./pkg/plugins/shell -v
```

~95% of the package `shell` is covered

## Additional Information

* Not tested on Windows, but as it relies on `os/exec` Golang package AND each command is a single strings, it should work as expected

### Potential improvement

* If someone raises the issue, the `command` specification could be improved to support "array mode" ala Dockerfile (for entrypoints/CMD in `Dockerfiles`)

### DIFF
--- a/examples/updateCli.generic/shell/shell.yaml
+++ b/examples/updateCli.generic/shell/shell.yaml
@@ -3,16 +3,16 @@ sources:
   default:
     kind: shell
     spec:
-      command: "bash -x ./examples/updateCli.generic/shell/source.sh"
+      command: "./examples/updateCli.generic/shell/source.sh"
 conditions:
   default:
     kind: shell
     spec:
-      command: "bash -x ./examples/updateCli.generic/shell/condition.sh"
+      command: "./examples/updateCli.generic/shell/condition.sh"
 targets:
   default:
     name: setGrepVersion
     sourceID: default
     kind: shell
     spec:
-      command: "bash -x ./examples/updateCli.generic/shell/target.sh"
+      command: "./examples/updateCli.generic/shell/target.sh"
--- a/examples/updateCli.generic/shell/shell.yaml
+++ b/examples/updateCli.generic/shell/shell.yaml
@@ -1,0 +1,18 @@
+---
+sources:
+  default:
+    kind: shell
+    spec:
+      command: "bash -x ./examples/updateCli.generic/shell/source.sh"
+conditions:
+  default:
+    kind: shell
+    spec:
+      command: "bash -x ./examples/updateCli.generic/shell/condition.sh"
+targets:
+  default:
+    name: setGrepVersion
+    sourceID: default
+    kind: shell
+    spec:
+      command: "bash -x ./examples/updateCli.generic/shell/target.sh"
--- a/pkg/core/engine/condition/main.go
+++ b/pkg/core/engine/condition/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/updatecli/updatecli/pkg/plugins/helm/chart"
 	"github.com/updatecli/updatecli/pkg/plugins/jenkins"
 	"github.com/updatecli/updatecli/pkg/plugins/maven"
+	"github.com/updatecli/updatecli/pkg/plugins/shell"
 	"github.com/updatecli/updatecli/pkg/plugins/yaml"
 )
 
@@ -182,6 +183,16 @@ func Unmarshal(condition *Condition) (spec Spec, err error) {
 		}
 
 		spec = &y
+
+	case "shell":
+		shellResourceSpec := shell.ShellSpec{}
+
+		err := mapstructure.Decode(condition.Spec, &shellResourceSpec)
+		if err != nil {
+			return nil, err
+		}
+
+		spec = shell.New(shellResourceSpec)
 
 	default:
 		return nil, fmt.Errorf("Don't support condition: %v", condition.Kind)
--- a/pkg/core/engine/condition/main.go
+++ b/pkg/core/engine/condition/main.go
@@ -187,8 +187,7 @@ func Unmarshal(condition *Condition) (spec Spec, err error) {
 	case "shell":
 		shellResourceSpec := shell.ShellSpec{}
 
-		err := mapstructure.Decode(condition.Spec, &shellResourceSpec)
-		if err != nil {
+		if err := mapstructure.Decode(condition.Spec, &shellResourceSpec); err != nil {
 			return nil, err
 		}
 
--- a/pkg/core/engine/condition/main.go
+++ b/pkg/core/engine/condition/main.go
@@ -185,7 +185,7 @@ func Unmarshal(condition *Condition) (spec Spec, err error) {
 		spec = &y
 
 	case "shell":
-		shellResourceSpec := shell.ShellSpec{}
+		var shellResourceSpec shell.ShellSpec
 
 		if err := mapstructure.Decode(condition.Spec, &shellResourceSpec); err != nil {
 			return nil, err
--- a/pkg/core/engine/condition/main.go
+++ b/pkg/core/engine/condition/main.go
@@ -191,7 +191,10 @@ func Unmarshal(condition *Condition) (spec Spec, err error) {
 			return nil, err
 		}
 
-		spec = shell.New(shellResourceSpec)
+		spec, err = shell.New(shellResourceSpec)
+		if err != nil {
+			return nil, err
+		}
 
 	default:
 		return nil, fmt.Errorf("Don't support condition: %v", condition.Kind)
--- a/pkg/plugins/shell/command.go
+++ b/pkg/plugins/shell/command.go
@@ -27,7 +27,7 @@ type commandExecutor interface {
 type nativeCommandExecutor struct {}
 
 func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResult, error) {
-	if inputCmd.Cmd == "" {
+	if strings.TrimSpace(inputCmd.Cmd) == "" {
 		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
 	}
 
--- a/pkg/plugins/shell/command.go
+++ b/pkg/plugins/shell/command.go
@@ -3,7 +3,6 @@ package shell
 import (
 	"bytes"
 	"errors"
-	"fmt"
 	"os/exec"
 	"strings"
 )
@@ -24,13 +23,9 @@ type commandExecutor interface {
 	ExecuteCommand(cmd command) (commandResult, error)
 }
 
-type nativeCommandExecutor struct {}
+type nativeCommandExecutor struct{}
 
 func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResult, error) {
-	if strings.TrimSpace(inputCmd.Cmd) == "" {
-		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
-	}
-
 	var stdout, stderr bytes.Buffer
 	cmdFields := strings.Fields(inputCmd.Cmd)
 	command := exec.Command(cmdFields[0], cmdFields[1:]...) //nolint: gosec
--- a/pkg/plugins/shell/command.go
+++ b/pkg/plugins/shell/command.go
@@ -34,9 +34,7 @@ func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResul
 	var stdout, stderr bytes.Buffer
 	cmdFields := strings.Fields(inputCmd.Cmd)
 	command := exec.Command(cmdFields[0], cmdFields[1:]...) //nolint: gosec
-	if inputCmd.Dir != "" {
-		command.Dir = inputCmd.Dir
-	}
+	command.Dir = inputCmd.Dir
 	command.Stdout = &stdout
 	command.Stderr = &stderr
 	// Pass current environment to process and append the customized environment variables used internally by updatecli (such as DRY_RUN)
@@ -63,5 +61,4 @@ func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResul
 		Stdout:   out,
 		Stderr:   stderr.String(),
 	}, nil
-
 }
--- a/pkg/plugins/shell/command.go
+++ b/pkg/plugins/shell/command.go
@@ -1,0 +1,68 @@
+package shell
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+type command struct {
+	Cmd string
+	Dir string
+	Env []string
+}
+
+type commandResult struct {
+	ExitCode int
+	Stdout   string
+	Stderr   string
+}
+
+type commandExecutor interface {
+	ExecuteCommand(cmd command) (commandResult, error)
+}
+
+type nativeCommandExecutor struct {
+}
+
+func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResult, error) {
+	if inputCmd.Cmd == "" {
+		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmdFields := strings.Fields(inputCmd.Cmd)
+	command := exec.Command(cmdFields[0], cmdFields[1:]...) //nolint: gosec
+	if inputCmd.Dir != "" {
+		command.Dir = inputCmd.Dir
+	}
+	command.Stdout = &stdout
+	command.Stderr = &stderr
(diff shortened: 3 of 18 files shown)


pkg/plugins/shell/condition.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (s *Shell) condition(source, workingDir string) (bool, error) {
+	customCommand := s.spec.Command
+	if customCommand == "" {
+		return false, fmt.Errorf(ErrEmptyCommand)
+	}
+	if source != "" {
+		customCommand += " " + source
+	}
```
jlevesy, line 28: This block feels pretty unclear, could you explain it? 🙏

pkg/plugins/shell/condition.go
```diff
@@ -0,0 +1,46 @@
...
+		Cmd: customCommand,
+		Dir: workingDir,
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	if cmdResult.ExitCode != 0 {
+		logrus.Infof("%v The shell 🐚 command %q failed to validate the condition:\nexit code was %q\nstderr=\n%v\nstdout=\n%v\n.", result.FAILURE, customCommand, cmdResult.ExitCode, cmdResult.Stderr, cmdResult.Stdout)
```
jlevesy, line 39: Not sure if it is relevant, but logrus allows to do structured logging, ie adding attributes to the log line and let the logger handle the formatting, instead of doing it by yourself.

The upside here is that you can change the formatting of the whole log line to `JSON` for example and still have those attributes parse-able.

For (incomplete) example....

```suggestion
		 logrus.WithFields(log.Fields{
			"exit_code": cmdResult.ExitCode,
			"stdout": cmdResult.Stdout,
			"stderr": cmdResult.Stderr, 
  		}).Info("Unable to execute command...")
```

pkg/plugins/shell/main_test.go
```diff
@@ -0,0 +1,53 @@
+package shell
+
+import "fmt"
+
+////////////// Test Utilities method aimed at mocking calls to other objects
```
jlevesy, line 5: That's a whole lot of `//` 😁 😛

pkg/plugins/shell/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+// mocking commandExecutor.ExecuteCommand
```
jlevesy, line 7:
```suggestion
// mockCommandExecutor is a stub implementation of the `commandExecutor` interface to be used in our test suite. It stores the received `command` and returns the preconfigured `result` and `err`.
```

pkg/plugins/shell/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+	gotCommand command
+	result     commandResult
+	err        error
+}
+
+func (mce *mockCommandExecutor) ExecuteCommand(cmd command) (commandResult, error) {
+	mce.gotCommand = cmd
+	if mce.gotCommand.Cmd == "" {
+		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
+	}
```
jlevesy, line 18: Is this necessary? 🤔

pkg/plugins/shell/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+}
+func (m *mockScm) Commit(message string) error {
+	return nil
+}
+func (m *mockScm) Clean() error {
+	return nil
+}
+func (m *mockScm) PushTag(tag string) error {
+	return nil
+}
```
jlevesy, line 53:
```suggestion
type mockScm struct {
	scm.Scm

	workingDir string
}

func (m *mockScm) GetDirectory() (directory string) {
	return m.workingDir
}
```

In that case you could `embed` the interface you're trying to stub into your stub structure:
- It saves you defining a bunch of noop methods
- An unexpected call to a non stubbed method results in a panic: because it should not happen and you need to be aware it happened.

When you `embed` something into a structure, if the parent structure doesn't define a method call, then it is implicitely forwarded to the embedded value. 

In our case, let's say I'm calling `Clean` on our mock: 
- it is not defined on the mock, so it  will get delegated to the embedded value
- ...which is an interface, and the zero value of an interface is nil
- ... so it results on a call on a nil value, causing a panic.

embedding is not the most obvious thing to grasp, I would be happy to explain it to you live if you want 😄 

Does it makes sense?

pkg/plugins/shell/source.go
```diff
@@ -0,0 +1,29 @@
...
+		Cmd: s.spec.Command,
+		Dir: workingDir,
+	})
+
+	if err != nil {
+		return "", err
+	}
+
+	if cmdResult.ExitCode != 0 {
+		return "", fmt.Errorf("%v The shell 🐚 command %q failed with the following message: \nstderr=\n%v\nstdout=\n%v\n", result.FAILURE, s.spec.Command, cmdResult.Stderr, cmdResult.Stdout)
```
jlevesy, line 23: I do believe that a typed error could save you a whole lot of formatting duplication here :)

```
// Could be private
type ShellError struct {
    ErrCode int
    Stdout string
    Stderr string
}

func  (e *ShellError) Error() string {
    return fmt.Sprintf("yourformatted error %v%v %q", e.code, e.stdout, e.stderr)
}
```

So you can use it here

```suggestion
		return "",  &ShellError{Code: code, Stdout: stdout, Stderr: stderr}
```

pkg/plugins/shell/target.go
```diff
@@ -0,0 +1,59 @@
...
+//	- Any other exit code means "failed command with no change"
+// The environment variable 'DRY_RUN' is set to true or false based on the input parameter (e.g. 'updatecli diff' or 'apply'?)
+func (s *Shell) target(source, workingDir string, dryRun bool) (changed bool, commands []string, message string, err error) {
+	customCommand := s.spec.Command
+	if customCommand == "" {
+		return false, commands, message, fmt.Errorf(ErrEmptyCommand)
+	}
+	if source != "" {
+		customCommand += " " + source
+	}
```
jlevesy, line 33: I do believe that you could factorize this snippet using a private function for example.

jlevesy (commented):
👀 

Some style nits and suggestions, feel free to disregard if not relevant 😄

pkg/core/engine/condition/main.go
```diff
@@ -183,6 +184,16 @@ func Unmarshal(condition *Condition) (spec Spec, err error) {
 
 		spec = &y
 
+	case "shell":
+		shellResourceSpec := shell.ShellSpec{}
+
+		err := mapstructure.Decode(condition.Spec, &shellResourceSpec)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
olblak: Indeed, I usually prefer to inline this syntax

examples/updateCli.generic/shell/shell.yaml
```diff
@@ -0,0 +1,18 @@
...
+    kind: shell
+    spec:
+      command: "bash -x ./examples/updateCli.generic/shell/condition.sh"
+targets:
+  default:
+    name: setGrepVersion
+    sourceID: default
+    kind: shell
+    spec:
+      command: "bash -x ./examples/updateCli.generic/shell/target.sh"
```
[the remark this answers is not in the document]
dduportal: Your proposal is legit and make sense. We'll take care of NOT using bash in the documentation for this. The reason why it is explicitely defined here is to show a generic example that would be portable, as Windows should be part of the extent: and there is not a shebang on windows

pkg/plugins/shell/command.go
```diff
@@ -0,0 +1,68 @@
...
+	if inputCmd.Cmd == "" {
+		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmdFields := strings.Fields(inputCmd.Cmd)
+	command := exec.Command(cmdFields[0], cmdFields[1:]...) //nolint: gosec
+	if inputCmd.Dir != "" {
+		command.Dir = inputCmd.Dir
+	}
```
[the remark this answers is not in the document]
dduportal: Sounds good, tests are not failing with your suggestion, and it makes sense, thans!

pkg/plugins/shell/main_test.go
dduportal: I had to make sure that the reader see them :trollface:

pkg/plugins/shell/main_test.go
dduportal: It is not, in regard to your feedback about the "spec" checked before calling. If ExecuteCommand does NOT check for empty command (as it's the role of the caller), then this part should be deleted 👍

pkg/plugins/shell/main_test.go
dduportal: That makes 100% sense and is exactly what I was looking for (I mean, this bunch of noop code is annoying to have..). Many thanks!

examples/updateCli.generic/shell/shell.yaml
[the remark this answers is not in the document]
jlevesy: Alright, resolving :)

pkg/plugins/shell/main_test.go
jlevesy: Well, is this comment really useful? Doesn't the typeName and the type comment already testifies of their purpose? :grin:

pkg/plugins/shell/main_test.go
jlevesy: Even if you don't: you could configure the mock in your case to raise an error instead of hard-coding this behavior.

pkg/plugins/shell/condition.go
dduportal: Added a comment: WDYT?

pkg/plugins/shell/command.go
```diff
@@ -0,0 +1,68 @@
...
+	ExecuteCommand(cmd command) (commandResult, error)
+}
+
+type nativeCommandExecutor struct {
+}
+
+func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResult, error) {
+	if inputCmd.Cmd == "" {
+		return commandResult{}, fmt.Errorf(ErrEmptyCommand)
+	}
```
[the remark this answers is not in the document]
dduportal: It totally makes sense! I've moved the "ErrEmptyCommand" logic to the `shell.NEw()` method, so it raises the error before anything else. WDYT?

pkg/plugins/shell/command.go
```diff
@@ -0,0 +1,68 @@
...
+
+type commandExecutor interface {
+	ExecuteCommand(cmd command) (commandResult, error)
+}
+
+type nativeCommandExecutor struct {
+}
+
+func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResult, error) {
+	if inputCmd.Cmd == "" {
```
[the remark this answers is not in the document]
dduportal: Good point, even if this is only a YAML scalar value (not a key, not a complex structure). I'm not sure though: I fear that it would risk having side effects if there is a `grep` with these characters within. I don't think I'll do this in this PR: either waiting for an issue to be raised with a use case, or maybe while adding an "array" form for specifying the spec.command (instead of the actual string).

pkg/plugins/shell/condition.go
```diff
@@ -0,0 +1,46 @@
...
+// ConditionFromSCM tests if the provided command (concatenated with the source) is executed with success from the SCM root directory
+func (s *Shell) ConditionFromSCM(source string, scm scm.Scm) (bool, error) {
+	return s.condition(source, scm.GetDirectory())
+}
+
+func (s *Shell) condition(source, workingDir string) (bool, error) {
+	customCommand := s.spec.Command
+	if customCommand == "" {
+		return false, fmt.Errorf(ErrEmptyCommand)
+	}
```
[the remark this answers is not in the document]
dduportal: Done: moved to `shell.New()`

pkg/plugins/shell/main_test.go
dduportal: Yep, I thought the same and removed it :)

pkg/plugins/shell/main_test.go
dduportal: With the validation logic moved to shell.New, this code had been removed

pkg/plugins/shell/condition.go
```diff
@@ -17,12 +17,8 @@ func (s *Shell) ConditionFromSCM(source string, scm scm.Scm) (bool, error) {
...
 func (s *Shell) condition(source, workingDir string) (bool, error) {
-	customCommand := s.spec.Command
+	customCommand := s.customCommand(source)
 
-	// Append the source as last argument if not empty
-	if source != "" {
-		customCommand += " " + source
-	}
 	cmdResult, err := s.executor.ExecuteCommand(command{
 		Cmd: customCommand,
```
jlevesy, line 23:
```suggestion
		Cmd: s.customCommand(source),
```

You don't need a variable anymore.

pkg/plugins/shell/main.go
```diff
@@ -18,3 +18,12 @@ func New(spec ShellSpec) (*Shell, error) {
 		spec:     spec,
 	}, nil
 }
+
+func (s *Shell) customCommand(source string) string {
```
jlevesy, line 22:
```suggestion
// Append the source as last argument if not empty. 
func (s *Shell) appendSource(source string) string {
```

customCommand doesn't mean much.

pkg/plugins/shell/main_test.go
```diff
@@ -75,3 +75,35 @@ func TestShell_New(t *testing.T) {
 		})
 	}
 }
+
+func TestShell_customCommand(t *testing.T) {
```
jlevesy, line 79: Testing internal methods is not necessary, this behavior is already ensured by your public methods tests.

jlevesy (commented):
👀

jlevesy (approved):
:eyes:

Style nit, otherwise LGTM

olblak (approved):
Nice pull request, now I have to upper the quality of mines :D

dduportal:
> Nice pull request, now I have to upper the quality of mines :D

Most of the credit goes to @jlevesy ❤️ for his careful reviews, explanation and teaching me :)

PR #265, opened 2021-08-22
Add a new resource to manipulate AWS AMI

### LINKED ISSUE #222: Add a new resource of type "AMI" (Amazon Machine Image)

I would like to use to keep up-to-date the AMI references of some of my projects:

- Jenkins CasC YAML (EC2 Cloud setup)
- Terraform project throug variables written in a [`.tfvars`](https://www.terraform.io/docs/language/values/variables.html#variable-definitions-tfvars-files`) file
- Packer project (chaining AMI builds) through a file

A new resource of kind `ami`:

* Source: allow to filter AMI on my AWS account to retrieve the latest version
* Condition: ensure that an AMI exists, given a set of filters
* Target: none (I don't see any use case immediatly)

Terraform and Packer have AMI filters that could do the job (almost) but using an updatecli process allows to force a static version and delegate the update process.

**Additional context**

- Command `describe-images` of the CLI `aws ec2`, to have an idea of what the AWS API allows: <https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html>
- Data source `aws_ami` used in terraform to retrieve an AMI: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami

Configuration example proposal:

```yml
---
title: "Bump ubuntu1804 arm Jenkins ec2 agent AMI"
sources:
  getLatestUbuntu1804ArmAMI:
    ## Only the latest ami should be returned (e.g. only 1 value)
    ## If multiple results are returned, then only the latest is kept but a warning is issued in the stoud with the returned values
    kind: ami
    name: Get the latest AMI ID
    spec:
      owners:
        - self # Default value
        - amazon # String value (common name of the publisher in the AWS Marketplace)
        - 12234567 # Int value (ID of the publisher in the AWS Marketplace)
      executable_users:
        - self # Default value
        - cloudbees@infra-admin # String of IAM user allowed to start an instance with the returned AMI
      filters: # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html#Filtering_Resources_CLI
        # Single value (easy case)
        - name: "root-device-type"
          values: "ebs"
        # Multiple values: string with comma as a separator
        - name: "instance-type"
          values: "m5.large,m5d.large"
        # Patterns
        - name: "name"
          values: "*ubuntu*1804*arm*"
      region: "us-east-1" # Region to use for the research. Can be used if different from the current AWS config
conditions:
  checkLatestUbuntu1804ArmAMI:
    ## Returns true if at least 1 AMI is returned by the research
    kind: ami
    name: Check if there is an ubuntu 1804 ARM image
    spec:
      owners:
        - self # Default value
        - amazon # String value (common name of the publisher in the AWS Marketplace)
        - 12234567 # Int value (ID of the publisher in the AWS Marketplace)
      executable_users:
        - self # Default value
        - cloudbees@infra-admin # String of IAM user allowed to start an instance with the returned AMI
      filters: # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html#Filtering_Resources_CLI
        # Single value (easy case)
        - name: "root-device-type"
          values: "ebs"
        # Multiple values: string with comma as a separator
        - name: "instance-type"
          values: "m5.large,m5d.large"
        # Patterns
        - name: "name"
          values: "*ubuntu*1804*amd*"
      region: "us-east-1" # Region to use for the research. Can be used if different from the current AWS config
...
```

olblak:
# Add Amazon AMI resource

- [x] [Documentation](https://github.com/updatecli/website/pull/134)

This pull request introduces to retrieve AMi using a source resource or to test if a AMI exist.

Fix #222 

## Test

```
  cp pkg/plugins/aws/ami/
  go test
```

! Required authentication 

### Settings

Available updatecli configuration
...

### Authentication

Different mechanisms can be used to authenticate with AWS API.

**Environment Variables**

updatecli retrieves credentials from the environment variables. Environment credentials never expire.

Environment variables used:

* Access Key ID: AWS_ACCESS_KEY_ID or AWS_ACCESS_KEY

* Secret Access Key: AWS_SECRET_ACCESS_KEY or AWS_SECRET_KEY

**Shared Credentials File**

By default updatecli looks for "AWS_SHARED_CREDENTIALS_FILE" env variable. If the 
env value is empty will default to the current user's home directory.
Linux/OSX: "$HOME/.aws/credentials"
Windows:   "%USERPROFILE%\.aws\credentials"

AWS Profile to extract credentials from the shared credentials file. If empty
will default to environment variable "AWS_PROFILE" or "default" if
the environment variable is also not set.

**Updatecli configuration**

We can provide credentials directly from a updatecli configuration by using the two following settings

.updatecli configuration
```
kind: aws/ami
name: resource to manipulate aws ami id
spec:
  access_key: "insert aws access key"
  secret_key: "insert aws secret key"
  filters:
    - name: "name"
      values: "image name "
```

### Filters
Different filters can be used by updatecli to know which ami-id to look after.

.updatecli configuration
```
kind: aws/ami
name: resource to manipulate aws ami id
spec:
  access_key: "insert aws access key"
  secret_key: "insert aws secret key"
  filters:
    - name: "name"
      values: "image name "
```

The filters.

* architecture - The image architecture (i386 | x86_64 | arm64).

* block-device-mapping.delete-on-termination - A Boolean value that indicates
whether the Amazon EBS volume is deleted on instance termination.

* block-device-mapping.device-name - The device name specified in the
block device mapping (for example, /dev/sdh or xvdh).

* block-device-mapping.snapshot-id - The ID of the snapshot used for the
EBS volume.

* block-device-mapping.volume-size - The volume size of the EBS volume,
in GiB.

* block-device-mapping.volume-type - The volume type of the EBS volume
(gp2 | io1 | io2 | st1 | sc1 | standard).

* block-device-mapping.encrypted - A Boolean that indicates whether the
EBS volume is encrypted.

* description - The description of the image (provided during image creation).

* ena-support - A Boolean that indicates whether enhanced networking with
ENA is enabled.

* hypervisor - The hypervisor type (ovm | xen).

* image-id - The ID of the image.

* image-type - The image type (machine | kernel | ramdisk).

* is-public - A Boolean that indicates whether the image is public.

* kernel-id - The kernel ID.

* manifest-location - The location of the image manifest.

* name - The name of the AMI (provided during image creation).

* owner-alias - The owner alias (amazon | aws-marketplace). The valid
aliases are defined in an Amazon-maintained list. This is not the AWS
account alias that can be set using the IAM console. We recommend that
you use the Owner request parameter instead of this filter.

* owner-id - The AWS account ID of the owner. We recommend that you use
the Owner request parameter instead of this filter.

* platform - The platform. To only list Windows-based AMIs, use windows.

* product-code - The product code.

* product-code.type - The type of the product code (devpay | marketplace).

* ramdisk-id - The RAM disk ID.

* root-device-name - The device name of the root device volume (for example,
/dev/sda1).

* root-device-type - The type of the root device volume (ebs | instance-store).

* state - The state of the image (available | pending | failed).

* state-reason-code - The reason code for the state change.

* state-reason-message - The message for the state change.

* sriov-net-support - A value of simple indicates that enhanced networking
with the Intel 82599 VF interface is enabled.

* tag:<key> - The key/value combination of a tag assigned to the resource.
Use the tag key in the filter name and the tag value as the filter value.
For example, to find all resources that have a tag with the key Owner
and the value TeamA, specify tag:Owner for the filter name and TeamA for
the filter value.

* tag-key - The key of a tag assigned to the resource. Use this filter
to find all resources assigned a tag with a specific key, regardless of
the tag value.

* virtualization-type - The virtualization type (paravirtual | hvm).


pkg/plugins/aws/ami/condition.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/scm"
+)
+
+// Condition test if a image matching specific filters exist.
```
dduportal, line 12:
```suggestion
// Condition tests if an image matching the specific filters exists.
```

pkg/plugins/aws/ami/condition.go
```diff
@@ -0,0 +1,74 @@
...
+				Values: source,
+			})
+		}
+
+	} else {
+		if len(a.Filters) == 0 {
+			logrus.Infof("\u2717 No AMI could be found as no AMI filters defined\n")
+			return false, nil
+		}
+	}
```
dduportal, line 41:
```suggestion
// It's an error if the upstream source is empty and the user does not provide any filter
// then it mean
if source == "" && len(a.Filters) == 0 {
	logrus.Infof("\u2717 No AMI could be found as no AMI filters defined\n")
	return false, nil
}

// Based on source information,
// we try to define a default image-id resource
// if not researched
isImageIDDefined := false
for i := 0; i < len(a.Filters); i++ {
	if strings.Compare(a.Filters[i].Name, "image-id") == 0 {
		isImageIDDefined = true
		break
	}

}

// Set image-id to source output if not yet defined
if !isImageIDDefined {
	a.Filters = append(a.Filters, Filter{
		Name:   "image-id",
		Values: source,
	})
}
```

As a general rule of thumb, try to return as early as possible. When you need to use a `else` then it might be a smell. The goal is to avoid overloading the code.

dduportal: (Check the gofmt 😅)

pkg/plugins/aws/ami/condition.go
```diff
@@ -0,0 +1,74 @@
...
+
+	svc, errs := a.Init()
+
+	for _, err := range errs {
+		logrus.Error(err)
+	}
+	if len(errs) > 0 {
+		return false, errors.New("Too many errors")
```
dduportal, line 49: What is the meaning of this error?

pkg/plugins/aws/ami/helpers.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/sirupsen/logrus"
+)
+
+// GetLatestAmiID query the AWS API to return the newest AMI image id
```
dduportal, line 12:
```suggestion
// GetLatestAmiID queries the AWS API to return the latest AMI id with the provided filters 
```

pkg/plugins/aws/ami/helpers.go
```diff
@@ -0,0 +1,65 @@
...
+func (a *AMI) GetLatestAmiID(svc ec2iface.EC2API) (string, error) {
```
dduportal, line 13:
```suggestion
func (a *AMI) getLatestAmiID(svc ec2iface.EC2API) (string, error) {
```
this method could be private

. Warning: you'll have to update the references across sources :)

pkg/plugins/aws/ami/condition.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	result, err := a.GetLatestAmiID(svc)
+
+	if err != nil {
+		return false, err
+	}
+
+	if len(result) > 0 {
```
dduportal, line 58: Question: if there are no result found, should'nt the method raise an error?

pkg/plugins/aws/ami/condition.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return false, nil
+}
+
+// ConditionFromSCM is a placeholder to validate the condition interface
+func (a *AMI) ConditionFromSCM(source string, scm scm.Scm) (bool, error) {
+
+	fmt.Printf("\u2717 Condition with SCM is not supported, please remove the scm block \n")
+
+	return false, errors.New("condition with SCM is not supported")
```
dduportal, line 73:
```suggestion
	fmt.Printf("\u2717 Condition with SCM is not supported, please remove the scm block \n")

	return false, errors.New("condition with SCM is not supported")
```

You might want to define a custom error and raise it (instead of printing twice and manipulating strings)

pkg/plugins/aws/ami/main.go
```diff
@@ -0,0 +1,107 @@
...
+	DryRun     bool
+	ec2Filters []*ec2.Filter
+}
+
+// Filter represents the updatecli configuration to define AMI filter
+// This datatype need to be convert the ec2.Filter.
+type Filter struct {
+	Name   string
+	Values string
+}
```
dduportal, line 31: you might want to define 2 types:
- amiSpec, a private type used to represent the "data" parsed from yaml.
  - this type should manage the yaml parsing and the validation of the data (raising error when there are semantic errors such as valid yaml but missing or empty directives)
- AMI, the top level type that has a private object `spec` of type `amispec`. This object would expose the public methode such as `Condition` , etc. and of course should have the private attributes related to AWS api such as `ec2Filters`.
  - The `Init` method (I like to use `New)  would be changed to have a spec object (from the YAML parsing) and would respond a fully initialized `AWS` object.

pkg/plugins/aws/ami/main.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return str
+}
+
+// Init run basic parameter initiation
+func (a *AMI) Init() (svc *ec2.EC2, errs []error) {
+	if len(a.Region) == 0 {
+		a.Region = "us-east-1"
+	}
```
dduportal, line 57: I would raise an error instead (as the "default" valuation of the region is the responsibility of the AWS configuration system)

pkg/plugins/aws/ami/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if len(a.Endpoint) == 0 {
+		a.Endpoint = fmt.Sprintf("https://ec2.%s.amazonaws.com", a.Region)
+	}
```
dduportal, line 61: Same here

pkg/plugins/aws/ami/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Init ec2Filters
+	for i := 0; i < len(a.Filters); i++ {
+		filter := ec2.Filter{
+			Name: func(input string) *string {
+				output := strings.ToLower(input)
+				return &output
+			}(a.Filters[i].Name),
+			Values: values(strings.Split(a.Filters[i].Values, ",")),
```
dduportal, line 80: It's not clear that `values` is a multi-valued field. WDYT about defining the rule "if it is a string, then it must be string passed "as it" itherwise the yaml should be a list of strings ?

pkg/plugins/aws/ami/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Convert []string to []*string as required by the ec2.Filter values field
+	values := func(input []string) []*string {
+		var output []*string
+		for i := range input {
+			s := input[i]
+			output = append(output, &s)
+		}
+		return output
+	}
```
dduportal, line 71: To avoid this conversion, you might be interested in https://gist.github.com/simeji/945165da059d2f4ef4bf#file-descrive_ec2_instances-go-L23-L25

pkg/plugins/aws/ami/condition.go
olblak: We are  connected, I was thinking exactly the same quite recently

pkg/plugins/aws/ami/main.go
olblak: I love this suggestion very much

pkg/plugins/aws/ami/main.go
olblak: As discussed early today, I propose printing a message that we fallback by default to us-east-1

pkg/plugins/aws/ami/main.go
olblak: In the current state, this configuration is useless as it strongly depends on the region

pkg/plugins/aws/ami/main.go
olblak: I think It would be better to document that because most of the time the filters accept a single such as name, platform, etc..

pkg/plugins/aws/ami/main.go
olblak: Thanks for the suggestion, I realised after some refactoring that I should have used to functions

pkg/plugins/aws/ami/main.go
dduportal: I'll document this on the doc PR

pkg/plugins/aws/ami/condition.go
olblak: That's a good question that I asked myself a while ago and I decided that it wasn't an error because I want to know if an image id exist or not, To me an error could using a wrong endpoint

pkg/plugins/aws/ami/condition.go
olblak: It think this should be an improvement done outside this pull request as it would cover every other resources

pkg/plugins/aws/ami/condition.go
olblak: I improved that part :)

pkg/plugins/aws/ami/main.go
olblak: Implemented

pkg/plugins/aws/ami/condition.go
dduportal: Not really sure (as it would be done here at least), but it's not blocking so I'm resolving the convo.

dduportal (changes requested):
Played with the following example:

```
---
title: Bump agent templates version on all controllers
sources:
  packerImageVersion:
    kind: githubRelease
    spec:
      owner: "jenkins-infra"
      repository: "packer-images"
  getLatestUbuntuAgentAMIAmd64:
    kind: aws/ami
    depends_on:
      - packerImageVersion
    spec:
      region: us-east-2
      filters:
        - name: "name"
          values: "jenkins-agent-ubuntu-20-amd64-*"
        - name: "tag:build_type"
          values: "prod"
        - name: "tag:version"
          values: '{{ source "packerImageVersion" }}'
  getLatestWindowsAgentAMIAmd64:
    kind: aws/ami
    depends_on:
      - packerImageVersion
    spec:
      region: us-east-2
      filters:
        - name: "name"
          values: "jenkins-agent-windows-2019-amd64-*"
        - name: "tag:build_type"
          values: "prod"
        - name: "tag:version"
          values: '{{ source "packerImageVersion" }}'
  getLatestUbuntuAgentAMIArm64:
    kind: aws/ami
    depends_on:
      - packerImageVersion
    spec:
      region: us-east-2
      filters:
        - name: "name"
          values: "jenkins-agent-ubuntu-20-arm64-*"
        - name: "tag:build_type"
          values: "prod"
        - name: "tag:version"
          values: '{{ source "packerImageVersion" }}'

targets:
  setUbuntuAgentAMIAmd64:
    name: "Bump AMI ID for Ubuntu AMD64 agents"
    kind: shell
    sourceID: getLatestUbuntuAgentAMIAmd64
    spec:
      command: echo ubuntu-amd
  setUbuntuAgentAMIArm64:
    name: "Bump AMI ID for Ubuntu ARM64 agents"
    kind: shell
    sourceID: getLatestUbuntuAgentAMIArm64
    spec:
      command: echo ubuntu-arm
  setWindowsAgentAMIAmd64:
    name: "Bump AMI ID for Windows ARM64 agents"
    kind: shell
    sourceID: getLatestWindowsAgentAMIAmd64
    spec:
      command: echo windows-amd
```



- First of all it works flawlessly 👏 🥳 it's really fast, and used my (weird) AWS config
- The only issue I got is how to handle when there are multiple results to the query.

On my current use case, i'm retrieving the latest image by sorting on the CreationDate attribute, using the `--query` flag of the aws CLI:

```
aws ec2 describe-images \
  --owners 200564066411 `# Owner ID for AWS account cloudbees-jenkins` \
  --region=us-east-2 `#AWS region where ci.jenkins.io runs its EC2 agents` \
  --filters "Name=name,Values=jenkins-agent-${operating_system}-${cpu_architecture}-*" `# Search by name, with a pattern for the end (timestamps)` \
    "Name=tag:build_type,Values=prod" `# Only retrieve production ready AMIs` \
    "Name=tag:version,Values=${image_version}" `# image version is a semantic version v2 string` \
  --query 'sort_by(Images, &CreationDate)[-1].[ImageId]' `# Get the most recent one by sorting by creation date and get the latest and extract its AMI ID` \
  --output text `# We don't want JSON`
```

olblak:
@dduportal I had a new spec key "sortBy: creationDateAsc" that you can use to order the list of AMI ID, the resource still returns the latest AMI found but we can now decide to sort the result by ascending or descending based on creation date.

Sortby can now accept the following values ["","creationDateAsc", "creationDateDesc"] with default set to ""
So 
```
    getLatestWindowsAgentAMIAmd64:
      kind: aws/ami
      spec:
        region: us-east-2
        sortBy: creationDateAsc
        filters:
          - name: "name"
            values: "jenkins-agent-windows-2019-amd64-*"
          - name: "tag:build_type"
            values: "prod"
          - name: "tag:version"
            values: '0.5.0'

```

is equivalent to 
```
aws ec2 describe-images --owners 200564066411 --region=us-east-2 --filters 'Name=name,Values=jenkins-agent-windows-2019-amd64-*' Name=tag:build_type,Values=prod Name=tag:version,Values=0.5.0 --query 'sort_by(Images, &CreationDate)[-1].ImageId'

```

dduportal:
@olblak that sounds really cool! thanks for the extra effort!

dduportal (approved):
🚀 Works really well!

PR #304, opened 2021-10-07
Feat [Shell]: Add an attribute which suppress the source appending

### LINKED ISSUE #288: Feature Request [Resource "shell"] Allow disabling adding the source as line argument

The resource "shell", when used for condition or target, execute a command with the associated source output appended by default.

But sometimes, when the source is not needed (when using templating for instance), this "last argument" makes the command harder to read/write, or forces the user to create a separated shell script.

Consider this example:

```yaml
sources:
  mavenVersion:
    kind: githubRelease
    name: Get the latest Maven version
    spec:
      owner: "apache"
      repository: "maven"
      versionFilter:
        kind: latest
    transformers:
      - trimPrefix: "maven-"
 
conditions:
  checkIfMavenReleaseIsAvailable:
    kind: shell
    name: check if the found release is available for download
    sourceID: mavenVersion
    depends_on:
      - mavenVersion
    spec:
      command: curl --connect-timeout 5 --location --head --fail --silent --show-error https://archive.apache.org/dist/maven/maven-3/{{ source "mavenVersion" }}/binaries/apache-maven-{{ source "mavenVersion" }}-bin.tar.gz 

targets:
  showMavenVersion:
    name: Prints the latest Maven available version
    sourceID: mavenVersion
    kind: shell
    spec:
      command: echo
```

which results in the following output (truncated):

```text
SOURCE:
=======

⚠ No GitHub Release found. As fallback Looking at published git tags
Searching for version matching pattern "latest"
✔ Github Release version "maven-3.8.3" found matching pattern "latest"


CHANGELOG:
==========
No Github Release found for maven-3.8.3 on https://github.com/apache/maven


CONDITIONS:
===========

The shell 🐚 command "curl --connect-timeout 5 --location --head --fail --silent --show-error https://archive.apache.org/dist/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.tar.gz 3.8.3" failed with an exit code of 28 and the following messages: 
stderr=

stdout=
curl: (22) The requested URL returned error: 404 Not Found
curl: (28) Connection timed out after 5004 milliseconds


=============================

REPORTS:


✗ UPDATECLI.YAML
        Sources:
                ✔  Get the latest Maven version(githubRelease)
        Condition:
                ✗  check if the found release is available for download(shell)
        Target:
                  ()



ERROR: ✗ 1/1 job(s) failed
ERROR: command failed
```

I would like to be able to specify an optional attribute `appendSource` on the condition or the target.
By default, this attributes should be set to "true" to keep the current behavior.
When set to false, then the source should not be appended.

- Create a shell script to use the source's value instead of using templating
- Do not set the a condition 😅

**Additional context**
N.A.

jplayout:
# Add an attribute which suppress the source appending

Fix #288 

I added an boolean attribute named suppressSource which removes the source added to the command


olblak:
Thank you a lot @jplayout  for the PR.
I am facing the same issue on other "resources" such as "aws/ami" and I am wondering if it wouldn't be better to have something generic at the condition and at target level such as (not at the spec level like you are suggesting). 

I am thinking of something like:

```
conditions:
    condition:
        disableSourceInput: true (default to false)
        kind: shell
        spec:
            command: xxx
```
And we could just have a condition that test if disableSourceInput is set to true then we set the source to an empty string
[here](https://github.com/updatecli/updatecli/blob/cfb5f8e50b93f30ed5aa74dead8fd2e309749dd3/pkg/core/engine/condition/main.go#L45)

olblak:
weird that the pr got merged anyway you made a good suggestion, I need to think a little bit about my suggestion and both are not incompatible. So let's move forward. I would just target a new minor version instead of a patch version

PR #330, opened 2021-10-21
feat [conditions] use lower case for new attribute disablesourceinput

dduportal:
As per https://github.com/updatecli/updatecli/pull/321#discussion_r733580775, this PR does damage control on the naming convention by setting the attribute to lower case.

The doc. PR at https://github.com/updatecli/website/pull/154 had been updated


lemeurherve:
@dduportal make sense :)

Let's add a 1337case to the mix for the next one then!

<details>
<summary></summary>

image

</details>

PR #356, opened 2021-11-11
Add transformer findSubMatch

### LINKED ISSUE #280: Add new transformers "findSubmatch"

At this stage there is no way to extract submatch using regex filter.
In the context of transformers it's useful to modify information retrieved from a "source".
The only regex transformer is "find" which return the string that match the regex
So the regex `go-(\d+.\d+)*` on the string "go-1.17.3" will return `go-1.17.3` but we need a way to return `1.17`

I am interested to have a new "transformer" named "findSubmatch" which would use the golang function [findStringSubmatch](https://pkg.go.dev/regexp#Regexp.FindStringSubmatch) to return the submatch

```
   name: Upate go.mod
    kind: file
    spec:
      file: go.mod
    transformers:
      - findSubmatch: 'go (\d+.\d+)'
```

should return 1.16

Another option would be to change the transformer "find" to use "FindStringSubmatch" instead of "FindString". The same result can be achieve with both function but that would have the consequence to break existing updatecli configuration which is why  I am proposing to not modify "find" until we introduce the next major version of updatecli.

**Additional context**
* The code must be implemented in the package [core/transformer](https://github.com/updatecli/updatecli/blob/main/pkg/core/transformer/main.go)
* Once this issue is solved, we will be able to finalize the [PR273](https://github.com/updatecli/updatecli/pull/273)

olblak:
# Add transformer findSubMatch

Fix #280 

The purpose of this transformer is to use a regex to extract a substring

## Test

```
  cp pkg/core/transformer
  go test
```


dduportal (commented):
Great job ! LGTM

Issue #359, opened 2021-11-11
[yaml] attribute 'value' is ignored by targets


dduportal:
**Describe the bug**

With a target of kind `yaml`, specifying the attribute `value` as no effect. The input source value is always used.

**To Reproduce**

```yaml
# test.yaml
github:
  user: olblak
  repository: charts
```

```yaml
---
title: Show a set of yaml resources as a generic example
sources:
  getGithubUser:
    kind: yaml
    spec:
      file: test.yaml
      key: "github.user"
targets:
  setGitHubRepository:
    kind: yaml
    spec:
      file: test.yaml
      key: "github.repository"
      value: "charts"
```

```shell
# ...
SOURCES
=======

getGithubUser
-------------
✔ Value 'olblak' found for key github.user in the yaml file test.yaml


TARGETS
========

setGitHubRepository
-------------------

**Dry Run enabled**

✔ Key 'github.repository', from file 'test.yaml', was updated from 'charts' to 'olblak'

=============================
# ...
```

**Expected behavior**

I expect that the target sets the specified key to the specified content, instead of the one from the default source value.

e.g.

```shell
# ...
SOURCES
=======

getGithubUser
-------------
✔ Value 'olblak' found for key github.user in the yaml file test.yaml


TARGETS
========

setGitHubRepository
-------------------

**Dry Run enabled**

✔ Key 'github.repository', from file 'test.yaml', is already set to 'charts'

=============================
# ...
```
```

**Screenshots**
N.A.

**Version (please complete the following information):**
 - OS: macOS
 - Version 0.13.1

**Additional context**
N.A.

Issue #358, opened 2021-11-11
[yaml] condition with a source input value fails to validate


dduportal:
**Describe the bug**

When using a yaml condition without the attribute `value` defined, the conditions fails even if the input source value matches the value of the specified YAML key.

The documentation at https://www.updatecli.io/docs/resources/yaml does not allow to conclude if it should work or not.

**To Reproduce**

```yaml
# test.yaml
github:
  user: asterix
```

```yaml
# updatecli.yaml
title: Check a YAML file
sources:
  default:
    kind: yaml
    spec:
      file: test.yaml
      key: "github.user"
conditions:
  checkGithubUsername:
    kind: yaml
    sourceID: default
    spec:
      file: test.yaml
      key: "github.user"
``` 

```shell
# ..
#####################
# CHECK A YAML FILE #
#####################


SOURCES
=======

default
-------
✔ Value 'asterix' found for key github.user in the yaml file test.yaml


CONDITIONS:
===========

checkGithubUsername
-------------------
✗ Key 'github.user', from file 'test.yaml', is incorrectly set to asterix and should be '

✗ condition not met, skipping pipeline

=============================
# ...
```

**Expected behavior**

```shell
SOURCES
=======

default
-------
✔ Value 'asterix' found for key github.user in the yaml file test.yaml


CONDITIONS:
===========

checkGithubUsername
-------------------
✔ Key 'github.user', from file 'test.yaml', is correctly set to 'asterix'

=============================
```

**Screenshots**
N.A.

**Version (please complete the following information):**
 - OS: macOs
 - Version 0.13.1

**Additional context**
N.A.

PR #360, opened 2021-11-11
fix(yaml) Allow specifying file path as an URL

### LINKED ISSUE #357: [yaml] error when the "file" key is an URL
**Describe the bug**

As per the documentation (https://www.updatecli.io/docs/resources/yaml/#_file) the attribute `file` of YAML resource can be either a file path, or an "URL"-like (with a prefix scheme).

But the URL-like values are resulting on errors, at least on a YAML source.

**To Reproduce**

The following snippet fails when for both `file://` and `https://` URL schemes:

```yaml
---
title: Show a set of yaml resources as a generic example
sources:
  default:
    kind: yaml
    spec:
      file: examples/values.yaml
      key: "github.user"
  fromExplicitFile:
    kind: yaml
    spec:
      file: file://examples/values.yaml
      key: "github.user"
  fromUrl:
    kind: yaml
    spec:
      file: https://raw.githubusercontent.com/updatecli/updatecli/dc90759914718beee17101c693911088cfbf6deb/examples/values.yaml
      key: "github.user"
```

```shell
#####################################################
# SHOW A SET OF YAML RESOURCES AS A GENERIC EXAMPLE #
#####################################################


SOURCES
=======

fromUrl
-------
ERROR: ✗ open /Users/dadou/workspace/updatecli/updatecli/https:/raw.githubusercontent.com/updatecli/updatecli/dc90759914718beee17101c693911088cfbf6deb/examples/values.yaml: no such file or directory

fromExplicitFile
----------------
ERROR: ✗ open /Users/dadou/workspace/updatecli/updatecli/file:/examples/values.yaml: no such file or directory

default
-------
✔ Value 'olblak' found for key github.user in the yaml file examples/values.yaml
```

**Expected behavior**

- I would expect the case with `file://` to work as expected (e.g. the scheme prefix should be removed)
- For the URL like, I'm not sure if it should be supported, or not. WDYT @olblak ?

**Screenshots**
N.A.

 - OS: macOS
 - Version 0.13.1

**Additional context**
N.A.

dduportal:
Fix #357 

This PRs allow specifying Yaml file path as URLs.

## Test

```
go build -o ./dist/updatecli && ./dist/updatecli diff --config ./examples/updateCli.generic/yaml.yaml
```

## Additional Information

- Introduced a `New` method to create a Yaml object
- Separate `YamlSpec` from `Yaml` objects to allow injection for unit testing/mocking + easier feature upgrades
- Use the same strategy as the file resource by adding a `Read()` method
- Updated examples to provide more cases


pkg/plugins/yaml/main.go
```diff
@@ -15,13 +18,61 @@ var (
...
+	File string
+	Key  string
+	// Deprecated: use File instead
 	Path  string
 	Value string
 }
 
+// Yaml defines a resource of type "yaml"
+type Yaml struct {
+	Spec             YamlSpec
```
olblak, line 31: Any reason to make public Spec instead of spec?

olblak (approved):
Thank you for the cleanup, excepted a  minor question, everything look good to me

pkg/plugins/yaml/main.go
dduportal: Because of the resource of type `helm` which uses the `Yaml` types, as you can see in this change (diff of the current PR): https://github.com/updatecli/updatecli/pull/360/files#diff-beb8c6d04ce8efe2a8e014a3137114c2d8e1f9dd850ed2d5df13e40e3f685446R36

I chose to not work on decoupling both (like with `text`  / `file` `yaml`) in the context of this PR since it's a bug fix (and not a feature).

If it's ok for you, after fixing this bug,  #358 and #359 , I would like to factorize the logging logic of the yaml module (as in #354 for shell) , that would be a perfect spot for decoupling.

dduportal:
@olblak if the proposal is ok for you, can I let you approve and merge?

PR #382, opened 2021-11-17
feat(file) allow condition to only check for file existence

### LINKED ISSUE #290: Feature Request [Resource "File"] allow condition to only check for existence of a file (and not the value)

In the spirit of #289, with the same use case as #288 below:

```yaml
sources:
  mavenVersion:
    kind: githubRelease
    name: Get the latest Maven version
    spec:
      owner: "apache"
      repository: "maven"
      versionFilter:
        kind: latest
    transformers:
      - trimPrefix: "maven-"
 
conditions:
  checkIfMavenReleaseIsAvailable:
    kind: shell
    name: check if the found release is available for download
    sourceID: mavenVersion
    depends_on:
      - mavenVersion
    spec:
      command: curl --connect-timeout 5 --location --head --fail --silent --show-error https://archive.apache.org/dist/maven/maven-3/{{ source "mavenVersion" }}/binaries/apache-maven-{{ source "mavenVersion" }}-bin.tar.gz 

targets:
  showMavenVersion:
    name: Prints the latest Maven available version
    sourceID: mavenVersion
    kind: shell
    spec:
      command: echo
```

I would like to check for the existence of the remote file at the provided URL, or on my filesystem.

Adding a new attribute `existenceOnly` only available on the condition of kind "file", that would ignore the value and only check for the existence of the file.

- Using a shell condition with a curl/wget command (or test -f)
- Implements #268 to check for a remote (URL) file

**Additional context**
N.A.

dduportal:
Fix #290 

No need for a specific atttribute: only have to disable the source input in condition.
2 examples added in the PR to show how to use it + test units to cover additional cases.

It also add improved output for end user:

- info message for each condition to report if it passed, did not pass, or errored (easier to read when multiple conditions)
- more debug messages on the text package and file condition

## Test

```shell
go build -o ./dist/updatecli
./dist/updatecli diff --config ./examples/updateCli.generic/file.yaml
```

### DIFF
--- a/pkg/core/text/main.go
+++ b/pkg/core/text/main.go
@@ -37,7 +37,7 @@ func readFromURL(url string, line int) (string, error) {
 	}
 	logrus.Debugf("HTTP GET to %q got a response with code %d", url, resp.StatusCode)
 	if resp.StatusCode > 399 {
-		logrus.Debugf("Status code for URL %q is not in the 1XX, 2XX or 3XX ranges: %d", url, resp.StatusCode)
+		logrus.Debugf("HTTP return code %q for URL %q", resp.StatusCode, url)
 		return "", fmt.Errorf("URL %q not found or in error", url)
 	}
 
--- a/pkg/core/text/main.go
+++ b/pkg/core/text/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hexops/gotextdiff"
 	"github.com/hexops/gotextdiff/myers"
 	"github.com/hexops/gotextdiff/span"
+	"github.com/sirupsen/logrus"
 )
 
 type TextRetriever interface {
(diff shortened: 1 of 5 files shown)


pkg/core/text/main.go
```diff
@@ -34,14 +35,22 @@ func readFromURL(url string, line int) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	logrus.Debugf("HTTP GET to %q got a response with code %d", url, resp.StatusCode)
+	if resp.StatusCode > 399 {
+		logrus.Debugf("Status code for URL %q is not in the 1XX, 2XX or 3XX ranges: %d", url, resp.StatusCode)
```
olblak, line 40: I think, I would make the message simpler but it's a minor suggestion

```suggestion
		logrus.Debugf("HTTP return code %q for URL %q", resp.StatusCode, url)
```

pkg/plugins/file/condition.go
```diff
@@ -8,23 +8,40 @@ import (
...
+
+func (f *File) checkCondition(source string) (bool, error) {
+	passing, err := f.condition(source)
+	if err != nil {
+		logrus.Infof("%s Condition on file %q errored", result.FAILURE, f.spec.File)
+	} else {
+		if passing {
+			logrus.Infof("%s Condition on file %q passed", result.SUCCESS, f.spec.File)
+		} else {
+			logrus.Infof("%s Condition on file %q did not pass", result.FAILURE, f.spec.File)
```
olblak, line 40: What would you think to use a different message than failiure? 

```suggestion
			logrus.Infof("%s Condition on file %q did not pass", result.ATTENTION, f.spec.File)
```
"\u26A0"

olblak (approved):
Excepted minor comment, this pull request looks great to me. Thanks @dduportal for the improvement

olblak:
Feel free to merge once you are ready

pkg/plugins/file/condition.go
dduportal: It feels weird to use « attention » since it is the symbol used when a target changes a file. Might be something to reconsider overall: choosing 2 symbols for failure and not passing, and 2 other symbols for « success » and « changed »

dduportal: But the fact of sharing the same error symbol is why I took care of having 2 really different text messages to easily discern failure from condition not passing. Maybe we could remove the « cross » from failure since it generate a clear error message in red.

PR #386, opened 2021-11-20
fix(log messages) homogenize the emojis used for user output messages

dduportal:
It's a "short term" fix to ensure that the log messages have the same behavior and all use the same emoji.
It also improve code readability as it's easier to understand the constant `result.ATTENTION` than `\u26A0`.

Please note the following changes:
- As we use the package name `result`, when a variable had a conflicting name then it had been changes (3 occurends, 2 are in the AMI package)
- There was slight differences of emojis (2416/2415 for instance) that have been simplified to all use the same from the `result` package
- Sources only have `FAILURE` or `SUCCESS` states
- Same for conditions: only `FAILURE` or `SUCCESS` states
- For targets, `FAILURE` indicates a real error, while `SUCCESS` is used when nothing had been changed, and `ATTENTION` when a change is expected (diff) or applied (apply).

Please note that a long term fix is to implement a custom `logger` package that would provide explicit method such as `ValidationErrorLog()`, `SuccessErrorLog()`, `ContentChangedLog()`, `FailureLog()` etc. to move the multiple usages of emojis to a central place, and get rid of the `result` package.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -38,7 +38,7 @@ func Execute() {
 	logrus.SetFormatter(log.NewTextFormat())
 
 	if err := rootCmd.Execute(); err != nil {
-		logrus.Errorf("\u26A0 %s", err)
+		logrus.Errorf("%s %s", result.FAILURE, err)
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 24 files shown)


olblak (approved):
Thank you.
Iit goes in the right direction.
The next step is to have a specific package for UI

PR #387, opened 2021-11-21
fix(githubRelease) retrieve changelog based on the original release/tag instead of the transformed one

### LINKED ISSUE #244: semver filtering affect changelog output
**Describe the bug**
It seems that when we filter a version retrieved from a githubRelease, using the filter "semver" retrieve the "v" in front of the version and so we can't retrieve the changelog anymore

**To Reproduce**
```
  helmfile:
    kind: githubRelease
    name: Get the latest helmfile version
    spec:
      owner: "gohugoio"
      repository: "hugo"
      #versionFilter:
      #  kind: semver
      #  pattern: "~0"
    # TrimPrefix so we can retrieve the changelog
    transformers:
      - trimPrefix: "v" 
```

**Expected behavior**
We should keep track of the real version retrieved from a github release

```
  helmfile:
    kind: githubRelease
    name: Get the latest helmfile version
    spec:
      owner: "gohugoio"
      repository: "hugo"
      #versionFilter:
      #  kind: semver
      #  pattern: "~0"
```

 - OS: linux
 - Version: v0.4.0

**Additional context**

dduportal:
Fix #244 

This PR refactored the `version` package by including the `semver` code to allow storing a couple of string for the found versions (parsed and original).

It allows the githubrelease provider to store the "original name" of the found release (or tag) and retrieve the changelog independently of the transformed source (through semver filter, or text transformers for instance).

## Test

The generic example for `githubRelease` is updated to provide different cases. One if these cases is failing to retrieve the changelog with updatecli `0.15.0` but it's working with this code:

```shell
go build -o ./dist/updatecli  && ./dist/updatecli --debug diff -c ./examples/updatecli.generic/githubRelease.yaml
```


examples/updatecli.generic/githubRelease.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
lemeurherve, line 1: How about taking back the top comment block of the old template? Or is this content stored elsewhere?

lemeurherve (approved):

examples/updatecli.generic/githubRelease.yaml
dduportal: the former top comment block was "telling a story" about the former example (including the conditions and targets that are not supported by githubrelease).

The goal of the change is to provide a generic example specific to githubrelease.

Also, https://www.updatecli.io/docs/resources/github_release/#_example has plenty of the same content and explanation :)

olblak (approved):

PR #396, opened 2021-11-25
fix: skip sources when their parent source failed

### LINKED ISSUE #285: Bug: source executed while their parent source failed
**Describe the bug**

In a scenario with sources depending on each other explictly (e.g. using the directive `depends_on`),

if the upstream source failed then the downtsream sources are still executed.
It leads to pipelines trying to execute with empty values, which is hard to catch at first sight.

**To Reproduce**

With this file:

```yml
# updateCli.yaml
sources:
  parentSource:
    kind: shell
    spec:
      command: /nonexistingcommandtofail
  childSource:
    kind: shell
    depends_on:
      - parentSource
    spec:
      command: echo CHILD-{{ source "parentSource" }}
target:
  printResult:
    kind: shell
    sourceID: childSource
    spec:
      command: echo
```

The command `updatecli diff` reports 1 job failure as expected, BUT the source `childSource` had been executed depsite the source `parentSource` failing:

```shell
SOURCE:
=======

ERROR: ✗ fork/exec /nonexistingcommandtofail: no such file or directory


SOURCE:
=======

✔ The shell 🐚 command "echo CHILD-{{ source \"parentSource\" }}" ran successfully and retrieved the following source value: "CHILD-{{ source \"parentSource\" }}"
```

**Expected behavior**

The source `childSource` should not be executed and the command line output should show it as "error because upstream".

**Additional context**
N.A.

dduportal:
Fix #285

When a source, referenced through a template function ``{{ source `<source id>` }}`` on another source, is failing,
then the "children" sources are now skipped with an error message.

No worry for conditions: if any source fail, then the condition are never run.

## Test

```shell
$ cat .tmp/updatecli.yaml
sources:
  parentSource:
    kind: shell
    spec:
      command: ls /titi
  childSource:
    kind: shell
    depends_on:
      - parentSource
    spec:
      command: echo CHILD-{{ source "parentSource" }}
target:
  printResult:
    kind: shell
    sourceID: childSource
    spec:
      command: echo

$ go build -o ./dist/updatecli && ./dist/updatecli --debug diff -c .tmp/updatecli.yaml
```

example of output:

```
SOURCES
=======

parentSource
------------
The shell 🐚 command "ls /titi" exited on error (exit code 1) with the following output:
----
----

command stderr output was:
----
ls: /titi: No such file or directory

----
ERROR: ✗ Shell command exited on error.
Pipeline "UPDATECLI.YAML" failed
Skipping due to:
        "sources stage:\t\"template: cfg:25:35: executing \\\"cfg\\\" at <source \\\"parentSource\\\">: error calling source: Parent source \\\"parentSource\\\" failed\"\n"
sources stage:  "template: cfg:25:35: executing \"cfg\" at <source \"parentSource\">: error calling source: Parent source \"parentSource\" failed"

=============================

REPORTS:


✗ UPDATECLI.YAML:
        Sources:
                - [childSource] (shell)
                ✗ [parentSource] (shell)
        Target:
```

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/updatecli/updatecli/pkg/core/engine/condition"
 	"github.com/updatecli/updatecli/pkg/core/engine/source"
 	"github.com/updatecli/updatecli/pkg/core/engine/target"
+	"github.com/updatecli/updatecli/pkg/core/result"
 	"gopkg.in/yaml.v3"
 )
 
@@ -253,26 +254,27 @@ func (config *Config) Update(data interface{}) (err error) {
 				Retrieve the value of a third location key from
 				the updatecli contex.
 				It returns an error if a key doesn't exist
-				It returns {{ pipeline "<key>" }} if a key exist but still set to zero value,
+				It returns {{ source "<key>" }} if a key exist but still set to zero value,
 				then we assume that the value will be set later in the run.
 				Otherwise it returns the value.
 				This func is design to constantly reevaluate if a configuration changed
 			*/
 
-			val, err := getFieldValueByQuery(
-				data, []string{"Sources", s, "Output"})
-
+			sourceResult, err := getFieldValueByQuery(data, []string{"Sources", s, "Result"})
 			if err != nil {
 				return "", err
 			}
 
-			if len(val) > 0 {
-				return val, nil
+			switch sourceResult {
+			case result.SUCCESS:
+				return getFieldValueByQuery(data, []string{"Sources", s, "Output"})
+			case result.FAILURE:
+				return "", fmt.Errorf("Parent source %q failed", s)
+			// If the result of the parent source execution is not SUCCESS or FAILURE, then it means it was either skipped or not already run.
+			// In this case, the function is return "as it" (literrally) to allow retry later (on a second configuration iteration)
+			default:
+				return fmt.Sprintf("{{ source %q }}", s), nil
 			}
-			// If we couldn't find a value, then we return the function so we can retry
-			// later on.
-			return fmt.Sprintf("{{ source %q }}", s), nil
-
 		},
 	}
 
--- a/pkg/core/pipeline/sources.go
+++ b/pkg/core/pipeline/sources.go
@@ -33,7 +33,16 @@ func (p *Pipeline) RunSources() error {
 		logrus.Infof("\n%s\n", id)
 		logrus.Infof("%s\n", strings.Repeat("-", len(id)))
 
-		err = source.Run()
+		shouldRunSource := true
+		for _, parentSource := range source.Config.DependsOn {
+			if p.Sources[parentSource].Result != result.SUCCESS {
+				logrus.Warningf("Parent source[%q] did not succeed. Skipping execution of the source[%q]", parentSource, id)
+				shouldRunSource = false
+			}
+		}
+		if shouldRunSource {
+			err = source.Run()
+		}
 		rpt.Result = source.Result
 
 		if len(source.Changelog) > 0 {
--- a/pkg/core/config/main_test.go
+++ b/pkg/core/config/main_test.go
@@ -8,11 +8,13 @@ import (
 	"github.com/updatecli/updatecli/pkg/core/engine/condition"
 	"github.com/updatecli/updatecli/pkg/core/engine/source"
 	"github.com/updatecli/updatecli/pkg/core/engine/target"
+	"github.com/updatecli/updatecli/pkg/core/result"
 )
 
 // Mocking the context package
 type mockSourceContext struct {
 	Output string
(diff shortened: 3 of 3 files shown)


pkg/core/config/main.go
```diff
@@ -242,26 +243,25 @@ func (config *Config) Update(data interface{}) (err error) {
...
-				return val, nil
+			switch sourceResult {
+			case result.SUCCESS:
+				return getFieldValueByQuery(data, []string{"Sources", s, "Output"})
+			case result.FAILURE:
+				return "", fmt.Errorf("Parent source %q failed", s)
+			default:
+				return fmt.Sprintf("{{ source %q }}", s), nil
 			}
-			// If we couldn't find a value, then we return the function so we can retry
```
olblak, line 261: @dduportal Any reason to have to drop the comment that explains the default value?

olblak:
I think you are solving half of the problem here. I see two scenarios that we need to cover

1. Source B that depends on Source A with templating involved which is solved by this PR 🎉 
2. Source B that depends on Source A without templating.

In the second scenario, I think before executing a source, we need to validate that sources listed in the dependency list succeeded.
Just before this [line](https://github.com/updatecli/updatecli/blob/1efbf6006d72d227480de617fb675aec8ba7452a/pkg/core/pipeline/sources.go#L36)

Something like 

```
for _, dependedSource := range s.dependsOn {
    if p.Sources[dependencySource].Result == result.Failure {
    // skipping source as depended source already failed
   }
}
```

olblak (changes requested):
I don't think that the pull request in the current stage, fully solve the problem

pkg/core/config/main.go
dduportal: Fixed, I've added a comment that explain the "default" case WDYT?

dduportal:
> I think you are solving half of the problem here. I see two scenarios that we need to cover
> 
>     1. Source B that depends on Source A with templating involved which is solved by this PR 🎉
> 
>     2. Source B that depends on Source A without templating.
> 
> 
> In the second scenario, I think before executing a source, we need to validate that sources listed in the dependency list succeeded. Just before this [line](https://github.com/updatecli/updatecli/blob/1efbf6006d72d227480de617fb675aec8ba7452a/pkg/core/pipeline/sources.go#L36)
> 
> Something like
> 
> ```
> for _, dependedSource := range s.dependsOn {
>     if p.Sources[dependencySource].Result == result.Failure {
>     // skipping source as depended source already failed
>    }
> }
> ```

Good catch, thanks! Implemented (same idea with the loop, sligthly different if cases) WDYT?

dduportal:
Let me rebase @olblak (in the next 5 min)

dduportal:
@olblak ready to review :)

pkg/core/pipeline/sources.go
```diff
@@ -33,7 +33,16 @@ func (p *Pipeline) RunSources() error {
 		logrus.Infof("\n%s\n", id)
 		logrus.Infof("%s\n", strings.Repeat("-", len(id)))
 
-		err = source.Run()
+		shouldRunSource := true
+		for _, parentSource := range source.Config.DependsOn {
+			if p.Sources[parentSource].Result != result.SUCCESS {
+				logrus.Warningf("Parent source[%q] did not succeed. Skipping execution of the source[%q]", parentSource, id)
+				shouldRunSource = false
```
olblak, line 40: What do you think to break early if we identify that a "source" failed.

```suggestion
				shouldRunSource = false
				break
```

olblak: I am wondering if we need to explicitly set the current source result to `result.Failure` if one of the dependency failed. 
Such as 
```suggestion
				shouldRunSource = false
				source.Result = result.FAILURE
```

olblak (approved):
Thanks for the pull request, I made minor suggestions that I don't consider blocker for merging.
Feel free to proceed

pkg/core/pipeline/sources.go
dduportal: > I am wondering if we need to explicitly set the current source result to result.Failure if one of the dependency failed.

That feels weird since the source was not run (e.g. skipped).
Do not forget that, as for todat, the default state of a source is "skipped", which is the reason why no state change is done here (otherwise I would have set it to "skipped" instead of whatever default, because it's not run).

dduportal: > What do you think to break early if we identify that a "source" failed.

Could be discussed, but not in the scope of this PR, because it would be a breaking change in the UX behavior. Do you mind raising an issue on that topicas enhancement?

dduportal:
@olblak thanks for the review. I've answered to your 2 elements: both are outside of the scope of the PR but are legit questions (and worth an issue).

I'm merging (as it's expected to be minor) since I got your approval, don't hesitate to open issues on the topic, or revert the PR if you see something we missed :)

Issue #289, opened 2021-09-29
Feature Request [Resource "YAML"] allow condition to only check for existence of a key (and not the value)


dduportal:
**Is your feature request related to a problem? Please describe.**

Let's consider the following YAML manifest: https://github.com/jenkins-infra/jenkins-infra/blob/958091069665a88a32a72e154e644d55f54e1b71/updatecli/weekly.d/buildmaster-tools.yaml#L70

I would like to add a condition that would check for the existence of the key in the YAML file.

The rationale is that as soon as #60 is implemented, then the execution would not fail when the key does not exist, leading to false positive (the key would be created and updatecli would report a success while the user would want to be notifieed of this as a failure because the YAML was changed recently but not updatecli manifest).

**Describe the solution you'd like**

Adding a new attribute `keyOnly` only available on the condition of kind "YAML", that would ignore the value and only check for the existence of the key.

**Describe alternatives you've considered**

- Using a shell condition with the command `yq`

**Additional context**
N.A.

olblak:
That would definitely be usefull, I am pretty sure we could easily detect

PR #398, opened 2021-11-26
fix(conditions) prints validation error when found

dduportal:
This PR is a fix for a bug I found while working on #289 .

It ensures that the validation error messages are printed to the end user when a **condition** failed to be validated.

## Test

With a manifest having the following condition (missing `spec.file`)

```yaml
# ...
conditions:
  checkUsername:
    kind: yaml
    spec:
      key: github.username
```

With updatecli 0.16.0, you got:

```shell
# ...
CONDITIONS:
===========

checkUsername
-------------
```

=> result is expected but no message to help the end user.

With this PR, you got:
```shell
$ go build -o ./dist/updatecli && ./dist/updatecli diff
# ...
CONDITIONS:
===========

checkUsername
-------------
ERROR: Validation error: the provided manifest configuration had the following validation errors:
Invalid spec for yaml resource: 'file' is empty.

```

=> which is clearly an improvement

### DIFF
--- a/pkg/core/engine/condition/main.go
+++ b/pkg/core/engine/condition/main.go
@@ -54,6 +54,7 @@ func (c *Condition) Run(source string) (err error) {
 	spec, err := Unmarshal(c)
 	if err != nil {
 		c.Result = result.FAILURE
+		logrus.Errorf("%s", err)
 		return err
 	}
 


dduportal:
@olblak I've added this to the 0.16.1 as it's clearly worth a patch: sounds good for you?

olblak (approved):

PR #400, opened 2021-11-26
fix: fail validation when SourceID is missing on conditions and targets

### LINKED ISSUE #281: No error return when sourceID doesn't exist
**Describe the bug**
We don't trigger an error if we reference a sourceID which doesn't exist such as in the following example.

```
sources:
  getLatestVersion:
    kind: shell
    spec:
      command: echo ami-XXXXX
conditions:
  checkIfAmiExist:
    name: Test if ami1 exist
    sourceID: doNotExist
    kind: shell
    spec:
      command: echo OK
```

**To Reproduce**
Just specific a sourceID which doesn't exist in any configuration

**Expected behavior**
I would expect an early error message

**Screenshots**
-

 - OS: linux
 - Version: V0.7.1

**Additional context**
An early validation could be done when we load the configuration [here](https://github.com/updatecli/updatecli/blob/90e37f8ec93a945097b767762957aad746c9e022/pkg/core/config/main.go#L124)

jplayout:
# fix: fail validation when SourceID is missing on conditions and targets

Fix #281

fix(condition) fail validation when SourceID is missing
fix(target) fail validation when SourceID is missing

## Test

test.yaml
```
sources:
  getLatestVersion:
    kind: shell
    spec:
      command: echo ami-XXXXX
conditions:
  checkIfAmiExist:
    name: Test if ami1 exist
    sourceID: doNotExist
    kind: shell
    spec:
      command: echo OK
targets:
  updateAmiValue:
    name: update the AMI value
    sourceID: doNotExist
    kind: shell
    spec:
      command: echo OK
```

```shell
go build -o ./dist/updatecli
./dist/updatecli diff --config test.yaml
```

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -144,7 +144,12 @@ func (config *Config) Validate() error {
 	}
 
 	for id, c := range config.Conditions {
-
+		if len(c.SourceID) > 0 {
+			if _, ok := config.Sources[c.SourceID]; !ok {
+				logrus.Errorf("the specified SourceID %q for condition[id] does not exist", c.SourceID)
+				return ErrBadConfig
+			}
+		}
 		// Only check/guess the sourceID if the user did not disable it (default is enabled)
 		if !c.DisableSourceInput {
 			// Try to guess SourceID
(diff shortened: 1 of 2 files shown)


olblak (approved):
🎉

PR #388, opened 2021-11-21
Move scm directive to top level

olblak:
# Move SCM directive to top-level

Fix #260 
Fix #261 

Before this pull request, "scm" configurations are configured per resource at the source/condition/target level.
The consequence of that approach is that it generates a lot of duplicates.
Starting from now, each scm configuration is defined once at the top level and each resource (source/condition/target) references the scm using a scm id.

Due to how "scm" is a central component of updatecli, I had to differentiate scm "configuration" provide via a configuration, scm interface used by git and Github to manipulate git repository.
I also split `pull request` configuration into a top-level one as explained in fix #261 . I was refactoring the `scm` code anyway

Here is an example of a new configuration
.example
```
scms:
    updatecliRepository:
        kind: github
        spec: 
            owner: updatecli
            repository: updatecli
           # Skipping some parameter for better example visibility
source:
    updatecliVersion:
        kind: yaml
        scmID: updatecliRepository
        spec: 
            file: example.yaml
            key: image
           # Skipping some parameter for better example visibility
```

## Test

To test this pull request, you can run the following commands:

```shell
go build -o bin/updatecli; ./bin/updatecli diff --config examples/updateCli.generic/githubRelease.tpl --values examples/values.yaml
./bin/updatecli show --config examples/updateCli.generic/githubRelease.tpl --values examples/values.yaml
```

## Additional Information

In order to not break the existing updatecli pipeline configuration, I did the extra effort to convert "internally" the old syntax to the new one with the specificity that a GitHub scm would also automatically add pull request configuration.
I am still not sure if I want that to be the definitive approach.

In this pull request, I also decided the move the go packages `source`, `condition`, `target`, under the package `pipeline` instead of `engine` as I think it makes more sense to be there, next to scm and pull request. 
    
### Changelog

An updatecli pipeline configuration can now use two additional top resources, `scms` and `pullrequests` where they respectively allow to configuration sc's configuration and pullrequests' configuration.
Similiar to `sources`, `conditions`, and `targets`: `scms` and `pullrequests` expect a "map" of configuration.

The resources of type "source", "condition", "target", "pullrequest" now uses the key `scmID` to specify an scm dependency.
This avoids duplicating the scm configuration in every resource like it used to be.

Here is an example

```
scms:
  defaultSCMConfig:
    name: Defines github repository
    kind: github
    spec:
      ...

pullrequests:
  defaultPullrequestConfig:
    name: Defines github pullrequest
    kind: github
    spec:
      ...

sources:
  default:
[...]

Issue #373, opened 2021-11-14
Remove field prefix and postfix for sources,conditions,target


olblak:
**Is your feature request related to a problem? Please describe.**
Time to remove obsolete code

**Describe the solution you'd like**
Field `prefix` and `postfix` have been deprecated in favor of transformers on the 2021/01/03.
It's need time to do some cleanup

**Describe alternatives you've considered**
-

**Additional context**
Cleanup example and documentation as well
* https://www.updatecli.io/docs/stages/target/#_parameters
* https://www.updatecli.io/docs/stages/source/#_parameters

olblak:
@dduportal the current issue is set for 0.19 but I would like to release 0.19 soon 
Any objections to move this issue to release 0.20?

dduportal:
No objections for moving it to 0.20.

dduportal:
- moved to 0.20.0 milestone
- I'm willing to take it (like #413 )
- It needs #412 to be merged

dduportal:
Closed by https://github.com/updatecli/updatecli/pull/591

PR #412, opened 2021-12-11
feat(manifests) resource kind is case-insensitive (internal refactor)

dduportal:
Depends on #388 , #414, #415, #418, #453, #456, #457, #479, #480, #481, #482 (will be rebased once all merged).

This PR is a refactor around the plugins to factorize common behaviors of resources (e.g. sources, conditions and targets) when possible, by using struct composition (https://golangbot.com/inheritance/) to look-a-like OO Object-Oriented's inheritance.

The expected outcomes are:

- (end-users) the type of resource, specified by the attribute `kind`, is now case un-sensitive
- (contributors) A single place to initialize the list of plugins (centralized unmarshaling for instance)
- (contributors) improved code convention for plugins (meaning constant behavior)
- (contributors) code base easier to browse as resources plugins are separated from scms plugins and "utils/generic" plugins

I recommend you to review this PR commit by commit (without checking the 4 commits from #482 of course) since it is refactoring!
It introduces the following changes (1 per commit):

- Introduces a new package `resource` with a ResourceConfig struct. This package defines the common behavior that any plugin resource should implement.
- Factorize the unmarshalling away from source, condition and target, into this new resource package. It allows a centralized place to list the resource plugin + delegating parsing logic on each resource plugin.
- This "centralization" allows to define the `kind` of each resource to be case insensitive 🥳 
- The changelog retrieval is now implemented by all the resources. If not implemented, then it returns an empty string. It makes the changelog retrieval way easier to implement/change and makes it usable with the internal state of resources.
- Refactor the directory + package structure of plugin. It makes it really clear which package implements which kind of "behavior" in updatecli

```shell
$ tree -L 1 pkg/plugins
pkg/plugins
├── resources # plugins that implements the interface resource (e.g. a source/condition/target)
├── scms # plugins that implements the SCM interface (git / github)
└── utils # other plugins which behaviors are reused across other plugins but not in core
```

## Test

```shell
make test
make test-short
make test-e2e
```

## Additional Information

Please note that this PR lays the foundation for #373, #413 and #497 (deprecating unused fileds)


pkg/core/pipeline/resource/main.go
```diff
@@ -0,0 +1,78 @@
...
+	// Deprecated in favor of Transformers on 2021/01/3
+	Postfix      string
+	Transformers transformer.Transformers
+	Spec         interface{}
+	// Deprecated field on version [1.17.0]
+	Scm   map[string]interface{}
+	SCMID string `yaml:"scmID"` // SCMID references a uniq scm configuration
+}
+
+// Unmarshal decode a source spec and returned its typed content
```
lemeurherve, line 38:
```suggestion
// Unmarshal decode a source spec and return its typed content
```

pkg/core/pipeline/source/main.go
```diff
@@ -1,63 +1,36 @@
...
-	Spec         interface{}
-	Scm          map[string]interface{} // Deprecated field on version [x.y.z]
-	SCMID        string                 `yaml:"scmID"` // SCMID references a uniq scm configuration
-}
-
-// Sourcer source is an interface to handle source spec
-type Sourcer interface {
-	Source(workingDir string) (string, error)
+	resource.ResourceConfig `yaml:",inline"`
+	// Deprecated in favor of Transformers on 2021/01/3
```
lemeurherve, line 26: What does it mean?

pkg/plugins/resources/dockerdigest/main.go
```diff
@@ -28,7 +29,13 @@ type DockerDigest struct {
...
+func New(spec interface{}) (*DockerDigest, error) {
+	newSpec := Spec{}
+
+	err := mapstructure.Decode(spec, &newSpec)
+	if err != nil {
+		return nil, err
+	}
 
 	newImage, err := dockerimage.New(
 		// concatenate user-provided image name (that may contain registry and namespace along with the namespace) with the user-provided tag
```
lemeurherve, line 41:
```suggestion
		// concatenate user-provided image name (which may contain registry and namespace along with the namespace) with the user-provided tag
```
(Out of scope, I know..)

pkg/plugins/resources/helm/changelog.go
```diff
@@ -8,43 +8,37 @@ import (
 	"net/http"
 
 	"github.com/sirupsen/logrus"
-	"github.com/updatecli/updatecli/pkg/plugins/version"
 )
 
-// Changelog return any information available for a helm chart
-func (c *Chart) Changelog(release version.Version) (string, error) {
+// Changelog returns a rendered template with the informations of this version of the chart
```
lemeurherve, line 13:
```suggestion
// Changelog returns a rendered template with this chart version informations
```

pkg/core/pipeline/resource/main.go
dduportal:
```suggestion
// Unmarshal decodes a source spec and returns its typed content
```

better with 3rd person

pkg/core/pipeline/source/main.go
dduportal: It's the comment associated to the line below (e.g. the attrobite Replaces Repalcers).

If find really hard comment at the end of a line to read, but I realize that most of the code of updatecli does it so I'll revert (I did it by habit while experimenting and it made it to this PR...)

pkg/core/pipeline/source/main.go
```diff
@@ -1,63 +1,36 @@
...
+	Replaces Replacers
```
dduportal, line 27:
```suggestion
	Replaces Replacers // Deprecated in favor of Transformers on 2021/01/3
```

pkg/core/pipeline/source/main.go
dduportal: (applied a modification to make it a single line).

lemeurherve: I don't understand the meaning and goal of this comment.
Why something one year old should be mentioned?

dduportal: I don't either, but the goal of this PR is not to cleanup so I don't see the point of changing it here.

However it's legit: https://github.com/updatecli/updatecli/issues/497 just opened an issue to track it :)

lemeurherve: Oh I see, I initially thought you've created and added this comment, and thus I was wondering why, or if that was a typo.

dduportal: No problem, I understood that afterwards, hence the issue because it's totally legit :) Thanks!

pkg/core/pipeline/resource/main.go
```diff
@@ -0,0 +1,34 @@
...
+	// Deprecated in favor of Transformers on 2021/01/3
+	Postfix      string
+	Transformers transformer.Transformers
+	Spec         interface{}
+	// Deprecated field on version [1.17.0]
+	Scm   map[string]interface{}
+	SCMID string `yaml:"scmID"` // SCMID references a uniq scm configuration
+}
+
+// type ResourceSpec interface {
```
olblak: Why is it commented?

pkg/core/pipeline/source/main.go
olblak: I guess it's a taste :p

olblak:
@dduportal Well done, it's a nice pull request. It's interesting to see that you implemented in ways hesitated to do. I hesitated a Like the same interface for source, condition, target, but in the end, your approach makes things clearer.
Considering that it's a very big refactoring, I would like to take additional  time for testing.

olblak:
I think before merging this PR we should also clean up the [CONTRIBUTING](https://github.com/updatecli/updatecli/blob/main/CONTRIBUTING.adoc#plugins)

olblak:
So far I didn't identify any regression 🎉

olblak (approved):

lemeurherve (approved):

PR #415, opened 2021-12-14
Chore: bump golangcilint to 1.43 and apply updated code suggestions

dduportal:
This PR comes from the work in #412 .

It updates the golantci-lint configuration file to support the latest 1.43.x line.

It also applies some of its suggestions.

## Test

```shell
go build -o dist/updatecli
./dist/updatecli diff --config ./examples/updatecli.generic/file.yaml
./dist/updatecli diff --config ./examples/updatecli.generic/yaml.yaml
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,16 +33,16 @@ linters-settings:
       - wrapperFunc
   gocyclo:
     min-complexity: 15
-  goimports:
   golint:
     min-confidence: 0
   gomnd:
     settings:
       mnd:
         # don't include the "operation" and "assign"
-        checks: argument,case,condition,return
+        checks: [argument,case,condition,return]
   govet:
     check-shadowing: true
+    fieldalignment:
     settings:
       printf:
         funcs:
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -121,6 +122,6 @@ run:
 # golangci.com configuration
 # https://github.com/golangci/golangci/wiki/Configuration
 service:
-  golangci-lint-version: 1.27.x   # Use the fixed version to not introduce new linters unexpectedly
+  golangci-lint-version: 1.43.x   # Use the fixed version to not introduce new linters unexpectedly
```
olblak, line 125: Can I ask you to also add an updatecli configuration to bump this version.
It should be pretty straightforward https://github.com/golangci/golangci-lint/releases/tag/v1.43.0

olblak (dismissed):
Thanks for looking into that, I have been pushing back this for too long. 
I made a minor suggestion that could help us to keep track of new golangci new versions.
The current PR is good as is

.golangci.yml
dduportal: Don't merge: I'll add a commit with this change

dduportal: PR upgraded: in fact we won't need an automated version.
I've deleted the `service:` section in `.golangci.yml` as per https://github.com/golangci/golangci/wiki/Configuration#service-configuration : we don't need it.
Thus we don't need to specify the golangcitlint version as well.

The only other location where we specify a fixed version for golangcilint is the githubaction but it's tracked by dependandabot.

So it sounds like we should be ready to roll.

PR #422, opened 2021-12-20
fix: correctly escape changelog

lemeurherve:
To avoid putting the changelog as unescaped/preformated text like in this PR: https://github.com/jenkins-infra/docker-builder/pull/26

Fixes https://github.com/jenkins-infra/docker-builder/issues/27

### DIFF
--- a/pkg/plugins/github/pullrequest.go
+++ b/pkg/plugins/github/pullrequest.go
@@ -38,7 +38,7 @@ const PULLREQUESTBODY = `
 
 <details><summary>Click to expand</summary>
 
-` + "```\n{{ .Description }}\n```" + `
+` + "````\n{{ .Description }}\n````" + `
 
 </details>
 


dduportal:
Just tried this change: https://github.com/dduportal/docker-builder/pull/1 .

At first glance, it break the rendering (I might have given you a bad syntax): I'm currently checking more indetails

dduportal (changes requested):
Currently trying to debug this change

dduportal:
Ping @olblak : I'm trying to confirm that, but it seems that the "empty" changelog is a regression introduced by 0.17.0.

Example: https://github.com/jenkins-infra/kubernetes-management/pull/1831

dduportal (approved):
Looks good to me!

The "empty block" is another issue (see #423).

This changes behaves as expected.

PR #436, opened 2021-12-31
Don't try to open a pullrequest when not needed

### LINKED ISSUE #431: Updatecli tries to open a pullrequest even if nothing changed
**Describe the bug**
Updatecli tries to open a pull request even if nothing changed.
This issue was first identified on [asciidoctor/docker-asciidoctor](https://github.com/asciidoctor/docker-asciidoctor/runs/4613040569?check_suite_focus=true#step:6:183
)
This problem is related to https://github.com/updatecli/updatecli/issues/424

**Steps to Reproduce**
Run a updatecli pipeline configured to open a Github Pull Request which contains no change so the working branch doesn't exist on the remote location.

**Expected behavior**
We shouldn't try to open a pull request when the branch doesn't exist in the remote location
We should try to open a pull request if a change is detected between the remote branch and the remote target branch

**Screenshots**
[](https://github.com/asciidoctor/docker-asciidoctor/runs/4613040569?check_suite_focus=true#step:6:183)

**Version**

- OS: linux
- Updatecli version: v0.17.1

**Additional context**

olblak:

# Don't try to open a pullrequest when not needed

Fix #431 

Add a function to test if the latest commits of two branches are similar. The intend of this approach is to only try to create a pullrequest when two branch content mismatch
 
## Test

```shell
cp pkg/plugins/git/generic
go test
```

### Potential improvement

I am planning to also look at a way to test if two git tags are similar

### DIFF
--- a/pkg/plugins/git/generic/main_test.go
+++ b/pkg/plugins/git/generic/main_test.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestIsSimilarBranch(t *testing.T) {
@@ -38,7 +40,7 @@ func TestIsSimilarBranch(t *testing.T) {
 			branchA:        "main",
 			branchB:        "doNotExist",
 			workingDir:     "../../../..",
-			expectedResult: true,
+			expectedResult: false,
 			expectedError:  fmt.Errorf("reference not found"),
 		},
 	}
@@ -49,11 +51,13 @@ func TestIsSimilarBranch(t *testing.T) {
 			d.branchB,
 			d.workingDir)
 
-		if err != nil {
-			fmt.Println(err)
+		if !assert.Equal(t, err, d.expectedError) {
+			t.Errorf("Expected error '%v' but got '%v'", d.expectedError, err)
+
 		}
-		if got != d.expectedResult {
-			t.Errorf("Expected %v but got %v", d.expectedResult, got)
+
+		if !assert.Equal(t, got, d.expectedResult) {
+			t.Errorf("Expected result '%v' but got '%v'", d.expectedResult, got)
 		}
 	}
 }
--- a/pkg/plugins/git/generic/main_test.go
+++ b/pkg/plugins/git/generic/main_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 )
 
-func TestDiffBranch(t *testing.T) {
+func TestIsSimilarBranch(t *testing.T) {
 
 	type data struct {
 		branchA        string
@@ -44,7 +44,7 @@ func TestDiffBranch(t *testing.T) {
 	}
 
 	for _, d := range dSet {
-		got, err := DiffBranch(
+		got, err := IsSimilarBranch(
 			d.branchA,
 			d.branchB,
 			d.workingDir)
--- a/pkg/plugins/git/generic/main_test.go
+++ b/pkg/plugins/git/generic/main_test.go
@@ -1,66 +1,67 @@
 package generic
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
-func TestIsSimilarBranch(t *testing.T) {
-
-	type data struct {
-		branchA        string
-		branchB        string
-		workingDir     string
-		expectedResult bool
-		expectedError  error
-	}
-
-	type dataSet []data
-
-	dSet := dataSet{
-		{
-			branchA:        "main",
-			branchB:        "issue-285",
-			workingDir:     "../../../..",
-			expectedResult: false,
-			expectedError:  nil,
-		},
-		{
-			branchA:        "main",
-			branchB:        "main",
-			workingDir:     "../../../..",
-			expectedResult: true,
-			expectedError:  nil,
-		},
-		{
-			branchA:        "main",
-			branchB:        "doNotExist",
-			workingDir:     "../../../..",
-			expectedResult: false,
-			expectedError:  fmt.Errorf("reference not found"),
-		},
-	}
-
-	for id, d := range dSet {
-		t.Run(fmt.Sprint(id), func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


olblak:
Well for some reason the github action doesn't find the branch "main"

olblak:
I successfully tested this PR with https://github.com/jenkins-infra/kubernetes-management/pull/1879
But I can't figure out why the test is not working from Github Action

dduportal:
The error also appears on my machine (macOS 12.1, MBP 16" Intel, go1.17.5):

```shell
$ gh pr checkout 436
$ make test-short    
go test ./... -short
# ...
time="2022-01-02T09:27:53+01:00" level=error msg="reference \"main\" - reference not found"
time="2022-01-02T09:27:53+01:00" level=error msg="reference \"main\" - reference not found"
time="2022-01-02T09:27:53+01:00" level=error msg="reference \"main\" - reference not found"
--- FAIL: TestIsSimilarBranch (0.00s)
    main_test.go:54: 
                Error Trace:    main_test.go:54
                Error:          Not equal: 
                                expected: *errors.errorString(&errors.errorString{s:"reference not found"})
                                actual  : <nil>(<nil>)
                Test:           TestIsSimilarBranch
    main_test.go:55: [0] Expected error '<nil>' but got 'reference not found'
    main_test.go:54: 
                Error Trace:    main_test.go:54
                Error:          Not equal: 
                                expected: *errors.errorString(&errors.errorString{s:"reference not found"})
                                actual  : <nil>(<nil>)
                Test:           TestIsSimilarBranch
    main_test.go:55: [1] Expected error '<nil>' but got 'reference not found'
    main_test.go:61: 
                Error Trace:    main_test.go:61
                Error:          Not equal: 
                                expected: false
                                actual  : true
                Test:           TestIsSimilarBranch
    main_test.go:62: [1] Expected result 'true' but got 'false'
FAIL
FAIL    github.com/updatecli/updatecli/pkg/plugins/git/generic  7.026s
#...
```

=> I'm trying to debug it since it appears locally.

pkg/plugins/git/generic/main_test.go
```diff
@@ -1,11 +1,72 @@
...
+		}
+
+		if !assert.Equal(t, got, d.expectedResult) {
+			t.Errorf("[%v] Expected result '%v' but got '%v'",
+				id,
+				d.expectedResult,
+				got)
+		}
+	}
+}
```
dduportal, line 68:
```suggestion
	for id, d := range dSet {
		t.Run(fmt.Sprint(id), func(t *testing.T) {
			got, err := IsSimilarBranch(
				d.branchA,
				d.branchB,
				d.workingDir)

			if d.expectedError != nil {
				assert.Equal(t, err, d.expectedError)
				return
			}

			require.NoError(t, err)
			assert.Equal(t, got, d.expectedResult)
		})
	}
```

- Use a ["sub-test"](https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks) with the instruction `t.Run` to identify the dataSet
- When using the `assert` methods, there is no need to write the error message by yourself: it's the goal of using assertion (it takes care of the message .. and more... for you).
- There is also the `require + return` pattern for the error checking:
  - Thanks to the `t.Run(){}` block, you can use the `return` instruction to exit the block (and iterate to the next subtest / test). It allows you not having to write conditional when the error is expected by the test.
  - The `require.` is a library from testify (like assert) that behaves like `assert` but fails the sub-test/test immediatly if it is false

pkg/plugins/git/generic/main_test.go
```diff
@@ -1,11 +1,72 @@
...
+		workingDir     string
+		expectedResult bool
+		expectedError  error
+	}
+
+	type dataSet []data
+
+	dSet := dataSet{
+		{
+			branchA:        "main",
```
dduportal, line 26: This test assumes that the project directory hosting the code has a local git reference (short) named `main`.

- It isn't verified in CI systems (GH Actions, but alos Gitlab and Jenkins and many others): only the current HEAD's reference (e.g. the PR source branch) is checkouted (to avoid heavy checkouts all the time). In the case of this PR it's even worse: the local HEAD's reference is not deterministic because it's a merge of the HEAD reference with the destination's branch.
- It isn't verified on contributor machines: as the contributing rules says, a topic/feature branch is needed to open a PR: there is no reason to have checkouted the `main` before running the test.


This test is not valid because it relies on a state that you do not control.

WDYT about creating (and initializing) an empty git workdir for each dataSet, so you control the git references (this is an integration test because it interacts with a real git project)?

The alternative is to make it a unit test, but it needs more heavy lifting (as you would have to mock the git project).

dduportal (changes requested):

dduportal:
The fix is working for me on both the asciidoctor/docker-asciidoctor and jenkins-infra/* repositories on my manual tests ✅ .

Many thanks for this huge work to start 2022 🍾

pkg/plugins/git/generic/main_test.go
olblak: Thanks @dduportal for the suggestion and the explanation

pkg/plugins/git/generic/main_test.go
olblak: > It isn't verified in CI systems (GH Actions, but alos Gitlab and Jenkins and many others): only the current HEAD's reference (e.g. the PR source branch) is checkouted (to avoid heavy checkouts all the time). In the case of this PR it's even worse: the local HEAD's reference is not deterministic because it's a merge of the HEAD reference with the destination's branch.

Thanks, that was the purpose of this [commit](https://github.com/updatecli/updatecli/pull/436/commits/bf6067c4d33fbf7a9f8789db82e1a54cf1cdfadb), I thought that I could rely on the branches but it doesn'
t seem so, I'll comment my test.

> WDYT about creating (and initializing) an empty git workdir for each dataSet, so you control the git references (this is an integration test because it interacts with a real git project)?
The alternative is to make it a unit test, but it needs more heavy lifting (as you would have to mock the git project).

I prefer the UT approach but I won't have the time to work on this now and I think the fix is more urgent. so I'll comment my tests for now

dduportal (approved):
🚀

PR #437, opened 2022-01-02
feat(yaml, helm) allow condition to check only for a key (and not its value) + refactoring

### LINKED ISSUE #289: Feature Request [Resource "YAML"] allow condition to only check for existence of a key (and not the value)

dduportal:
Fix #289

This PR introduces the following changes to the YAML and `helm` resources:

* [yaml] allow condition to check only for a key (and not its value) (#289)
* [yaml] fix a validation error when file path started with the prefix `file://`
* [yaml] refactor to use the same naming conventions as others like `aws/ami` or `shell`, and
* [helm] rename package to follow golang convention (no need for subpackage if there is only one)
* [helm] introduce a `New()` method to have a constructor.
* [yaml] ensure that private attributes are used (decoupled from Helm resource)
* Add examples

## Test

```shell
go build -o ./dist/updatecli
```

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Diff
-        uses: updatecli/updatecli-action@v1.17.0
+        uses: updatecli/updatecli-action@v1.19.0
         with:
           command: diff
           flags: "--config ./updatecli/updatecli.d"
@@ -27,7 +27,7 @@ jobs:
           UPDATECLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Apply
-        uses: updatecli/updatecli-action@v1.17.0
+        uses: updatecli/updatecli-action@v1.19.0
         if: github.ref == 'refs/heads/main'
         with:
           command: apply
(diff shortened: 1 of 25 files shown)


dduportal:
@olblak I propose to add this to the `0.18.0` milestone, WDYT?

olblak:
> @olblak I propose to add this to the `0.18.0` milestone, WDYT?

It sounds good to me but as a reminder, we decided to assign issues to milestone and not pull request

dduportal:
> > @olblak I propose to add this to the 0.18.0 milestone, WDYT?
> It sounds good to me but as a reminder, we decided to assign issues to milestone and not pull request

Sorry, I thought I was commenting in the issue. Good thing that you point it out to me :)

olblak (approved):
Thanks for the refactoring as well

Issue #390, opened 2021-11-24
target ENV for Dockerfile is adding the new value instead of replacing the old one


lemeurherve:
**Describe the bug**
Using an ENV keyword on a dockerfile type target, I've noticed the new value is added before the previous value instead of replacing it:
> 🐋 On (Docker)file "/var/folders/z6/xykmq_m902n_7v4j8sxhzkdm0000gn/T/updatecli/jenkins-infra/docker-packer/Dockerfile":
> ✔ The line #3, matched by the keyword "ENV" and the matcher "WINDOWS_UPDATE_VERSION", is changed from "ENV WINDOWS_UPDATE_VERSION 0.10.1" to "ENV WINDOWS_UPDATE_VERSION=v0.14.0 0.10.1".

**To Reproduce**
On the https://github.com/jenkins-infra/jenkins-packer repository, try this updatecli manifest:
```
title: Bump Packer Windows Update Provisioner plugin version
sources:
  latestRelease:
    kind: githubRelease
    name: Get latest rgl/packer-plugin-windows-update version
    spec:
      owner: "rgl"
      repository: "packer-plugin-windows-update"
      token: "{{ requiredEnv .github.token }}"
      username: "{{ .github.username }}"

# No condition here as we're downloading the plugin from github directly (no docker image involved)
targets:
  updateReleaseInDockerfile:
    name: "Update the version of the Packer Windows Update Provisioner plugin in the Dockerfile"
    kind: dockerfile
    spec:
      file: Dockerfile
      instruction:
        keyword: "ENV"
        matcher: "WINDOWS_UPDATE_VERSION"
    scm:
      github:
        user: "{{ .github.user }}"
        email: "{{ .github.email }}"
        owner: "{{ .github.owner }}"
        repository: "{{ .github.repository }}"
        token: "{{ requiredEnv .github.token }}"
        username: "{{ .github.username }}"
        branch: "{{ .github.branch }}"
```

For reference, here is the Dockerfile:
```
FROM hashicorp/packer:1.6.5

ENV WINDOWS_UPDATE_VERSION 0.10.1

ADD https://github.com/rgl/packer-provisioner-windows-update/releases/download/v${WINDOWS_UPDATE_VERSION}/packer-provisioner-windows-update_${WINDOWS_UPDATE_VERSION}_checksums.txt ./
[...]

PR #453, opened 2022-01-09
chore: refactor resource dockerfile

dduportal:
While working on #390 and #412, the need to refactor the `dockerfile` resource was raised.

This PR does not change the functionnality of the resource: it's why the label says "chore", except for the resulting commit messages (that were not really correct).

- Fix the example to allow end to end testing from the root of the repository
- Apply the same conventions as other refactored resources:
  - Define a `Spec` type on the package with only the public attributes
  - Add a `New` initializer method
- Improve Developer experience:
  - Full unit test with mocked injected dependencies (no more file to be read/written on the FS while running tests!)
  - Increased test coverage that helped cleaning up unused struct attributes

## Test

```shell
go build -o ./dist/updatecli && ./dist/updatecli diff --config examples/updatecli.generic/dockerfile
```


pkg/core/pipeline/condition/main.go
```diff
@@ -154,14 +154,16 @@ func Unmarshal(condition *Condition) (conditioner Conditioner, err error) {
 		}
 
 	case "dockerfile":
-		d := dockerfile.Dockerfile{}
+		var sourceSpec dockerfile.Spec
```
olblak, line 157: Shouldn't it be named "conditionSpec"?

pkg/core/pipeline/target/main.go
```diff
@@ -85,15 +85,17 @@ func Unmarshal(target *Target) (targeter Targeter, err error) {
 		targeter = &ch
 
 	case "dockerfile":
-		d := dockerfile.Dockerfile{}
+		sourceSpec := dockerfile.Spec{}
```
olblak, line 88: Shouldn't it be targetSpec?

pkg/core/pipeline/condition/main.go
dduportal: Good catch! PR update (and rebased)

pkg/core/pipeline/target/main.go
dduportal: Good catch! PR update (and rebased)

olblak (approved):
It looks good to me
🚢

PR #470, opened 2022-01-16
hotfix(conditions) show error message when a condition errors

dduportal:
Not usre when this "misbehavior" was introduced but currently, conditions that fails because of an error does not show the error message.

It's not really convenient, particularly when it is a validation error 😅 

This PR fixes the print of error message (minor cleanups around happy path / string comparisons)

## Test

```shell
make build
make test
make test-e2e
```

### DIFF
--- a/pkg/core/pipeline/conditions.go
+++ b/pkg/core/pipeline/conditions.go
@@ -4,7 +4,6 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
-	"github.com/updatecli/updatecli/pkg/core/result"
 )
 
 // RunConditions run every conditions for a given configuration config.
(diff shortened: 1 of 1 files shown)


pkg/core/pipeline/conditions.go
```diff
@@ -42,22 +42,24 @@ func (p *Pipeline) RunConditions() (globalResult bool, err error) {
 				p.Sources[condition.Config.SourceID].Output +
 				p.Sources[condition.Config.SourceID].Config.Postfix)
 
+		if err != nil {
+			// Show error to end user
+			logrus.Error(err)
+			if condition.Result != result.SUCCESS {
```
lemeurherve, line 48: Does this check has to be done here instead of l51 outside the "if err"?

pkg/core/pipeline/conditions.go
```diff
@@ -42,22 +42,24 @@ func (p *Pipeline) RunConditions() (globalResult bool, err error) {
...
+				globalResult = false
+			}
+		}
 		rpt.Result = condition.Result
 
+		// Update pipeline after each condition run
 		if err == nil {
```
lemeurherve, line 55: Does this check is useful as the opposite is checked l45 now?

dduportal: yes it is: my understanding of the code here (yes, no tests nor comment) to run ALL conditions despite their return status (true/fase + error or not).

I've updated the code with an "else" directive as they are no "early return". It should help for readibility WDYT?

pkg/core/pipeline/conditions.go
dduportal: see updated code

pkg/core/pipeline/conditions.go
lemeurherve: I like it jetter indeed :)

lemeurherve (approved):

olblak (approved):

PR #479, opened 2022-01-22
chore(githubrelease) factorize githubrelease plugin out of the github package

dduportal:
This PR is a refactor of the github plugin to allow continuing the work in #412.
No functionnal changes (hence the chore) label.

- Move all the "githubrelease" logic into a separated plugin packaged as `githubrelease` with its own unit tests
- Introduces news interfaces in the `github` plugin to define the expect contract of this plugin, and allowing to mock it out on other elements
- Refactor the releases, tags, label and changelog GH v4 API logics into their own custom query types to allow mocking
- Moved github tests to unit tests (with mocking the GH API)
- Cleanup methods and attributes on both plugins (using private element as much as possible)

## Test

```shell
go build -o ./dist/updatecli && ./dist/updatecli diff --config examples/updatecli.generic/githubRelease.yaml
```


pkg/core/pipeline/source/main.go
```diff
@@ -179,22 +184,18 @@ func (s *Source) Unmarshal() (sourcer Sourcer, changelog Changelog, err error) {
...
-		err := mapstructure.Decode(s.Config.Spec, &githubSpec)
-
-		if err != nil {
+		if err := mapstructure.Decode(s.Config.Spec, &sourceSpec); err != nil {
 			return nil, nil, err
 		}
 
-		g, err := github.New(githubSpec)
-
+		grPrResource, err := githubrelease.New(sourceSpec)
```
olblak, line 193: What's the meaning of the variable grPrResource?

pkg/plugins/github/changelog.go
```diff
@@ -17,70 +18,66 @@ type Changelog struct {
...
+}
+type queriedRelease struct {
+	Description string
+	Url         string
+	PublishedAt time.Time
+}
 
-	client := g.NewClient()
+// Changelog returns a changelog description based on a release name
+func (g *Github) Changelog(version version.Version) (string, error) {
```
olblak, line 52: Without being a blocker, for now, I think it wouldn't have been better to create a new struct directly in the changelog package such as "changelogInformation or "changelogVersion"?
We now have an unnecessary strong binding between this package and version.

pkg/plugins/github/label.go
```diff
@@ -7,76 +7,77 @@ import (
...
-					Node   struct {
-						ID          string
-						Name        string
-						Description string
-					}
-				}
-			} `graphql:"labels(last: 5, before: $before)"`
-		} `graphql:"repository(owner: $owner, name: $repository)"`
-	}
+	var query labelsQuery
```
olblak, line 77: I really like this refactoring :)

olblak (dismissed):
What a nice refactoring.
Excepted minor questions, this pull request looks really great

olblak:
@dduportal  I propose to trigger a release just with the current change

pkg/core/pipeline/source/main.go
dduportal: Good catch, it is a typo. Should be the following:

```suggestion
		ghrResource, err := githubrelease.New(sourceSpec)
```

WDYT?

pkg/plugins/github/changelog.go
dduportal: Not sure to understand?

- Using a private struct for a public method is not possible (how would the externe caller be able to use the struct)?
- The intent is to change the signature of the method `Changelog` globally (see the changes on the `Changelog` interface below) and use the version.Version as argument type. Rationale is that it depends on the implementation of the changelog if either the parsed or originl version are required.

dduportal:
> @dduportal I propose to trigger a release just with the current change

No strong advise so go for it if you feel it

pkg/core/pipeline/source/main.go
dduportal: I went for `newGhrResource`. WDYT?

pkg/plugins/github/changelog.go
olblak: I was thinking to create a public struct defined in the changelog package so we wouldn't need to import the package version. But the way you implemented it is very good

olblak (approved):

PR #481, opened 2022-01-23
chore(maven) factorize metadata into a `mavenmetadata` package + apply code conventions for plugin

dduportal:
This PR is a refactor of the maven plugin to allow continuing the work in #412.
No functionnal changes (hence the chore) label.

- Update the maven example + add it to the end 2 end testing framework
- Moved the mavenmetadata logic into its own package + use it on both `maven` and `jenkins` to decouple
  - Include unit test with HTTP mocking (reusing the work from #453)
- Apply code conventions for the plugin `maven` (exports a `Spec` object and a `New(spec)` constructor method
  

## Test

```shell
go build -o ./dist/updatecli && ./dist/updatecli diff --config examples/updatecli.generic/maven.yaml
make test-e2e
```

## Additional Information

Jenkins plugin is the next to be refactored. Almost there \o/


examples/updatecli.generic/maven.yaml
```diff
@@ -1,81 +1,30 @@
 ---
```
olblak, line 1: @dduportal What do you think to deprecate the directory examples in favor of e2e which is tested with the latest updatecli version

pkg/plugins/jenkins/main.go
```diff
@@ -54,51 +51,19 @@ func (j *Jenkins) Validate() (err error) {
...
-	}
-
-	URL := fmt.Sprintf("https://%s/%s/%s/%s/maven-metadata.xml",
-		m.URL,
-		m.Repository,
-		strings.ReplaceAll(m.GroupID, ".", "/"),
-		m.ArtifactID)
-
-	req, err := http.NewRequest("GET", URL, nil)
+	m := mavenmetadata.New("https://repo.jenkins-ci.org/releases/org/jenkins-ci/main/jenkins-war/maven-metadata.xml")
```
olblak, line 54: I think it would be better to move this information to a package constant.

pkg/plugins/maven/condition.go
```diff
@@ -15,50 +11,25 @@ import (
...
-		return false, err
-	}
-
-	for _, version := range data.Versioning.Versions.Version {
-		if version == m.Version {
-			logrus.Infof("%s Version %s is available on Maven Repository", result.SUCCESS, m.Version)
+	for _, version := range versions {
+		if version == m.spec.Version {
+			logrus.Infof("%s Version %s is available on the Maven Repository (%s)",
+				result.SUCCESS, m.spec.Version, m.metadataHandler.GetMetadataURL())
```
olblak, line 26: What would you think to move the m.metadataHandler.GetMetadataURL to debug instead of info?

pkg/plugins/mavenmetadata/main.go
```diff
@@ -0,0 +1,90 @@
...
+
+// New returns a newly initialized DefaultHandler object
+func New(metadataURL string) *DefaultHandler {
+	return &DefaultHandler{
+		metadataURL: metadataURL,
+		webClient:   http.DefaultClient,
+	}
+}
+
+// getMetadatas is an internal method that returns the parsed metadata object
```
olblak, line 28:
```suggestion
// getMetadata is an internal method that returns the parsed metadata object
```
The "s" seems weird to me because we only get "one" metadata file content

examples/updatecli.generic/maven.yaml
dduportal: Look good to me. Incoming PR for this (so I'll add all the others yaml from example to e2e).

pkg/plugins/jenkins/main.go
dduportal: Good idea. Pointed in https://github.com/updatecli/updatecli/pull/482 (as it is centered on jenkins plugin)

pkg/plugins/mavenmetadata/main.go
dduportal: - You're correct, it is "1 metatdata file"
- But we retrieves the (multiple) metadataS from the file (e.g. the list of versions)
=> hence the plural form for the private method.

(Good catch for the comment thoug, I'm fixing it now)

dduportal: Given what this method does, WDYT renaming it to `getMetadataFile()` ? that would make it clearer and closer to the singular form you are proposing?

pkg/plugins/maven/condition.go
dduportal: Output looks like this with this set to debug:

```
SOURCES
=======

getLatestJenkinsWarVersion
--------------------------

getLatestJunitVersion
---------------------


CONDITIONS:
===========

checkIfOlderJunitExist
----------------------

=============================
```

=> Does not look like very well. Are you ok to keep it as info?

dduportal:
PR ready with your last feedbacks (and blocking #482).

pkg/core/httpclient/mocks.go
```diff
@@ -0,0 +1,13 @@
+package httpclient
+
+import "net/http"
+
+// MockClient is a stub implementation of an http.Client with a custom Do method
+type MockClient struct {
+	DoFunc func(req *http.Request) (*http.Response, error)
+}
+
+// Do is the mock client's `Do` func that you can customize
```
lemeurherve, line 10:
```suggestion
// Do is the mock client's `Do` function that you can customize
```

pkg/core/httpclient/mocks.go
```diff
@@ -0,0 +1,13 @@
+package httpclient
+
+import "net/http"
+
+// MockClient is a stub implementation of an http.Client with a custom Do method
```
lemeurherve, line 5:
```suggestion
// MockClient is a stub implementation of an http.Client with a custom `Do` method
```

pkg/core/pipeline/source/main.go
```diff
@@ -236,14 +236,17 @@ func (s *Source) Unmarshal() (sourcer Sourcer, changelog Changelog, err error) {
...
+		if err := mapstructure.Decode(s.Config.Spec, &sourceSpec); err != nil {
 			return nil, nil, err
 		}
 
-		sourcer = &m
+		mavenResource, err := maven.New(sourceSpec)
+		if err != nil {
+			return nil, nil, err
+		}
+		sourcer = mavenResource
```
lemeurherve, line 249: Is `mavenResource` used elsewhere?
If not,
```suggestion
		sourcer, err := maven.New(sourceSpec)
		if err != nil {
			return nil, nil, err
		}
```

lemeurherve (dismissed):

pkg/core/pipeline/source/main.go
dduportal: for info @lemeurherve 

```
sourcer, err = maven.New(sourceSpec)
```

=> without the `:=` !

Rationale is that, if you use the `:=` operator, then it means that you re-declare the variable sourcer in this scope, which hides the initial declaration.
The value assignation will be lost (and both linter and compiler complains that the variable sourcer is not used anywhere, thanks to them :) )

dduportal:
Thanks for the review @lemeurherve !

PR rebased with your proposal

lemeurherve (approved):

dduportal:
Since this PR is labelled as `chore` (e.g. no functionnal change), and that it has been reviewed with only minor elements, I'll put it as auto-merge with squash.

Thanks folks!

PR #490, opened 2022-01-26
[wip] feat(file): implement `spec.files` for `kind: file` (first part: `targets` block)

### LINKED ISSUE #435: Feature Request: ability to specify more than one target file
**Is your feature request related to a problem?**
When the same update has to be done on multiple files, multiple targets have to be declared, with the same content except the targeted file

Ability for the target file to take a map as input, or add a "files" keyword

N/A

**Additional context**

Exemple of target repetitions: same helm chart used in different helmfiles in https://github.com/jenkins-infra/kubernetes-management

lemeurherve:
# Title

Fix #435

## Test

```shell
cp <to_package_directory>
go test
```

## Additional Information

For now, with the new example "files.yaml" with "spec.file" it works.
Unfortunately, it doesn't when setting the same file in "spec.files". Current blocking issue, pushing this PR as draft for now.

<details>
<summary>Details</summary>

#### With `"spec.file": "LICENCE"`:
image

Success:
image

#### With `"spec.file": ""` & `"spec.files": ["LICENCE"]`:
image

Fails, even if f.files is the smae as above (???):
image

</details>


pkg/plugins/file/main.go
```diff
@@ -33,71 +48,135 @@ func New(spec Spec) (*File, error) {
...
+	// Return all the validation errors if any
 	if len(validationErrors) > 0 {
 		return fmt.Errorf("Validation error: the provided manifest configuration had the following validation errors:\n%s", strings.Join(validationErrors, "\n\n"))
 	}
 
 	return nil
 }
 
-// Read defines CurrentContent to the content of the file which path is specified in spec.File
+// TODO: to be superseded by ReadOrForceCreate
```
lemeurherve, line 122: No

lemeurherve:
file.yaml: success :)

pkg/plugins/file/main_test.go
```diff
@@ -91,6 +92,16 @@ func Test_Validate(t *testing.T) {
...
+			file.files = make(map[string]string)
+			if len(file.spec.File) > 0 {
+				file.files[strings.TrimPrefix(file.spec.File, "file://")] = ""
+			}
+			// files
+			for _, currentFile := range file.spec.Files {
+				// TODO:? warn if already in? (duplicates)
+				// TODO:! only add if not already in
+				file.files[strings.TrimPrefix(currentFile, "file://")] = ""
+			}
```
dduportal, line 104: you should edit the existing list of cases, or add new ones.

I don't see any reason to change the test logic?

lemeurherve: (Must be doing it wrong) It doesn't go into main.New(), and thus the files map is empty.

I also don't know why the mock returns Exists: false

In a nut shell, will need help ^^

dduportal: no problem. We can have a session later today, or tomorrow if you want? (contact me in 1:1)

dduportal: Oh I see, you were in the correct direction, my aplogizie I've read too quickly (did not unfold the unmodified code the first time).

WDYT about changing the method validate to have a `file.Spec` receiver instead of a `file.File` ?
That would drastically simplifies the unit test logic here, as the SUT would be the provided spec instead of File?

lemeurherve: I've realized I didn't needed the file.files for main_test.validate, it passes without it.

But I'm still struggling with main_test.TestFile_Read() where I needed to keep f.files, and source_test.TestFile_Source()
I don't see how to call main.New() (should I?) in order to get File.files initialized.

lemeurherve: Validating the Spec instead of the File now, better.

lemeurherve:
I didn't found the report of CodeQL:
image

.golangci.yml
```diff
@@ -105,6 +105,7 @@ issues:
     - path: _test\.go
       linters:
         - gomnd
+        - funlen
```
dduportal, line 108: 👍 good idea ! I thought it was enabled by default: it means that a lot of functions might starte to trigger alarms :)

pkg/plugins/file/main_test.go
dduportal: Good work, you're in the correct direction!

> But I'm still struggling with main_test.TestFile_Read() where I still need to initialize f.files, and source_test.TestFile_Source() (same kind of problem)

The idea of UT is to test an atomic piece of code without being influenced by the rest. The question (escept of the value of the test itself) in testing is always: what is the "System Under Test (SUT)", eg. what is the scope to be tested?

The unit tests in this case had been designed with the following rule in mind: Testing per public method of object (e.g. we don't test a full object, we don't test a package and we don't test private methods).

If you want to test a method such as `file.Read()`, then you have to build you own File{} object and then call the method on it. It allows testing not only inputs and outputs on the method only, but it also allows you to "tune" the object as you want (including mock injection instead of the default objects from the constructor) AND eventually checking the object (and mocks) state after calling the method to assert on mutations.


If you look at the existing tests, there should be the same pattern:
- Create a <whatever> object from the current test members (`tt.<something>` usually)
- Call the method to be tested on this object
- Assert on the results

.golangci.yml
lemeurherve: I had a lint error on a test function now too long

lemeurherve: And since it's a test it didn't seem judicious to remove test cases or to split this function, and I don't see the harm of disabling this rule for tests (in fact it's one of the exception cases given as example on the lint documentation)

dduportal: OK thabks for the explanation! Remind me something: might be related to the amount of managed cases for this pugin resource.
Looks good, what you said make perfect sense!

pkg/plugins/file/main_test.go
lemeurherve: > You might want to provide the prebuilt File{} object with it internal files map to the method to be tested.

Doh! Thanks for the explainations 👍 
Will push a fix.

dduportal: Tried my best, because I find it hard to write down these things (compared to a synchronous communication) so it might be incomplete or perfectible, but at least it is a start: hope it helps.

If it does not help, don't hesitate to reach me for a 1:1 virtual coffee  :)

lemeurherve: Something like fca52ff?

dduportal: Exactly ! The declaration syntax of a map is a bit heavy (could declare a custom private type if it is really annoying) but otherwise that's perfect

pkg/plugins/file/condition.go
```diff
@@ -59,95 +68,89 @@ func (f *File) condition(source string) (bool, error) {
...
-		if f.CurrentContent != source {
-			logrus.Infof(
-				"%s %s is different than the input source value:\n%s",
-				result.FAILURE,
-				logMessage,
-				text.Diff(f.spec.File, f.CurrentContent, source),
-			)
+		// If a matchPattern is specified, then return its result
+		if len(f.spec.MatchPattern) > 0 {
+			logrus.Debugf("Attribute 'matchpattern' found: %s", f.spec.MatchPattern)
```
lemeurherve, line 85:
```suggestion
			logrus.Debugf("Attribute 'MatchPattern' found: %s", f.spec.MatchPattern)
```

pkg/plugins/file/condition.go
```diff
@@ -59,95 +68,89 @@ func (f *File) condition(source string) (bool, error) {
...
+			reg, err := regexp.Compile(f.spec.MatchPattern)
+			if err != nil {
+				logrus.Errorf("Validation error in condition of type 'file': Unable to parse the regexp specified at f.spec.MatchPattern (%q)", f.spec.MatchPattern)
```
lemeurherve, line 88:
```suggestion
				logrus.Errorf("Validation error in condition of type 'file': Unable to parse the regexp specified at 'spec.MatchPattern' (%q)", f.spec.MatchPattern)
```

pkg/core/text/mocks.go
```diff
@@ -8,9 +14,14 @@ type MockTextRetriever struct {
 	Err      error
 	Line     int
 	Exists   bool
+	// Files    map[string]MockFile
 }
 
 func (mtr *MockTextRetriever) ReadLine(location string, line int) (string, error) {
+	// mtr.Files[location] = MockFile{
```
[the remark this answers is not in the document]
dduportal: The idea of this mock is to replace the "normal" behavior of [`text.Text{}`](https://github.com/updatecli/updatecli/blob/697302865ad39a3bc1a3a998f6ca08e2bdd45482/pkg/core/text/main.go#L29).

To allow "swapping" implementations, there is an interface [`text.TextRetriever`](https://github.com/updatecli/updatecli/blob/697302865ad39a3bc1a3a998f6ca08e2bdd45482/pkg/core/text/main.go#L21-L27) which defines the "contract" that any implementation must fulfill.

This is why the `text.MockTextRetriever{}` struct receives the methods `ReadLine()`, `ReadAll()` etc. : it's to fulfill the contract and being usable as alternative implementation (during test phases).

I'm not why you would want to change this mock unless you change the whole interface contract.

pkg/plugins/file/main_test.go
dduportal: I'm not sure to understand the goal: the mock is designed to manipulate a single file and I don't see a reason (at least at first glance) to change it, since the code of the `file` package is changed to manipulate the file iterations.

The "mock" is there to allow isolate the interactions of the method being tested with the "rest of the system".

lemeurherve: The current mock return data for only one file, whereas I need it to be able to return the expected content of multiple files.

lemeurherve: I realize I don't want the mock to return info about multiple files, but I also need to specify the expected data of multiple files in target tests, that's why I've been looking in this direction.

dduportal: > I realize I don't want the mock to return info about multiple files, but I also need to specify the expected data of multiple files in target tests, that's why I've been looking in this direction.

I don't understand what do you mean?

lemeurherve: See here for example: https://github.com/updatecli/updatecli/pull/490/files?file-filters%5B%5D=.go&show-deleted-files=true&show-viewed-files=true#diff-dab848426c72c878db3ec93c1fb1650241e6a9d1a300d928c8decb0d2c01c6eaR78-R99
I need to be able to mock multiple expected file data

lemeurherve: I think I found how: `wantMockState       map[string]text.MockTextRetriever`

dduportal: Mmmh the mock might need to be updated to allow providing a custom function that would return the content, a bit like it's done with the http mock: 
- Mock definition: (The DoFunc is public and provided by caller): https://github.com/updatecli/updatecli/blob/main/pkg/core/httpclient/mocks.go
- Example fo usage: https://github.com/updatecli/updatecli/blob/f6981b693219a756a861cdb843bb047c5b372c5d/pkg/plugins/docker/dockerregistry/main_test.go#L174-L210
  - The test provides the function with the logic "if this URL then return A, else return B".

=> you migh want to update the text mock to provide such behavior, so the function would have the list of files and would return the value associated to key. WDYT?

lemeurherve: So... Wrong variables probably (should be mock*?), but something like that: ead5407

dduportal: I might be misunderstanding something but, there is only 1 `text.textRetriever` on a given `File{}`: how would you have the collection of mockS (it's a map of string -> MockRetreiever{} that you defined)?

lemeurherve: I think I need a collection of text.textRetriever, one for each file path of `files`
(Don't pay too much attention to the code of the last 2 commits, was more to get a feel of the kind of struct needed)

dduportal: I got a proposal which is a compromise:

- Add the "function" for the `MockRetriever` and add a logic like "if function == nil then {use <old behavior of the mock>}"
- On your test suite, add a `mockFile map[string]string` attribute on the test case struct, and fill this (close to what you did If i'm not mistaken)
- Define the function to be used for the mock one time in `file/main_test.go` as a private function.
- In the test's for loop, pass the packae private function to the mock

=> with this, you don't break the actual mock contract (if it is used somewhere else on the code), and you don't have to repeat the function (as I suppose you will need this function for source tests, conditions tests and target tests.. at least.

WDYT?

lemeurherve:
Ready for review, finally :sweat_smile:

pkg/plugins/resources/file/main_test.go
```diff
@@ -103,77 +149,115 @@ func Test_Validate(t *testing.T) {
...
-		wantContent         string
-		wantMockState       text.MockTextRetriever
+		name           string
+		spec           Spec
+		files          map[string]string
+		mockedContents map[string]string
+		mockedLines    map[string]int
+		mockedError    error
+		wantedContents map[string]string
+		wantedLines    map[string]int
```
lemeurherve, line 159: Not sure about the pertinence of `mockedLines` and `wantedLines` now, they don't add anything IMO and should be removed.

lemeurherve:
:sob:
image

pkg/plugins/resources/yaml/condition_test.go
```diff
@@ -53,27 +59,30 @@ github:
...
   repository: charts
 `,
-			wantResult: false,
-			wantMockState: text.MockTextRetriever{
-				Location: "test.yaml",
+				},
 			},
 		},
+		// 		{
+		// 			name: "Failing case with keyonly and input source",
```
olblak, line 72: What is the reason to comment this test?

olblak:
> Ready for review, finally sweat_smile

I'll look into it

pkg/plugins/resources/yaml/condition_test.go
lemeurherve: Because this one and the following were resisting me ^^

pkg/plugins/resources/file/main_test.go
dduportal: Haven't check yet: weren't some source and condition tests using it?
(but I feel lik that your improvement of the mock might have had the possibility to make these attributes obsolete, and god knows how I love code being removed \o/)

lemeurherve: Some tests include them but I don't see the added values since the mock now returns and writes with the specified line parameter

lemeurherve: There are no changes between the initial mocked values and the wanted ones, I've created another PR to remove them, WDYT?
https://github.com/lemeurherve/updatecli/pull/18/files

pkg/plugins/resources/file/source.go
```diff
@@ -14,7 +14,11 @@ import (
 // Source return a file content
 func (f *File) Source(workingDir string) (string, error) {
 	var validationErrors []string
+	var foundContent string
 
+	if len(f.spec.Files) > 1 {
+		validationErrors = append(validationErrors, "Validation error in source of type 'file': the attributes `spec.files` can't contain more than one element for sources")
```
olblak, line 20: Wouldn't make sense to keep the parameter `file`? Instead of deprecating it? 
It could be transparent to the user

olblak: I am asking because I wonder if from a UX stand point it makes sense to ask for switching from `file: changelog.md` to `files: ["changelog.md"]`

lemeurherve: You could also see it changing from
```
file: changelog.md
```
to
```
files:
- changelog
```
which is IMO far more than acceptable.

Also, keeping the 2 would needlessly complicate the doc and the comprehension burden for new users.

lemeurherve: I'm still deprecating it ;)

dduportal: Fair. However I'm not convinced that the code would be overcomplicated to keep the singular version: you are already initializing the list of files from the single argument form. It means removing the warning and the depreciation comment and that should be enough

lemeurherve: Of course, but why keep the 2? I don't think it's really usefull in the long term

dduportal: In this case, it's the same as `kubectl get pod` and `kubectl get pods` having the same behavior: both are doing the same thing, and avoid user having to retype commands. For me it's UX

olblak:
Thanks for the pullrequest, that's a huge one.
I think I spot an error with the following manifest when using the line from a source

```
sources:
  scenario-1:
    name: "Scenario 1"
    kind: file
    scmID: scenario-1
    spec:
      file: README.md
      line: 1

scms:
  scenario-1:
    kind: git
    spec:
      password: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
      url: "https://github.com/updatecli/experiment.git"
      branch: "main"
```

pkg/plugins/resources/file/source.go
lemeurherve: OK

lemeurherve: I've removed the deprecation warning and comments: [`751ce51` (#490)](https://github.com/updatecli/updatecli/pull/490/commits/751ce5166607701afc6329f7a8d0eada2d0aa4d8)

olblak: For me it's more a UX thing, you are doing the right thing by converting from file to files so I would just remove the deprecation message and keep `file`

lemeurherve: I'll let you resolve this conversation then :)

olblak:
> I think I spot an error with the following manifest when using the line from a source

Well it can't reproduce it :/

olblak:
> > I think I spot an error with the following manifest when using the line from a source
> 
> Well it can't reproduce it :/

Well I can randomly reproduce it

olblak:
The result of my experimentation 
https://gist.github.com/olblak/88351cf2047520abe44eefb65380abdf

lemeurherve:
I've got this error with your manifest: https://gist.github.com/lemeurherve/4d00d36115c170a40899f8d699abbb91

lemeurherve:
@olblak reproduced and fixed :)

Please also review https://github.com/lemeurherve/updatecli/pull/18/files ;)

olblak:
Thanks @lemeurherve it works great. 
I am waiting for @dduportal to provide a final feedback before merging this pullrequest.

Also so, would you have some time to add a manifest to test the different "file" usecases in the [e2e](https://github.com/updatecli/updatecli/tree/main/e2e) directory?

Configuration cannot have WARNING or Error as defined in the venom test [here](https://github.com/updatecli/updatecli/blob/main/e2e/venom.d/test_diff.yaml)

olblak:
@lemeurherve linting test are failing :)

lemeurherve:
Looking into e2e errors

pkg/plugins/resources/dockerfile/condition_test.go
```diff
@@ -58,10 +60,7 @@ func TestDockerfile_Condition(t *testing.T) {
 					"matcher": "golang",
 				},
 			},
-			mockTest: text.MockTextRetriever{
-				Exists: false,
-			},
-			wantErr: fmt.Errorf("the file NOTEXISTING.Dockerfile does not exist"),
+			wantErr: fmt.Errorf("The file NOTEXISTING.Dockerfile does not exist."),
```
olblak, line 63:
```suggestion
			wantErr: fmt.Errorf("the file NOTEXISTING.Dockerfile does not exist"),
```

olblak: Linter complain about error message starting with capital letter and ending with punctuation point. I guess is because error message can be merged

lemeurherve:
The e2e error comes from `udpatecli version`:
```bash
$ dist/updatecli_darwin_amd64/updatecli version

VERSION

Application:

Build Time :

$ updatecli version

VERSION

Application:    0.22.3
Golang     :    1.17.8 linux/amd64
Build Time :    2022-03-22T09:52:05Z
```

Not sure why though

olblak:
> The e2e error comes from `udpatecli version`:
> 
> ```shell
> $ dist/updatecli_darwin_amd64/updatecli version
> 
> VERSION
> 
> Application:
> 
> Build Time :
> 
> $ updatecli version
> 
> VERSION
> 
> Application:    0.22.3
> Golang     :    1.17.8 linux/amd64
> Build Time :    2022-03-22T09:52:05Z
> ```
> 
> Not sure why though

It depends who you build updatecli locally, because you may not have set the flag that define the different version information

olblak:
> OK. Do the e2e tests pass locally for you?

Yes it does 
```
 • Updatecli Version TestSuite (e2e/venom.d/test_version.yaml)
 	• Test-updatecli-version-return-code-is-0 SUCCESS

```

olblak:
Maybe @dduportal can help you in case of something specific on mac which I am missing

olblak:
Anyway e2e test are executed for each pullrequest in the build workflow so you'll know very quickly if something goes wrong. :D

dduportal (approved):
LGTM 🚀

PR #491, opened 2022-01-26
Support automerge for Github PullRequest

### LINKED ISSUE #488: Feature Request: Support Automerge for GitHub Pullrequest

Using Github PullRequest workflow is useful to get reviews and to trigger GitHub action jobs.
Sometimes we are only interested by the GitHub Action, so if tests are passing then we can safely merge the change which reduce alert fatigue for safe dependency update
 
It would be useful to automatically merge Pull Request on git repository with a lot of testing place.

### Solution you'd like

I would like to assign "auto-merge" to pullrequest created by updatecli such as in the following example 

```
pullrequests:
  helm-charts:
    kind: "github"
    scmID: "helm-charts"
    targets:
      - "helm-charts"
    spec:
      automerge: true
      labels:
        - "dependencies"
```

_No response_

### Anything else?

It seems to be possible now using the Graphql API
-> https://docs.github.com/en/graphql/reference/mutations#enablepullrequestautomerge

olblak:
# Support automerge for Github PullRequest

Fix #488 

This PR introduces an new parameter in Github PullRequest spec named "automerge".
When set to true, updatecli enables automerge to pullrequest that it creates.

## Test

```shell
cp <to_package_directory>
go test
```

### Tradeoff

At this stage, I missing the ability to define the merge method.


olblak:
I did some quick testing and it works **but**
The github API doesn't return a clear message when auto-merge doesn't meet requirements.
If auto-merge is not enabled at the repository level or if the PR is not in a blocked state due to a branch restriction, for instance, the API just returns "Pull request is not in the correct state to enable auto-merge"

Also, I am now wondering if we should also merge the PR even if the Github auto-merge feature is not available.

olblak:
I am still missing the ability to set the merge type like rebase/squash/merge

lemeurherve:
> I am still missing the ability to set the merge type like rebase/squash/merge

I think changing `merge_method` to `squash` or `rebase` in addition  to auto-merge should do the trick: https://docs.github.com/en/rest/reference/pulls#merge-a-pull-request

lemeurherve:
> I did some quick testing and it works **but** The github API doesn't return a clear message when auto-merge doesn't meet requirements. If auto-merge is not enabled at the repository level or if the PR is not in a blocked state due to a branch restriction, for instance, the API just returns "Pull request is not in the correct state to enable auto-merge"

For this check we can retrieve the `allow_squash_merge` value of the repository before enabling auto-merge: https://docs.github.com/en/rest/reference/repos#get-a-repository

> `allow_auto_merge`: Either `true` to allow auto-merge on pull requests, or `false` to disallow auto-merge. Default: `false`

lemeurherve:
> Also, I am now wondering if we should also merge the PR even if the Github auto-merge feature is not available.

I'm not sure I understand what do you mean here, could you expand a little bit on this?

olblak:
> Also, I am now wondering if we should also merge the PR even if the Github auto-merge feature is not available.

Let's say a updatecli configuration contains `automerge: true`

Scenario 1: The repository allows auto-merge pullrequest as defined at the repository level. So we enable the flag "automerge" which is the goal of the current pullrequest

Scenario 2: The repository disallows automerge, so updatecli just merges the pullrequest without using the automerge feature. The usecase that I see is if a project use release-drafter to generate changelog based on pull request but doesn't care about reviewing pull requests generated by updatecli.

lemeurherve:
> The repository disallows automerge, so updatecli just merges the pullrequest without using the automerge feature. 

I wouldn't do that and bypass the repository security setting.  
I would instead fail or warn in the output.

dduportal:
+1 with @lemeurherve . You are trying to solve multiple problems at once. The purpose of the PR here is to enable the "automerge" feature of GitHub pullrequests: let's keep that scope.

> Scenario 2: The repository disallows automerge, so updatecli just merges the pullrequest without using the automerge feature.

=> In this case, I expect `updatecli` to have the PR opened BUT to have failed with an error message that tells the end user that automerge is not allowed on this repository.

> I am still missing the ability to set the merge type like rebase/squash/merge

Do you mind proposing different YAML examples to show the different cases:

- No automerge (current case)
- Enable automerge (and let GitHub handle the default behavior)
- Enable automerge with squash
- Enable automerge with rebase
- Enable automerge with merge

Are there other cases?

pkg/plugins/github/pullrequest.go
```diff
@@ -247,6 +255,28 @@ func (p *PullRequest) updatePullRequest() error {
...
 }
 
+// EnablePullRequestAutoMerge updates an existing pullrequest with the flag automerge
+func (p *PullRequest) EnablePullRequestAutoMerge() error {
+
+	var mutation struct {
+		EnablePullRequestAutoMerge struct {
+			PullRequest PullRequestApi
+		} `graphql:"enablePullRequestAutoMerge(input: $input)"`
+	}
```
dduportal, line 265: The code looks good. Eventhough the "pullrequest" object is not ready for (yet) for dependency injection to allow decoupling from githubv4 + mocking for unit test, do you mind defining this type as package-private in the same way as it is done in the githib plugin: https://github.com/updatecli/updatecli/blob/main/pkg/plugins/github/release.go#L46-L66 
?

It would allow factorization a bit easier in the future (and the tests, either integ. or unit could access the type)

lemeurherve:
>>  I am still missing the ability to set the merge type like rebase/squash/merge
> Do you mind proposing different YAML examples to show the different cases

I wouldn't even bother, as it's in itself a new feature, independent of the automerge, the possibility to set the merge_method

lemeurherve:
> I'm only asking to see examples "as a end user" :)

Agreed, for those with this new mergeMethod parameter on a separate PR though?

pkg/plugins/github/pullrequest.go
olblak: I am not sure to understand what you are asking for. Are you asking me to refactor the code in pullrequest.go file?

olblak:
> +1 with @lemeurherve . You are trying to solve multiple problems at once. The purpose of the PR here is to enable the "automerge" feature of GitHub pullrequests: let's keep that scope.

It depends who you see it hence my question :p 
If you put "automerge: true" in an updatecli configuration.
Are you expecting to merge automatically the pullrequest or to enable "automerge" feature on the pullrequest. The latter case expect you to understand what the automerge flag from a GitHub point of vue.

> Do you mind proposing different YAML examples to show the different cases:
While I don't think it's mandatory in this PR, I think it would be useful to identify how the configuration will evolve

I was thinking something like 

```
pullrequests:
  helm-charts:
    kind: "github"
    scmID: "helm-charts"
    targets:
      - "helm-charts"
    spec:
      automerge: true
      mergemethod: squash # available options are sqash, merge,rebase
      labels:
        - "dependencies"
        - "epinio"
```

olblak:
> Agreed, for those with this new mergeMethod parameter on a separate PR though?

I agree but I would be interested to identify if it affects the "automerge" wording

olblak:
Anyway the merge method will not be implemented in this PR because GitHub API v4 documentation warn that it's evolving 

```
Preview notice
mergeStateStatus is available under the Merge info preview. During the preview period, the API may change without notice.
```

So I am mainly interested to identify blockers

lemeurherve:
> Are you expecting to merge automatically the pullrequest or to enable "automerge" feature on the pullrequest.

No more than Github offer as native feature, the activation of the "auto-merge" github feature, no less no more.

dduportal:
> > +1 with @lemeurherve . You are trying to solve multiple problems at once. The purpose of the PR here is to enable the "automerge" feature of GitHub pullrequests: let's keep that scope.
> 
> It depends who you see it hence my question :p If you put "automerge: true" in an updatecli configuration. Are you expecting to merge automatically the pullrequest or to enable "automerge" feature on the pullrequest. The latter case expect you to understand what the automerge flag from a GitHub point of vue.
> 
> > Do you mind proposing different YAML examples to show the different cases:
> > While I don't think it's mandatory in this PR, I think it would be useful to identify how the configuration will evolve
> 
> I was thinking something like
> 
> ```
> pullrequests:
>   helm-charts:
>     kind: "github"
>     scmID: "helm-charts"
>     targets:
>       - "helm-charts"
>     spec:
>       automerge: true
>       mergemethod: squash # available options are squash, merge,rebase
>       labels:
>         - "dependencies"
>         - "epinio"
> ```

I expect the "automerge" of the "GitHub Pull Request" object in updatecli to map to the "automerge" feature of... GitHub Pull Requests.

Because in the alternativ scnearion (e.g. updatecli merges immediately after opening the PR) seems unrealistic or really really edgy and I'm not sur to see why updatecli should do it:

- PR have branch protection quite often: that would be against the protections (and really hard to document 
- I don't really see the value of creating a PR to be merged compared to commit-pushing directly. I understand that it would mean release-drafter  cannot be used but there are so much tools to generate changelogs based on git commit messages that a PR in such cases looks really a hack
- What about the PR checks? How can you manage updatecli to wait for the check to complete: this would be an "asynchronous" task while updatecli is built as a synchronous system from its core

lemeurherve:
> I agree but I would be interested to identify if it affects the "automerge" wording

It's two distinct things IMO, you can set merge/squash/rebase independently of auto-merge. The combined github UI isn't mandatory

pkg/plugins/github/pullrequest.go
dduportal: I'm only asking you to declare the type of the variable `mutation` as a package type (but a private one).

So that mutation declaration would be something like:

```
var mutation mutationQuery
```

However if you want to jump on the "unit test" wagon, you can refactor the pullrequest.go ;)

lemeurherve:
For the merge-method feature I would state its value in the output: "merge-method: merge"
And for this other feature I would just state "auto-merge: enabled"

lemeurherve:
> Anyway the merge method will not be implemented in this PR because GitHub API v4 documentation warn that it's evolving

Not sure a lot would change, and we don't know when, should be trivial to adapt when needed, I would go for it anyway.

WDYT?

olblak:
> Not sure a lot would change, and we don't know when, should be trivial to adapt when needed, I would go for it anyway.

I could go with it I just have to find the correct API call

olblak:
> Because in the alternativ scnearion (e.g. updatecli merges immediately after opening the PR) seems unrealistic or really really edgy and I'm not sur to see why updatecli should do it:
> 
>     * PR have branch protection quite often: that would be against the protections (and really hard to document
> 
>     * I don't really see the value of creating a PR to be merged compared to commit-pushing directly. I understand that it would mean release-drafter  cannot be used but there are so much tools to generate changelogs based on git commit messages that a PR in such cases looks really a hack
> 
>     * What about the PR checks? How can you manage updatecli to wait for the check to complete: this would be an "asynchronous" task while updatecli is built as a synchronous system from its core

I am fine relying on the GH automerge feature. it's more secure by default because you have to explicitly configure branch protection and less code for me now.

olblak:
For the automerge key, I would go with `automerge: true` as it's boolean with default state to false

dduportal:
Thanks a lot @olblak for the brainstorming and the code ❤️ 
Thanks a lot @lemeurherve for the ideas, feedbacks, proposal (and the issue for the automerge type ;) )

pkg/plugins/github/pullrequest.go
olblak: As requested, I created the package struct `mutationEnablePullRequestAutoMerge`. I won't have the time to refactor the package to implement UT so I propose to move forward with the current state.

olblak:
Thanks, everybody for the reviews. This PR is ready to be shipped as is, I am preparing a PR to update the document. 
I propose to target 0.19 release. @lemeurherve @dduportal any objections?

dduportal (approved):
Looks good to me. Let's proceed for 0.19.0 :) 

Side note: we have the e2e test, but we'll have to consider adding unit tests or block PR in the near future, to ensure that we can trust the changes and keep releasing often without breaking (ref. https://en.wikipedia.org/wiki/Swiss_cheese_model)

lemeurherve:
No check of allow_auto_merge on the repo prior to the mutation ?

dduportal:
> No check of allow_auto_merge on the repo prior to the mutation ?

subsequent PR that would fix the issue you created (or did I misundertood)?

lemeurherve:
The issue I've created concerns the merge_method, no relation to PR automerge nor repo allow_auto_merge

dduportal:
cc @olblak ^^

olblak:
> The issue I've created concerns the merge_method, no relation to PR automerge nor repo allow_auto_merge
> 
> For me this last check has to be done in order to validate or warn the user trying to set automerge on a PR

This requires significantly more because we need to get repository information to know  if "automerge" is allowed at the repository level and verify that branch protection exists on the targeted branch. 
Imho it's a UX issue that should be tracked on a specific issue. 
What do you think @lemeurherve

lemeurherve:
Don't we already have access to the repository via `scms` since this feature concerns only github?
Is it really significant more work and not just a verification call on the repo specified in values?

lemeurherve:
It could be added later as it's only silently failing, not that big of a deal, but it would be better to have this included by default.

dduportal:
I just want to point that if a part of a feature require significant work, then it means that we should restrain to merge it even if we are interested. I approved this pr so i am culprit also but we need to be careful in the future

PR #493, opened 2022-01-30
chore(yaml) cleanup file generated during test phase

dduportal:
This is a "quick" PR that cleans up the files generated fby thet target tests for YAML resource.

It was driving me (and @lemeurherve) crazy that running `go test ./... -short` was generating a file in the source code.

Since I haven't had time on the topic for unit tests with mock for YAML, let's do it that way.

Labelled as "skip-changelog" since it does not hold anything useful for end user

## Test

```shell
$ git status
# No git changes
$ go test ./... -short
# Should be success
$ git status
# No git changes
```

### Tradeoff

- Could execute the tests in a temp directory

### Potential improvement

- Execute mocked tests without any interaction with the I/O file system (real unit tests)

### DIFF
--- a/pkg/plugins/yaml/target_test.go
+++ b/pkg/plugins/yaml/target_test.go
@@ -109,6 +109,7 @@ github-
 				contentRetriever: &mockText,
 			}
 			gotResult, gotErr := y.Target(tt.inputSourceValue, tt.dryRun)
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):
🥳

PR #516, opened 2022-02-09
Remove warning

olblak:
I guess after 3700 commits and 3558 pull request we can remove the warning


dduportal (approved):
LGTM.

Feel free to merge without waiting for build checks:

- golang 1.17.7 has just been tagged but not yet release, which makes the updatecli golang failing
- It's only markdown change

PR #527, opened 2022-02-15
fix(condition) do not execute target if a condition fails

### LINKED ISSUE #526: Target is executed even if one condition is failing

### Current Behavior

When I run `updatecli apply --config updatecli/updatecli.d/epinio.yaml  --debug --values updatecli/values.yaml` on the updatecli repository from https://github.com/olblak/epinio-helm-chart
I get  
```
REPORTS:


⚠ EPINIO.YAML:
	Sources:
		✔ [epinio] Get Latest epinio version (kind: githubRelease)
	Condition:
		✗ [dockerImage] Check that ghcr.io/epinio/epinio-server:0.2.1 is published (kind: dockerImage)
		✔ [dockerImageContainer] Check that ghcr.io/epinio/epinio-server is used in Helm chart epinio (kind: yaml)
	Target:
		⚠ [helm-charts] Update epinio version in chart epinio (kind: helmChart)



Run Summary
===========
Pipeline(s) run:
  * Changed:	1
  * Failed:	0
  * Skipped:	0
  * Succeeded:	0
  * Total:	1


```

### Expected Behavior

I am expecting the pipeline to be skipped as it used to be.
```
REPORTS:


⚠ EPINIO.YAML:
	Sources:
		✔ [epinio] Get Latest epinio version (kind: githubRelease)
	Condition:
		✗ [dockerImage] Check that ghcr.io/epinio/epinio-server:0.2.1 is published (kind: dockerImage)
		✔ [dockerImageContainer] Check that ghcr.io/epinio/epinio-server is used in Helm chart epinio (kind: yaml)
	Target:
		-  [helm-charts] Update epinio version in chart epinio (kind: helmChart)



Run Summary
===========
Pipeline(s) run:
  * Changed:	0
  * Failed:	0
  * Skipped:	1
  * Succeeded:	0
  * Total:	1
```

### Steps To Reproduce

When I run `updatecli apply --config updatecli/updatecli.d/epinio.yaml  --debug --values updatecli/values.yaml` on the updatecli repository from https://github.com/olblak/epinio-helm-chart

### Environment

```markdown
- OS: Linux
- updatecli: v0.19.2 and master branch
```

### Anything else?

_No response_

dduportal:
Fix #526 

This PR revamps what I broke in #470:

- No more updating the pipeline configuration (as it's only required after source execution: there is no point in rendering the template after each condition as for today)
- Keep showing the error (if any) of a condition execution (that was the value of #470)
- Fix the "flow" that was broken by #470 by setting the global result to "false" if there is an error OR if the conditions failed (e.g. did not error but returned with "false").

## Test

Please note that a new e2e test suite had been added to catch this case (non regression in the future), as the "flow" tests are hard to test today (pkg/core/pipeline is not unit-tested a lot).

```shell
make test
make test-short
make test-e2e
```

### DIFF
--- a/pkg/core/pipeline/conditions.go
+++ b/pkg/core/pipeline/conditions.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/result"
 )
 
 // RunConditions run every conditions for a given configuration config.
@@ -42,21 +43,18 @@ func (p *Pipeline) RunConditions() (globalResult bool, err error) {
 			p.Sources[condition.Config.SourceID].Config.Prefix +
 				p.Sources[condition.Config.SourceID].Output +
 				p.Sources[condition.Config.SourceID].Config.Postfix)
-
 		if err != nil {
-			// Show error to end user if any
+			// Show error to end user if any but continue the flow execution
 			logrus.Error(err)
-			// Fail globally (even though other conditions will be executed)
-			globalResult = false
-		} else {
-			// Update pipeline's configuration after each condition run
-			err = p.Config.Update(p)
-			if err != nil {
-				globalResult = false
-				return globalResult, err
-			}
 		}
+
+		// Reports the result of the execution of this condition
 		rpt.Result = condition.Result
+
+		// If there was an error OR if the condition is not successful then defines the global result as false
+		if err != nil || condition.Result != result.SUCCESS {
+			globalResult = false
+		}
 
 		p.Conditions[id] = condition
 		p.Report.Conditions[id] = rpt
--- a/e2e/skipped.targets.d/multiple-conditions.yaml
+++ b/e2e/skipped.targets.d/multiple-conditions.yaml
@@ -1,0 +1,36 @@
+title: Skip targets with only one (failing) condition
+
+sources:
+  default:
+    name: Default Source (dummy)
+    kind: shell
+    spec:
+      command: echo DUMMY
+
+conditions:
+  condition1:
+    name: This conditions always succeed
+    sourceID: default
+    kind: shell
+    spec:
+      command: "true"
+  condition2:
+    name: This conditions always fail
+    sourceID: default
+    kind: shell
+    spec:
+      command: "false"
+  condition3:
+    name: This conditions always succeed
+    sourceID: default
+    kind: shell
+    spec:
+      command: "true"
+
+targets:
+  shouldNeverRun:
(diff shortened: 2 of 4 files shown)


olblak (approved):
Thanks for the quick fix

olblak:
@dduportal  it's still in draft mode, are you waiting for something?

olblak (changes requested):
This doesn't fix it

pkg/core/pipeline/conditions.go
```diff
@@ -43,20 +44,24 @@ func (p *Pipeline) RunConditions() (globalResult bool, err error) {
...
+			// Update pipeline after each condition run
 			err = p.Config.Update(p)
 			if err != nil {
 				globalResult = false
 				return globalResult, err
 			}
 		}
-		rpt.Result = condition.Result
+
+		if err != nil {
```
olblak, line 58:
```suggestion
		if err != nil || strings.Compare(condition.Result, result.SUCCESS) != 0 {	
```

Otherwise the globalresult will be set to true if there is not error but that the condition failed

olblak: and `logrus.Error(err)` should only be execute if err != nil

dduportal:
> @dduportal it's still in draft mode, are you waiting for something?

Yep, my fix does not correct the issue + there is a edge case that I want to test manually.
I wanted to add an e2e test to cover this case but it's not mandatory (I would prefer to delay the e2e).

pkg/core/pipeline/conditions.go
dduportal: updated: WDYT?

dduportal:
- Added an E2E "basic suite"
- Took your feedback in account + removed the "config.update" as it is not needed

pkg/core/pipeline/conditions.go
```diff
@@ -42,22 +43,19 @@ func (p *Pipeline) RunConditions() (globalResult bool, err error) {
...
-			if err != nil {
-				globalResult = false
-				return globalResult, err
-			}
 		}
+
+		// Reports the result of the execution of this condition
 		rpt.Result = condition.Result
 
+		// If there was an error OR if the condition is not successful then define the global result as false
```
lemeurherve, line 54:
```suggestion
		// If there was an error OR if the condition is not successful then defines the global result as false
```

lemeurherve (dismissed):

olblak:
Very nice for the e2e test

lemeurherve (approved):

olblak (approved):

PR #537, opened 2022-02-23
fix: log the config name at validation

### LINKED ISSUE #536: Feature Request: add the concerned manifest filename in the error output

When updatecli is running on a folder of manifests like in https://github.com/jenkins-infra/kubernetes-management/pull/2048, if there is an error with one of them their filename is not shown in the error logs.

Ex: https://infra.ci.jenkins.io/blue/organizations/jenkins/kubernetes-management/detail/PR-2048/1/pipeline

```
updatecli diff --config ./updatecli/updatecli.d --values ./updatecli/values.yaml

+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/charts/accountapp.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/acme.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/autoscaler.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/aws-node-termination-handler.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/certmanager.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/codecentric-keycloak.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/custom-distribution-service.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/datadog.yaml"
Loading Pipeline "updatecli/updatecli.d/charts/env-jenkins-release.yaml"
[...]
Loading Pipeline "updatecli/updatecli.d/docker-images/datadog.yaml"
Loading Pipeline "updatecli/updatecli.d/docker-images/jenkins-lts.yaml"
Loading Pipeline "updatecli/updatecli.d/docker-images/jenkins-weekly.yaml"
Loading Pipeline "updatecli/updatecli.d/docker-images/keycloack-theme.yaml"
Loading Pipeline "updatecli/updatecli.d/docker-images/nginx.yaml"
Loading Pipeline "updatecli/updatecli.d/pod-templates/helmfile.yaml"
ERROR: the specified target "updateChartVersion" for the pull request "default" does not exist
ERROR: wrong updatecli configuration
```

### Solution you'd like

Have the concerned manifest filename in the error output:

```
Loading Pipeline "updatecli/updatecli.d/docker-images/nginx.yaml"
Loading Pipeline "updatecli/updatecli.d/pod-templates/helmfile.yaml"
ERROR: the specified target "updateChartVersion" for the pull request "default" does not exist
ERROR: wrong updatecli configuration in "updatecli/updatecli.d/charts/datadog.yaml"
```

It can also be added to the first error line, not sure about this one.

Cry, then search along modified files for the one(s) with mentioned ids.

### Anything else?

_No response_

lemeurherve:
Fixes #536

Log the config name (which default value is the manifest filename) so when updacli is running on a folder and there is an error it's easier to know which manifest has a problem.

## Test

```shell
cp <to_package_directory>
go test
```

### Tradeoff

One additional line per manifest in the output

### Potential improvement (and fix of the tradeoff)

This log should even replace the one line 82: https://github.com/updatecli/updatecli/blob/01da81e6c25062cd9f902e173208b766fa4cbdee/pkg/core/config/main.go#L82

WDYT? In that case, this PR could be "fix: log the config name at validation instead of at loading"

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -131,6 +131,8 @@ func (config *Config) Display() error {
 // Validate run various validation test on the configuration and update fields if necessary
 func (config *Config) Validate() error {
 
+	logrus.Infof("Validating %q", config.Name)
+
 	if config.Source.Kind != "" && (len(config.Sources) > 0) {
 		logrus.Errorf("Source and Sources can't be defined at the same time, please use the 'Sources' syntax")
 		return ErrBadConfig


olblak (approved):

lemeurherve:
> WDYT? In that case, this PR could be "fix: log the config name at validation instead of at loading"

@olblak I wanted to include it in this PR but waited for your return before doing so. No worry I can do it in another PR, but I'd still like to hear your opinion before :)

IMO it would be better as it would simply replace the current log line done when the files are loaded by one done when each file is validated.

olblak:
Sorry that I missed your question, I should have take more care before merging your pullrequest.

IMHO the message from line 82 `logrus.Infof("Loading Pipeline %q", cfgFile)` contains a different information than the pipeline name so it makes sense to have it. We want to easily identify what went wrong when trying to read the file.

Your PR is about identifying what went wrong when validating the pipeline from the file

For a reason that I ignore the line 105 is duplicated with 114. The line 105 should be removed.

Imho your change should have been a debug and not an info.

**thinking loud**
I am wondering if it wouldn't be better to notify which pipeline we are in from the engine package 
https://github.com/updatecli/updatecli/blob/5afdcfbc0c5f1ca88ebef3da3ec876ada884ed8a/pkg/core/engine/main.go#L145

Issue #410, opened 2021-12-10
Change non absolute paths to resource to be relative to YML manifest instead of current working dir?


dduportal:
**Is your feature request related to a problem?**
 
As underlined by @olblak in https://github.com/updatecli/updatecli/pull/388#issuecomment-991077726, the relative path for resources is resolved from the directory where the command `updatecli` command is run (e.g. the "current working directory").

This behavior is problematic as soon as you change this "current working dir": either you're browsing from a subdirectory or using absolute path.

Example, with the following YAML manifest:

```yaml
# config.yaml
sources:
  readFromFile:
    kind: file
    spec:
      file: ./updatecli/file.txt
targets:
  printsContent:
    kind: shell
    spec:
      # Echoes the source value
      command: echo
```

```shell
$ cat updatecli/file.txt 
Hello
$ tree ./updatecli
./updatecli
├── config.yaml
├── file.txt

## Success
$ updatecli diff --config ./updatecli/config.yaml

# ...
SOURCES
=======

readFromFile
------------
✔ Content: found from file "./updatecli/file.txt":
Hello
# ...

## Failure because relative path
$ cd ./updatecli/
$ updatecli diff --config ./config.yaml

# ...
SOURCES
=======

readFromFile
------------
ERROR: ✗ Validation error: the provided manifest configuration had the following validation errors:
Invalid spec for yaml resource: the file "./updatecli/file.txt" does not exist.
# ...

## Failure because relative path
$ cd ..
$ PROJECT_DIR=$(pwd)
$ cd /tmp
$ updatecli diff --config "${PROJECT_DIR}"/updatecli/config.yaml

# ...
readFromFile
------------
ERROR: ✗ Validation error: the provided manifest configuration had the following validation errors:
Invalid spec for yaml resource: the file "./updatecli/file.txt" does not exist.
# ...
```

**Describe the solution you'd like**

It would make sense to resolve relative path from the YAML manifest files provided as config: it would solve all the issues here.

**However** that would break a lot of actual configurations.

**Describe alternatives you've considered**
N.A.

**Additional context**
Trying to be "smart" by trying to autodetect if the path would wether be "relative to manifest, otherwise to CWD" could be risky (weird "magical" code).

lemeurherve:
There seem to be some copy/pastes left behind, all 3 updatecli commands are the same currently.

lemeurherve:
This FR seems a little bit too magic to me, is there other cli tools with the same behavior?

[...]

PR #543, opened 2022-03-01
Test at runtime if a yaml file exists

olblak:

# Test at runtime if a YAML file exists

There are situations where a first step is required such as cloning a git repository before validating that a file exists.
So instead I prefer to test that a file exists at runtime for the source/condition/target.

## Test

To test this pull request, we can run updatecli command with whatever updatecli manifest as long as we don't execute updatecli from the root of the git directory containing the updatecli manifest. 

### Tradeoff

This pull request doesn't fix https://github.com/updatecli/updatecli/issues/410 but instead, it moves the `fileexist` to a function executed at runtime.

### DIFF
--- a/pkg/plugins/resources/yaml/main.go
+++ b/pkg/plugins/resources/yaml/main.go
@@ -71,17 +71,21 @@ func (y *Yaml) Validate() error {
 	}
 	if y.spec.File == "" {
 		validationErrors = append(validationErrors, "Invalid spec for yaml resource: 'file' is empty.")
-	} else {
-		if !y.contentRetriever.FileExists(y.spec.File) {
-			validationErrors = append(validationErrors, fmt.Sprintf("Invalid spec for yaml resource: the file %q does not exist.", y.spec.File))
-		}
+		// As explained on https://github.com/updatecli/updatecli/issues/410 , we can't always test if a file exists.
+		// There are situation where a first step is required such as cloning a git repository before validating that a file exists.
+		// Hence why I am commenting the following the code until https://github.com/updatecli/updatecli/issues/410 is fixed.
+		// And I am testing at runtime in source/condition/target that the file exist.
+		//	} else {
+		//		if !y.contentRetriever.FileExists(y.spec.File) {
+		//			validationErrors = append(validationErrors, fmt.Sprintf("Invalid spec for yaml resource: the file %q does not exist.", y.spec.File))
+		//		}
 	}
 	if y.spec.Key == "" {
 		validationErrors = append(validationErrors, "Invalid spec for yaml resource: 'key' is empty.")
 	}
 	// Return all the validation errors if found any
 	if len(validationErrors) > 0 {
-		return fmt.Errorf("Validation error: the provided manifest configuration had the following validation errors:\n%s", strings.Join(validationErrors, "\n\n"))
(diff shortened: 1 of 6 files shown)


pkg/plugins/resources/yaml/main.go
```diff
@@ -71,17 +71,21 @@ func (y *Yaml) Validate() error {
...
-			validationErrors = append(validationErrors, fmt.Sprintf("Invalid spec for yaml resource: the file %q does not exist.", y.spec.File))
-		}
+		// As explained on https://github.com/updatecli/updatecli/issues/410 , we can't always test if a file exists.
+		// There are situation where a first step is required such as cloning a git repository before validating that a file exists.
+		// Hence why I am commenting the following the code until https://github.com/updatecli/updatecli/issues/410 is fixed.
+		// And I am testing at runtime in source/condition/target that the file exist.
+		//	} else {
+		//		if !y.contentRetriever.FileExists(y.spec.File) {
+		//			validationErrors = append(validationErrors, fmt.Sprintf("Invalid spec for yaml resource: the file %q does not exist.", y.spec.File))
+		//		}
```
dduportal, line 81:
```suggestion
```

Git is quite useful for such use cases :) If you comment, better to delete

pkg/plugins/resources/yaml/target.go
```diff
@@ -82,12 +87,13 @@ func (y *Yaml) target(source string, dryRun bool) (bool, []string, string, error
 	if !dryRun {
 
 		newFile, err := os.Create(y.spec.File)
-		defer newFile.Close()
```
dduportal, line 85: What is the reason to remove this line?

dduportal (changes requested):
LGTM, except the defer line (re)moved

pkg/plugins/resources/yaml/target.go
lemeurherve: Moved to L95

dduportal: yep: the line was moved. But the new position is a file handler leak: if the error is not null BUT the file handler is still opened (even in memory: case of a read only filesystem for instance), then the app will leak because the handler won't be closed even after the execution, and only a reboot of the instance would clean it up

olblak: I added an instruction to ignore the linter check

lemeurherve (dismissed):

pkg/plugins/resources/yaml/target.go
dduportal: Interesting, the static lint check allows to make sure that my assertion was wrong. Sounds like golang id able to catch the case I described during compilation, so your change made sense (to move the line) as it implies a better code readability. Don’t hesitate to explain sich change with a static link next time

dduportal (dismissed):
Approving as i should not have blocked earlier and lgtm

olblak:
Of course, yaml unit test can work anymore since we test if a file should exist :/

dduportal:
> Of course, yaml unit test can work anymore since we test if a file should exist :/

Yep, that is the purpose of unit tests :) Delete the unit test (that should have been done even before commenting out, ref. TDD :) )

olblak:
> Yep, that is the purpose of unit tests :) Delete the unit test (that should have been done even before commenting out, ref. TDD :) )

Well done for the unit test and the mock interface, I could easily fix the broken test \o/

Issue #517, opened 2022-02-10
Feature Request: give a way to capture source part with MatchPattern


smerle33:
### Is your feature request related to a problem?

I though I could use matchPattern from within sources to get a specific capture on a multi line regexp but it seems not to work as expected. 

`matchPattern: "aws_security_group.*\n.*name.*\"(.*)\".*\n.*infra\\.ci"`

(the regexp as been tried on regexp101.com as golang flavor with success)

let me assume that the source.SOURCENAME would be the captured part of the regexp "(.*)" but ... it's not ....

### Solution you'd like

get an option -regexpSubstr for example that allow regexp capture
I also tried the 
`transformers:
      - findSubMatch: 'name.*"(.*)"'`
   
with no luck

### Anything else?

I am probably mystaking something somewhere but no way to find out ... and it seems related to the multiline concept

olblak:
Hi, 
I thought it was already working
Have you tried using "findSubMatch"

olblak:
Ha I missed that you already tried findSubMatch, I need to check

smerle33:
Hi,
I see 2 ways of solving this within the transformers, 
   - first, just change line : https://github.com/updatecli/updatecli/blob/5afdcfbc0c5f1ca88ebef3da3ec876ada884ed8a/pkg/core/transformer/main.go#L127 with output = found[1] (instead of output = found[0])
   
   - second, offering the opportunity to choose the slice index that may be return by adding a parameter to the call to findSubMatch and then findSubMatch may look like replace ....

dduportal:
After discussions with the 2 of you, here is the proposal:

- (new feature) Allow to specify the index of the capture. It means that the transformer does not return the matched line, but **only** 1 capture. No more, no less.

```yaml
transformers:
  - findSubMatch: 
      # Note that it is indented, like the "replace"
      pattern: '"(.*?):(.*)'
      captureIndex: 1
```

- (new feature) Introduces a `findLine` transformer which behaves the same as the actual `findSubMatch` but with a clearer naming: it returns the line(s) matched by the provided regexp pattern.

```yaml
transformers:
  - findLine: '^key:.*$' # Returns all lines matching the provided pattern
```

[...]

PR #551, opened 2022-03-03
Feat/capture Adding a parameter for the transformer FindSubMatch

smerle33:
# Title

Fix [#517 ](https://github.com/updatecli/updatecli/issues/517)

## Adding a parameter for the transformer FindSubMatch

```shell
cd pkg/core/transformer/
go test
```

### Potential improvement

This pull request permit 2 ways of calling the FindSubMatch: 

New way : 

```
transformers:
  - findSubMatch: 
      # Note that it is indented, like the "replace"
      pattern: '"(.*?):(.*)'
      captureIndex: 1
```

or old way : 

```
transformers:
  - findSubMatch: '"(.*?):(.*)'
```

which is equivalent of : 

```
transformers:
  - findSubMatch: 
      pattern: '"(.*?):(.*)'
      captureIndex: 0
```

### DIFF
--- a/pkg/core/transformer/main.go
+++ b/pkg/core/transformer/main.go
@@ -119,7 +119,7 @@ func (t *Transformer) Apply(input string) (output string, err error) {
 				return "", fmt.Errorf("no regex provided")
 			}
 
-			// Trying to compile the RegExp
+			// Check if the regular expression can be compiled
 			re, err := regexp.Compile(f.Pattern)
 			if err != nil {
 				return "", err
--- a/pkg/core/transformer/main.go
+++ b/pkg/core/transformer/main.go
@@ -2,6 +2,7 @@ package transformer
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


olblak:
That's awesome, can I also ask to open a documentation pull request on https://github.com/updatecli/website/blob/master/content/docs/core/transformer.adoc?

pkg/core/transformer/main.go
```diff
@@ -100,31 +100,50 @@ func (t *Transformer) Apply(input string) (output string, err error) {
...
-			val, ok := value.(string)
-
-			if !ok {
-				return "", fmt.Errorf("unknown value for findSubmatch: %v", val)
+			f := FindSubMatch{}
+
+			// Here I deal with 2 ways of providing the data within the yaml
+			// only one parameter in string which is the pattern
+			// considering 0 as the CaptureIndex default
+			// or receiving both parameter as a map.
```
lemeurherve, line 108:
```suggestion
			// If the manifest value is only a string (the pattern), then 0 is the implied captureIndex value
			// Otherwise, both pattern and captureIndex are retrieved from the map value of the manifest
```

pkg/core/transformer/main.go
```diff
@@ -100,31 +100,50 @@ func (t *Transformer) Apply(input string) (output string, err error) {
...
 
+			// I check if nothing found
 			if len(found) == 0 {
-				logrus.Debugf("No result found after applying regex %q to %q", val, output)
+				logrus.Debugf("No result found after applying regex %q to %q", f.Pattern, output)
 				return "", nil
 			}
 
-			fmt.Println(found)
+			// I check if I got a capture with that index
```
lemeurherve, line 139:
```suggestion
			// Log if there can't be a submatch corresponding to the captureIndex
```

pkg/core/transformer/main_test.go
```diff
@@ -186,6 +186,32 @@ var (
...
+				Transformer{
+					"findSubMatch": FindSubMatch{
+						Pattern:      `\d*.\d*`,
+						CaptureIndex: 1,
+					},
+				},
+			},
+			expectedOutput: "",
+			expectedErr:    nil,
+		},
```
lemeurherve, line 214: I would also add a test with a too big CaptureIndex, and maybe another one with multiple submatches/capture groups (ex: `\d*.(\d*).(\d*)` (not sure if there is a use case for this one)

pkg/core/transformer/main.go
```diff
@@ -100,31 +100,50 @@ func (t *Transformer) Apply(input string) (output string, err error) {
...
+			if len(found) <= f.CaptureIndex {
+				logrus.Debugf("No capture found at position %v after applying regex %q to %q, full result with CaptureIndex 0 would be %v", f.CaptureIndex, f.Pattern, output, found)
+				return "", nil
+			}
 
-			output = found[0]
+			// here is the returned string
+			output = found[f.CaptureIndex]
```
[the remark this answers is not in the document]
smerle33: the check is done just above : 
if len(found) <= f.CaptureIndex {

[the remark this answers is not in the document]
lemeurherve: Erf, my bad

pkg/core/transformer/main.go
```diff
@@ -100,31 +100,48 @@ func (t *Transformer) Apply(input string) (output string, err error) {
...
 
-			fmt.Println(found)
+			// I check if I got a capture with that index
+			if len(found) <= f.CaptureIndex {
+				logrus.Debugf("No capture found at position %v after applying regex %q to %q, full result with CaptureIndex 0 would be %v", f.CaptureIndex, f.Pattern, output, found)
+				return "", nil
+			}
 
-			output = found[0]
+			// here is the returned string
```
lemeurherve, line 143:
```suggestion
			// Output the submatch corresponding to the captureIndex
```

pkg/core/transformer/main_test.go
smerle33: the choice is to avoid any error (as per discussion with damien about this behavior when populating the issue), so I added a test but with no error match .... (it's a logrus log)

lemeurherve (approved):

olblak (approved):

PR #556, opened 2022-03-07
Replace example with a video

olblak:
Update README with a video


lemeurherve:
I'd keep the text example in addition to the video as some people doesn't want to watch a video to see an example.

lemeurherve:
How about specifying the architecture to avoid the log `WARNING: No architecture specified for the image "registry-1.docker.io/jenkins/jenkins:". Using the default "amd64".`?

olblak:
@lemeurherve Thanks for you feedback, I'll implement them when I have some time

lemeurherve:
I would put the example in a separate file like `quickstart-example.adoc`, [include it](https://docs.asciidoctor.org/asciidoc/latest/directives/include/) in this README, and add a manifest to update this file every time [the corresponding file`*` in updatecli/website](https://github.com/updatecli/website/blob/master/content/docs/prologue/quick-start.adoc) changes so these docs could stay in sync.

`*` this might need to extract the "body" in a separate file, letting the header/title be out of it so it's easier to include in the file in this repo.

dduportal:
Moving back to draft

olblak:
I added back the example section

PR #557, opened 2022-03-10
Fix for Authentication Bug with Private Git Repos

hi-artem:
# Fix for Authentication Bug with Private Git Repos

As it is right now, authentication with Private Git Repo is not working properly. It can be replicated by trying to make a `pullrequest` to a private repository.

The following PR adds `Auth` property to all the methods supporting `AuthOptions` in `Checkout` method.

## Test

```shell
plugins/utils/gitgeneric on  feature/auth-checkout via 🐹 v1.17.5 on ☁️  (us-west-2)
❯ go test

PASS
ok  	github.com/updatecli/updatecli/pkg/plugins/utils/gitgeneric	4.447s
```

### Tradeoff

n/a

### DIFF
--- a/pkg/plugins/scms/git/scm.go
+++ b/pkg/plugins/scms/git/scm.go
@@ -19,7 +19,7 @@ func (g *Git) Add(files []string) error {
 
 // Checkout create and then uses a temporary git branch.
 func (g *Git) Checkout() error {
-	err := git.Checkout(g.Branch, g.remoteBranch, g.GetDirectory())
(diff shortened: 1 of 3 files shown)


olblak:
Thanks @hi-artem for the contribution. It looks very good to me. Let me test it before merging

olblak:
While testing the pullrequest, I spot https://github.com/updatecli/updatecli/issues/558 
I need to open a pullrequest with a fix (which I have locally)

Also to test the pullrequest, I wrote a updatecli.yaml which I now need to commit in the directory e2e so we catch the current regression in the futur

olblak (approved):

PR #565, opened 2022-03-10
Improve reporting of pipeline with wrong configuration

### LINKED ISSUE #552: Validating "FILE" is displayed too often

### Current Behavior

SOURCES
=======
Validating "UPDATECLI.YAML"

githubSemver
------------
Searching for version matching pattern "~0.1"
✔ Git tag "0.1.2" found matching pattern "~0.1"
Validating "UPDATECLI.YAML"

gitExample2
-----------
ERROR: ✗ no tag found
Validating "UPDATECLI.YAML"

gitExample1
-----------
Searching for version matching pattern "latest"
✔ Git tag "v0.20.1" found matching pattern "latest"
Validating "UPDATECLI.YAML"

githubRegex
-----------
Searching for version matching pattern "v0.*"
✔ Git tag "v0.20.1" found matching pattern "v0.*"

CONDITIONS:
===========
Validating "UPDATECLI.YAML"

### Expected Behavior

In the current state, VALIDATING is displayed each time we re-evaluate a template which is done after each source
I would like to only show VALIDATING once at the beginning of a run

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Linux
- updatecli: v0.20.1
```

### Anything else?

_No response_

olblak:
# Improve reporting of pipeline with wrong configuration

Fix #552 
This pull request allows to better identify which pipeline is failing due to a bad configuration. 
Now It would show

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/golang.yaml"
Loading Pipeline "updatecli/updatecli.d/updatecli.yaml"
Loading Pipeline "updatecli/updatecli.d/venom.yaml"
ERROR: failed loading pipeline(s)
	* "updatecli/updatecli.d/golang.yaml" - scms ID "default2" referenced by pullrequest ID "default", doesn't exist
	* "updatecli/updatecli.d/updatecli.yaml" - scm id "default222" doesn't exist

```
instead of the old 
```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/golang.yaml"
Validating ""
Validating "GOLANG.YAML"
ERROR: ✗ scms ID "default2" referenced by pullrequest ID "default", doesn't exist
ERROR: Continuing and trying to go as far as possible

```
## Test

To test this pull request, we needs updatecli configuration with wrong content and good content.

```shell
cp <to_package_directory>
go test
```

This pullrequest introduces a behavior change during the "prepare" stage.
Previously , it would just try to load as much configuration as possible and clone as much git repository as possible without failing. But now we exit if we couldn't load any pipeline or if something wrong happened with one of the git clones.

### Potential improvement

Implement the same behavior for git clones where it tries to clones as much repository as possible

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -131,8 +131,6 @@ func (config *Config) Display() error {
 // Validate run various validation test on the configuration and update fields if necessary
 func (config *Config) Validate() error {
 
-	logrus.Infof("Validating %q", config.Name)
-
 	if config.Source.Kind != "" && (len(config.Sources) > 0) {
 		logrus.Errorf("Source and Sources can't be defined at the same time, please use the 'Sources' syntax")
 		return ErrBadConfig
--- a/pkg/core/engine/main.go
+++ b/pkg/core/engine/main.go
@@ -154,9 +154,12 @@ func (e *Engine) LoadConfigurations() error {
 
 		loadedConfiguration, err := config.New(cfgFile, e.Options.ValuesFiles, e.Options.SecretsFiles)
 
-		if err == config.ErrConfigFileTypeNotSupported {
+		switch {
+		case err == config.ErrConfigFileTypeNotSupported:
+			// Updatecli accepts either a single configuration file or a directory containing multiple configurations.
+			// When browsing files from a directory, we don't want to record error due to unsupported files.
 			continue
-		} else if err != nil && err != config.ErrConfigFileTypeNotSupported {
+		case err != nil && err != config.ErrConfigFileTypeNotSupported:
 			err = fmt.Errorf("%q - %s", cfgFile, err)
 			errs = append(errs, err)
 			continue
--- a/pkg/core/engine/main.go
+++ b/pkg/core/engine/main.go
@@ -178,7 +178,6 @@ func (e *Engine) LoadConfigurations() error {
 			err := fmt.Errorf("%q - %s", cfgFile, err)
 			errs = append(errs, err)
 		}
-
 	}
 
 	if len(errs) > 0 {
--- a/pkg/core/engine/main.go
+++ b/pkg/core/engine/main.go
@@ -1,6 +1,7 @@
 package engine
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"sync"
@@ -72,7 +73,7 @@ func (e *Engine) InitSCM() (err error) {
 			}
 		}
 	}
-	logrus.Infof("Repository retrieved: %d\n", len(hashes))
+	logrus.Infof("\nSCM repository retrieved: %d", len(hashes))
 
 	return err
 }
@@ -126,14 +127,20 @@ func (e *Engine) Prepare() (err error) {
 	}
 
 	err = e.LoadConfigurations()
-	if err != nil {
-		return err
+
+	if len(e.Pipelines) == 0 {
+		logrus.Errorln(err)
+		return fmt.Errorf("no valid pipeline found")
 	}
 
-	err = e.InitSCM()
+	// Don't exit if we identify at least one valid pipeline configuration
 	if err != nil {
-		return err
+		logrus.Errorln(err)
+		logrus.Infof("\n%d pipeline(s) successfully loaded\n", len(e.Pipelines))
 	}
+
+	// If one git clone fails then we exit
+	err = e.InitSCM()
 
 	return err
 }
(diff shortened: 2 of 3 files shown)


pkg/core/engine/main.go
```diff
@@ -126,43 +127,65 @@ func (e *Engine) Prepare() (err error) {
...
-		if err != nil && err != config.ErrConfigFileTypeNotSupported {
-			logrus.Errorf("%s\n\n", err)
+		if err == config.ErrConfigFileTypeNotSupported {
 			continue
-		} else if err == config.ErrConfigFileTypeNotSupported {
+		} else if err != nil && err != config.ErrConfigFileTypeNotSupported {
+			err = fmt.Errorf("%q - %s", cfgFile, err)
+			errs = append(errs, err)
 			continue
 		}
```
dduportal, line 163: this if/else struct starts to be hard to read. Do you mind using a switch case there?

olblak: Sure, I definitely can

pkg/core/config/main.go
```diff
@@ -131,8 +131,6 @@ func (config *Config) Display() error {
 // Validate run various validation test on the configuration and update fields if necessary
 func (config *Config) Validate() error {
 
-	logrus.Infof("Validating %q", config.Name)
```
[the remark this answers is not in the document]
lemeurherve: https://github.com/updatecli/updatecli/issues/536

pkg/core/engine/main.go
lemeurherve: Can't it be just to "if" without "else" here?

One "if err == config.ErrConfigFileTypeNotSupported {"
Then one "if err != nil && err != config.ErrConfigFileTypeNotSupported {"

pkg/core/engine/main.go
```diff
@@ -126,43 +127,65 @@ func (e *Engine) Prepare() (err error) {
...
+			err := fmt.Errorf("%q - %s", cfgFile, err)
+			errs = append(errs, err)
 		}
 
-		e.Pipelines = append(e.Pipelines, newPipeline)
-		e.configurations = append(e.configurations, loadedConfiguration)
+	}
+
+	if len(errs) > 0 {
+
```
lemeurherve, line 182:
```suggestion
```

pkg/core/engine/main.go
olblak: I would also prefer switch case as it makes things more readable

olblak: I think it's nitpicking but I took the opportunity to a comment

pkg/core/engine/main.go
```diff
@@ -126,43 +127,67 @@ func (e *Engine) Prepare() (err error) {
...
 	for _, cfgFile := range GetFiles(e.Options.File) {
 
 		loadedConfiguration, err := config.New(cfgFile, e.Options.ValuesFiles, e.Options.SecretsFiles)
 
-		if err != nil && err != config.ErrConfigFileTypeNotSupported {
-			logrus.Errorf("%s\n\n", err)
+		switch {
+		case err == config.ErrConfigFileTypeNotSupported:
+			// Updatecli accepts either a single configuration file or a directory containing multiple configurations.
+			// When browsing files from a directory, we don't want to record error due to unsupported files.
```
dduportal, line 160:
```suggestion
		switch err.(type) {
		case config.ErrConfigFileTypeNotSupported:
			// Updatecli accepts either a single configuration file or a directory containing multiple configurations.
			// When browsing files from a directory, we don't want to record error due to unsupported files.
```

pkg/core/engine/main.go
```diff
@@ -126,43 +127,67 @@ func (e *Engine) Prepare() (err error) {
...
 			continue
-		} else if err == config.ErrConfigFileTypeNotSupported {
+		case err != nil && err != config.ErrConfigFileTypeNotSupported:
```
dduportal, line 162:
```suggestion
		case config.ErrConfigFileTypeNotSupported:
```

dduportal (dismissed):
Fine for me, but proposed improvement around the switch by asserting on the error's type.

It's not blocking but as soon as there is an `else` it's a smell: thanks for commenting as it helps to maintain this code (unless there is a unit test to help other maintainers than the one whom wrote the initial conditions).

pkg/core/engine/main.go
lemeurherve: I didn't realized the "nil" case you were speaking about, it's clear now, thx!

dduportal: Yeah, it was there for a (really) good reason. I personnaly like the switch with type assertion: easier to read.

pkg/core/engine/main.go
olblak: This suggestion won't work as `config.ErrConfigFileTypeNotSupported` is not a type

pkg/core/engine/main.go
olblak: @dduportal I implemented your suggestion

dduportal (approved):

PR #590, opened 2022-03-20
 cleanup(gittag, githubrelease) remove deprecated code for field `version`

dduportal:
This PR removes the deprecated code of the field `version`.

It also introduces the following changes:

- Minor change on the `Makefile` to allow executing the end to end test with a custom binary directory (e.g. allow `make test-e2e VENOM_VAR_binpath=./custom-dist/` to use the binary `updatecli` located in `./custom-dist/updatecli`)
- Decouple the SCM handlers from the concept of source. It removes th "source" argument from the method `Init()` (interface ScmHandler) as it was only used to define versionfilters.

Associated documentation PR: https://github.com/updatecli/website/pull/264

## Test

```shell
make test
```

### DIFF
--- a/pkg/plugins/resources/githubrelease/main.go
+++ b/pkg/plugins/resources/githubrelease/main.go
@@ -1,46 +1,56 @@
 package githubrelease
 
 import (
-	"path"
-
 	"github.com/mitchellh/mapstructure"
-	"github.com/updatecli/updatecli/pkg/core/tmp"
 	"github.com/updatecli/updatecli/pkg/plugins/scms/github"
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
+// Spec defines a specification for a "gittag" resource
+// parsed from an updatecli manifest file
+type Spec struct {
+	Owner         string         // Owner specifies repository owner
+	Repository    string         // Repository specifies the name of a repository for a specific owner
+	Token         string         // Token specifies the credential used to authenticate with
+	URL           string         // URL specifies the default github url in case of GitHub enterprise
+	Username      string         // Username specifies the username used to authenticate with Github API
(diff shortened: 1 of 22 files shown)


lemeurherve (dismissed):

lemeurherve (approved):

dduportal:
- I'm waiting for review from 2 maintainers on this one (to be really sure that I did not do something wrong).
- Proposed milestone: 0.22.3

pkg/plugins/resources/githubrelease/main.go
```diff
@@ -1,46 +1,56 @@
...
 	"github.com/mitchellh/mapstructure"
-	"github.com/updatecli/updatecli/pkg/core/tmp"
 	"github.com/updatecli/updatecli/pkg/plugins/scms/github"
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
+// Spec defines a specification for a "gittag" resource
+// parsed from an updatecli manifest file
+type Spec struct {
+	Owner         string         // Owner specifies repository owner
```
olblak, line 12: Not blocker for this PR, but as recently discussed, It would be good to put comment before a struct field and not on the same line.

dduportal: Yep, that would be a separated (and strictly scoped) PR to avoid messing up our rebases ;) But thanks to remind it to us 👍

olblak (approved):
Thanks for this huge cleanup, especially considering that I failed to correctly document the code which evolved many times

dduportal:
> Thanks for this huge cleanup, especially considering that I failed to correctly document the code which evolved many times

Thanks, but don't be too harsh with yourself :) It's a looot of work, so let's improve this tool and its doc one step after the other

PR #587, opened 2022-03-15
Add updatecli subcommand to generate json schema

olblak:
# Add updatecli subcommand to generate json schema

- [x] Require https://github.com/updatecli/updatecli/issues/444

PR related to https://github.com/updatecli/updatecli/pull/540 but the current one focus on generating json schema while #540 test updatecli manifest with json schema

Fix #524 

The current pullrequest introduces several requirements

1. To retrieve field description, a comment must be one line maximum and me precede the field such
```
// Source defines parameters to retrieve an information from a specific location
type Source struct {
    // Title defines a source title
    title string // This is not a valid comment anymore
} 
``` 
2. Configuration key must be lowercase to be valid, while we could use another convention, we can't mix them. 
```
--- 
    sources:
        default: 
            scmID: # not valid
            scmid: # valid  
```

## Test

To test this pullrequest, you can: 

1. Ensure that test in package `github.com/pkg/core/jsonschema` are passing.
```shell
make test
```

2. You can try to generate an Updatecli jsonschema using the new jsonschema argument 
```shell
go build -o bin/updatecli && ./bin/updatecli jsonschema --directory schema
```

3. Finally, you can configure your IDE such as vscode to match files like ["**/updatecli.di/**/*.yaml","**/updatecli.di/**/*.yml","**/updatecli.di/**/*.tpl"] with the jsonschema file that you generated locally such as "$HOME/Projects/Updatecli/updatecli/schema/config.json" as explained on [vscode documentation](https://code.visualstudio.com/Docs/languages/json#_file-match-syntax)  

### Tradeoff

/

### Potential improvement

This pullrequest won't be usable as is as we need additional cleanup.
.1 Ensure that every field expected lowercase value. for example this [tag](https://github.com/updatecli/updatecli/blob/bb4c8c94120bccab45a20549382721f51a129b25/pkg/core/pipeline/resource/main.go#L31) must be removed

PR #615, opened 2022-04-05
Add various spec comment and fix small linting issues

olblak:
# Add various spec comment and fix small linting issues

This pullrequest is adding some comment for spec structure so it can be used by https://github.com/updatecli/updatecli/pull/587
While I am in the plugin, I also take the opportunity to fix small linting issues

## Test

/

### Tradeoff

I am only focusing on spec as it's need for generating jsonschema with description

### Potential improvement

Still some plugin left


dduportal (dismissed):
🚀

dduportal (approved):

PR #621, opened 2022-04-07
Avoid directory when fetching updatecli comments

olblak:

# Avoid directory when fetching updatecli comments

For the command `updatecli jsonschema` to populate the field "description", we cloned the updatecli directory in a temporary location and then we delete the directory once we are done. For some stupid reason I was using "/tmp/updatecli" which is also used by default by the scm resource to manipulate git repository. So I ended up in a situation where I deleted `/tmp/updatecli` 
 before any git operation to happen.

## Test

To test this pull request, I used this configuration https://github.com/updatecli/website/blob/master/updatecli/updatecli.d/jsonschema.yaml

### Potential improvement

In the future, we have a risk of collusion if at some point we have a git repository "github.com/updatecli/_comments"

### DIFF
--- a/pkg/core/jsonschema/main.go
+++ b/pkg/core/jsonschema/main.go
@@ -27,7 +27,7 @@ var (
 	updatecliPackageName string = "github.com/updatecli/updatecli/"
 
 	// commentDir defines the temporary directory where
-	commentDir string = path.Join(os.TempDir(), "updatecli/")
+	commentDir string = path.Join(os.TempDir(), "updatecli", "_comments")
 	// commentURL defines the updatecli git url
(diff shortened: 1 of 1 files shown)


pkg/core/jsonschema/main.go
```diff
@@ -27,7 +27,7 @@ var (
 	updatecliPackageName string = "github.com/updatecli/updatecli/"
 
 	// commentDir defines the temporary directory where
```
lemeurherve, line 29: where what? ^^

olblak: Thanks for catching this, I guess I delete this line somehow

lemeurherve (approved):

dduportal (approved):

Issue #632, opened 2022-04-12
[updatecli 0.23.x] failing to open pull request + no error reported globally


dduportal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With updatecli 0.23.0 and 0.23.1, we have errors when trying to open a pull request, with the error `ERROR: 1 target(s) (updateChartVersion) failed for pullrequest "default"`.

Besides this failure, the execution of updatecli does NOT fail and does NOT report the failure, which makes it hard to detect.

### Expected Behavior

Can't be sure of the root cause error, but we have to fix updatecli so that, at least, it reports the failure, and eventually give details on the failure.

### Steps To Reproduce

* Any manifest at https://github.com/jenkins-infra/kubernetes-management/tree/56ff776b83f8677e16f8e8c9d422c7d3a7205a1f/updatecli/updatecli.d

  * In particular (a simple one):

```yaml
title: Bump aws-node-termination-handler helm chart version

scms:
  default:
    kind: github
    spec:
      user: "{{ .github.user }}"
      email: "{{ .github.email }}"
      owner: "{{ .github.owner }}"
      repository: "{{ .github.repository }}"
      token: "{{ requiredEnv .github.token }}"
      username: "{{ .github.username }}"
      branch: "{{ .github.branch }}"

sources:
  lastChartVersion:
    kind: helmChart
    spec:
      url: https://aws.github.io/eks-charts
      name: aws-node-termination-handler

targets:
  updateChartVersion:
    name: "eks/aws-node-termination-handler Helm Chart"
    kind: file
    scmid: default
    spec:
      file: clusters/cik8s.yaml
      matchpattern: 'chart: eks\/aws-node-termination-handler((\r\n|\r|\n)(\s+))version: .*'
      replacepattern: 'chart: eks/aws-node-termination-handler${1}version: {{ source "lastChartVersion" }}'

pullrequests:
  default:
    kind: github
    scmid: default
    targets:
      - updateChartVersion
    spec:
      labels:
        - dependencies
        - aws-node-termination
```

* Executing the diff works as expected and succeeds to report expected change + PR:

```shell
updatecli diff --values updatecli/values.yaml --config updatecli/updatecli.d/charts/aws-node-termination-handler.yaml


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/charts/aws-node-termination-handler.yaml"

SCM repository retrieved: 1


++++++++++++
[...]

PR #633, opened 2022-04-12
fix: return list of modified files for targets of kind: file

lemeurherve:
Ref: https://github.com/updatecli/updatecli/issues/632

# Title

Part of #632 

## Test

```shell
cp <to_package_directory>
go test
```


lemeurherve:
@dduportal this one is good for review

It fails correctly when I comment the else ;)

dduportal (approved):
LGTM great work!

PR #634, opened 2022-04-12
fix: return an error when no file has been changed

lemeurherve:
# fix: return an error when no file has been changed

Part of #632

Follow-up of/includes #633

## Test

```shell
cp <to_package_directory>
go test
```


lemeurherve:
Closed in favor of #636

PR #636, opened 2022-04-12
fix: return an error when no file has been changed

lemeurherve:
# fix: return an error when no file has been changed

Part of https://github.com/updatecli/updatecli/issues/632

Supersede #634

## Test

```shell
cp <to_package_directory>
go test
```


dduportal (approved):
LGTM 👍

Issue #617, opened 2022-04-06
Empty sources are displayed while it should not


olblak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Source is mentioned even though the updatecli manifest doesn't have some

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/schema.yaml"

SCM repository retrieved: 0


++++++++++++
+ PIPELINE +
++++++++++++



#####################
# UPDATE JSONSCHEMA #
#####################


SOURCES
=======


TARGETS
========

jsonschema
----------

**Dry Run enabled**

The shell 🐚 command "/home/olblak/Projects/Updatecli/updatecli/bin/updatecli jsonschema --baseID https://www.updatecli.io/latest/schema --directory ./schema" ran successfully with the following output:
----
----
No change detected

=============================

REPORTS:


✔ SCHEMA.YAML:
        Sources:
        Target:
                ✔ [jsonschema] Bump jsonschema (kind: shell)



Run Summary
===========
Pipeline(s) run:
  * Changed:    0
  * Failed:     0
  * Skipped:    0
  * Succeeded:  1
  * Total:      1
 ```

### Expected Behavior

Sources should not be mentioned if they are no source definition

```
#####################
# UPDATE JSONSCHEMA #
#####################

TARGETS
========

jsonschema
----------

**Dry Run enabled**

The shell 🐚 command "/home/olblak/Projects/Updatecli/updatecli/bin/updatecli jsonschema --baseID https://www.updatecli.io/latest/schema --directory ./schema" ran successfully with the following output:
----
----
No change detected

=============================

REPORTS:


✔ SCHEMA.YAML:
        Target:
                ✔ [jsonschema] Bump jsonschema (kind: shell)



Run Summary
===========
Pipeline(s) run:
  * Changed:    0
  * Failed:     0
  * Skipped:    0
  * Succeeded:  1
  * Total:      1
 ```

### Steps To Reproduce

Execute an updatecli manifest without source

### Environment

```markdown
- OS:linux
- updatecli:v0.22.3
```

PR #640, opened 2022-04-12
fix: no source/condition/target = no logs

lemeurherve:
# fix: no source/condition/target = no logs

Fixes https://github.com/updatecli/updatecli/issues/617

## Test

```shell
cp <to_package_directory>
go test
```

### DIFF
--- a/pkg/core/pipeline/conditions.go
+++ b/pkg/core/pipeline/conditions.go
@@ -10,8 +10,10 @@ import (
 // RunConditions run every conditions for a given configuration config.
 func (p *Pipeline) RunConditions() (globalResult bool, err error) {
 
-	logrus.Infof("\n\n%s:\n", strings.ToTitle("conditions"))
-	logrus.Infof("%s\n", strings.Repeat("=", len("conditions")+1))
+	if len(p.Conditions) > 0 {
+		logrus.Infof("\n\n%s:\n", strings.ToTitle("conditions"))
+		logrus.Infof("%s\n", strings.Repeat("=", len("conditions")+1))
+	}
 
 	// Sort conditions keys by building a dependency graph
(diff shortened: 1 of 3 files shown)


pkg/core/pipeline/conditions.go
```diff
@@ -10,8 +10,10 @@ import (
 // RunConditions run every conditions for a given configuration config.
 func (p *Pipeline) RunConditions() (globalResult bool, err error) {
 
-	logrus.Infof("\n\n%s:\n", strings.ToTitle("conditions"))
-	logrus.Infof("%s\n", strings.Repeat("=", len("conditions")+1))
+	if len(p.Conditions) > 0 {
```
olblak, line 13: Different approach but what do you think to exit early if nothing need to be done like 

```
if len(p.Conditions) == 0 {
    // Return early as no condition to run
    logrus.Debugln("no condition defined")
    return nil
}
```

lemeurherve: Oooh I like it! Let me take a look

lemeurherve: @olblak `logrus.Debugln` or `logrus.Infof` so the user knows why there isn't any output for the stage?

lemeurherve: I went with `logrus.Debugln` as you suggested.

olblak: I think debug is better as it doesn't make sense to me tell the user that he didn't specify a condition or a source (or a target)

olblak (approved):
Thanks for this PR

dduportal:
Do you have examples of runs for the 3 cases? Not sure of how it behaves?

PR #641, opened 2022-04-12
feat: sort stages if they don't have any 'depends_on'

### LINKED ISSUE #468: Set an order on sources execution queue

As it can be seen when running for example [./examples/updatecli.generic/files.yaml](https://github.com/updatecli/updatecli/blob/cf48e7b659e3faf9970f49225dc6489107297bff/examples/updatecli.generic/file.yaml) there isn't any apparent order of the execution of the different sources ("ContentFromURL" coming before "ContentFromLocalFile" while it's the second one in the manifest).

I can be confusing for newcomers, and inconvenient at least for others (I remember having ran several tests looking for a particular result at the end of the outputs before noticing it was in the middle)

### Solution you'd like

Sorting execution by a field like the sources id, alphabetically

Do nothing about it.

### Anything else?

_No response_

lemeurherve:
# feat: sort stages if they don't have any 'depends_on'

Fixes #468

For each type of stages (Sources, Conditions, Targets), execute them in alphabetical order if none of them depends on another, in which case we have to keep their descendant order.

## Test

```shell
cp <to_package_directory>
go test
```

### Tradeoff

We can't easily (at all?) sort stages when there is any dependency between them.

### DIFF
--- a/pkg/core/pipeline/sort.go
+++ b/pkg/core/pipeline/sort.go
@@ -2,6 +2,7 @@ package pipeline
 
 import (
 	"errors"
+	"sort"
 	"strings"
 
 	"github.com/heimdalr/dag"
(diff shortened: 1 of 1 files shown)


olblak:
I think, I should have asked this earlier but I am not sure to understand what problem you are trying to solved?
We already have the `depends_on` when we need to order the different stages.

Sorting the different stages alphabetically doesn't seem natural to me. Maybe using [orderedmap](https://github.com/iancoleman/orderedmap) would be better. That being said

Link: 
* [Why Golang maps are unordered](https://stackoverflow.com/questions/55925822/why-are-iterations-over-maps-random/55925880#55925880)

lemeurherve:
When debugging simple manifest with lots of stages it's easier (I think) to be able to follow outputs when they arrive in a predefined order than seemingly random.

(And yes I know map aren't ordered ;))

It's not a mandatory feature, but I also don't see the harm in adding it.

olblak:
After additional thoughts, I think that I am fine with this feature but I would also be interested by @dduportal input

dduportal:
I don't see any harm on this change so why not? At least it brings some determinism in the processing order.

The actual usage of map structure is not best suited. Did not realize back then, but after 1 year of usage, my conclusion is that a list would be more suited, except for referencing between templates

PR #645, opened 2022-04-13
fix: use the URL instead of the token to initialize GitHub Enterprise client

lemeurherve:
# fix: use the URL instead of the token to initialize GitHub Enterprise client

Fix #XXX

## Test

```shell
cp <to_package_directory>
go test
```

## Additional Information

Saw this while working on #644, quick fix as it is.


olblak (approved):
Oops

PR #652, opened 2022-04-17
Bump golanci-lint version used in Github Action

### LINKED ISSUE #651: Add updatecli manifest to automate golangci-lint version in Github Action
Golangci used in Github Action is outdated

It appears that we are still using version 1.29 https://github.com/updatecli/updatecli/blob/bc3500a92351cfc617eb198b7c00872ee763e3ba/.github/workflows/go.yaml#L49 

https://github.com/golangci/golangci-lint

olblak:
Fix #651 

* Bump golangci-lint version
* Add updatecli manifest to automate version update

## Test

```shell
updatecli diff --config updatecli/updatecli.d/golangci-lint.yaml
```

### Tradeoff
/

### Potential improvement

/

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -46,7 +46,7 @@ jobs:
           # Required: the version of golangci-lint is required
           # and must be specified without patch version:
           # we always use the latest patch version.
-          version: v1.29
+          version: v1.45
       - name: Build
         run: make build
       - name: Quick Test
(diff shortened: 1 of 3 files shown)


olblak:
After some investigation `gorelease` version 1.8 change the path for the binary generated from ``dist/updatecli_linux_amd64/updatecli` to `dist/updatecli_linux_amd64_v1/updatecli` so the venom don't work anymore as they don't have the correct binary

PR #665, opened 2022-04-25
Add transformers field to jsonschema

### LINKED ISSUE #653: Jsonschema doesn't include transformers spec

### Current Behavior

Vscode doesn't understand `transformers`

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- updatecli:
```

### Anything else?

_No response_

### LINKED ISSUE #655: Feature Request: Transformer key should all be lowercase

To fix #653 we would need  all transformers key to be working when lowercase. 

### Solution you'd like

This change must be backward compatible so both `addSuffix` and `addsuffix` works (and any other transformer key).

_No response_

### Anything else?

https://github.com/updatecli/updatecli/blob/main/pkg/core/transformer/main.go

olblak:
Fix #655 
Fix #653

* Convert transformer to a struct which allows to generate jsonschema fields
* Deprecate non lowercase transformer fields such as `semverInc` becomes `semverinc`

## Test

```shell
cp pkg/core/transformer
go test
 go build -o bin/updatecli && ./bin/updatecli jsonschema --directory schema
```

Ensure that jsonschema contains transformers field

### Tradeoff

A side effect of converting transformer to a stuct means that the two following examples are accepted and means different things

*Order is guaranteed because of the list 
```
# v1.0.0-alpha
transformers:
    - trimsuffix: alpha
    - replace:
          from: alpha
          to: beta
``` 

Order is not guaranteed because of the map
```
# v1.0.0-alpha
transformers:
    - trimsuffix: alpha
      replace:
          from: alpha
          to: beta
```


olblak:
Considering that this pullrequest is fully backward compatible, I am going to merge it and proceed.
The three main advantages of this pulllrequest

1. Allow to use lowercase transform key so `addprefix` instead of `addPrefix` which is more consistent with the naming convention used now
2. Generate jsonschema for transformer which is more convenient when writing Updatecli manifest from IDE support jsonschema store

olblak:
for context, I move the updatecli manifest in the e2e directory under e2e/updatecli.d so Vscode could enable jsonschema validation

olblak:
After adding e2e tests for transformers where I test both the old and new syntax. I consider this ready to be merged

PR #671, opened 2022-05-01
Add subcommand to upgrade Updatecli manifest 

### LINKED ISSUE #618: Feature Request: Ability to update updatecli manifest from updatecli command line

We regularly introduce manifest configuration changes. While we take care of not introducing breaking changes, at some point we need to review every configuration and update them accordingly warning messages.
It would be awesome to have a command that load every manifest and update them to the latest version.

Current behavior
```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/docker-images/404.yaml"
WARNING: "scmID" is deprecated in favor of "scmid".
WARNING: "scmID" is deprecated in favor of "scmid".
WARNING: "scmID" is deprecated in favor of "scmid".
WARNING: "scmID" is deprecated in favor of "scmid".
WARNING: "scmID" is deprecated in favor of "scmid".
```

### Solution you'd like

I am looking for a subcommand to read manifest and fix warning automatically using 

```
updatecli manifest upgrade
```

We'll need to store the updatecli version compatible in manifest so we always move configuration forward.

_No response_

### Anything else?

Having the version in the manifest that we can also trigger warning when a very old manifest syntax is used with a more recent updatecli version

olblak:
# Allow to update Updatecli manifest between Updatecli version

Fix #618 

This pullrequest is an attempt to allow upgrading updatecli manifest to the latest version.
This required the following changes

- [x] Refactor the package `config`
- [x] Allow to disable configuration templating
- [x] Allow to write back a manifest to disk 
- [x] Ensure we don't downgrade manifest version
- [x] Add version check between updatecli version and manifest version
## Test

```shell
go build -o bin/updatecli . && ./bin/updatecli manifest upgrade --config examples/updatecli.d
```

### Tradeoff

Because we load the configuration in memory, some fields are automatically added such as `pipelineId`, we need to identify those that should not be include

### Potential improvement

* Run `updatecli show` without templating
* Rename `updatecli show` to `updatecli manifest show` for more consistency with `updatecli manifeste upgrade`
* Clarify the different between config `title` and config `name`
* Due due to the config package refactoring, we should be able to easily add other data file such as json
* Deprecated extension `.tpl`


olblak:
I added the tag `yaml:,omitempty` to every spec fields otherwise when we upgrade a manifest, it automatically adds those fields with an empty value.

Instead of having 

```
targets:
    default:
        name: jenkins
        kind: jenkins
```

we get 
```
targets:
    default:
        disableSourceInput: fales
        name: jenkins
        kind: jenkins
        spec:
         ...
        transfomers:
        ...
```
if we don't omitempty when marshalling yaml

olblak:
Here is an update of what it looks like when running `updatecli manifest upgrade`
 
when no change detected
```
  ./bin/updatecli manifest upgrade --config updatecli/updatecli.d/venom.yaml --debug

++++++++++++++++++++
+ MANIFEST UPGRADE +
++++++++++++++++++++

Loading Pipeline "updatecli/updatecli.d/venom.yaml"
✔ No Updatecli manifest change required
```

When a change is detected
```
 ./bin/updatecli manifest upgrade --config updatecli/updatecli.d/venom.yaml  

++++++++++++++++++++
+ MANIFEST UPGRADE +
++++++++++++++++++++

Loading Pipeline "updatecli/updatecli.d/venom.yaml"
WARNING: "sourceID" is deprecated in favor of "sourceid".
```

wih debug enabled

```
 ./bin/updatecli manifest upgrade --config updatecli/updatecli.d/venom.yaml --debug  

++++++++++++++++++++
+ MANIFEST UPGRADE +
++++++++++++++++++++

Loading Pipeline "updatecli/updatecli.d/venom.yaml"
WARNING: "sourceID" is deprecated in favor of "sourceid".
DEBUG: ⚠ Updatecli manifest change required
--- updatecli/updatecli.d/venom.yaml(old)
+++ updatecli/updatecli.d/venom.yaml(new)
@@ -45,4 +45,4 @@
             file: .github/workflows/go.yaml
             matchpattern: 'VENOM_VERSION: .*'
         scmid: default
-        sourceid: latestVersion
+        sourceid: latestVersion
```

olblak:
I am wondering if having a "dryrun" would be usefull.

I am considering adding a flag like `--check` or `--dry-run` to `updatecli manifest upgrade`
or maybe to be more consistent with `updatecli diff && updatecli apply` by having something like `updatecli manifest upgrade diff` and `updatecli manifest upgrade apply`

dduportal:
>I added the tag yaml:,omitempty to every spec fields otherwise when we upgrade a manifest, it automatically adds those fields with an empty value.

Thanks for this! I was loosing my mind in #713 when trying to implement a merge method on the git objects, now I understand why. Tha

dduportal:
> I am wondering if having a "dryrun" would be usefull.
> 
> I am considering adding a flag like `--check` or `--dry-run` to `updatecli manifest upgrade` or maybe to be more consistent with `updatecli diff && updatecli apply` by having something like `updatecli manifest upgrade diff` and `updatecli manifest upgrade apply`

Proposal:

- `updatecli manifest check` that would return:
  - Exit code `2` if there is an upgrade to be run, with the diff on the stdout (no need for debug)
  - Exit code `0` if there is not, with nothing on the stdout except an informational message
- `updatecli maniferst upgrade` like you currently have
  - Exit code `0` if it upgraded the manifest successfully, as well if there isn't anything to do (idempotent operation)
  - If there is a change, then the diff is printed on  the stdout (no need for a debug flag)
  - Optionnaly, we can add a `--quiet` flag that would suppress the diff, but I'm not sure of its usefullness


The rationale is related to the fact that in both case, the "action" has a different intent and is applied on a manifest: using flags is confusing for end users, and I don't see a reason to not showing the diff (when there is one) by default.

Debug flag should only be for powerful users when they are debugging a weird (buggy?) behavior.

WDYT?

dduportal:
@olblak I tried to solve the conflicts caused by #717  (the `pipelineID` fixed typo + removal of pointer) but I failed to do so. May I let you rebase your branch from `upstream/master` and solve the conflict + push force to remove my change (and fix the build)?

olblak:
> > I am wondering if having a "dryrun" would be usefull.
> > I am considering adding a flag like `--check` or `--dry-run` to `updatecli manifest upgrade` or maybe to be more consistent with `updatecli diff && updatecli apply` by having something like `updatecli manifest upgrade diff` and `updatecli manifest upgrade apply`
> 
> Proposal:
> 
>     * `updatecli manifest check` that would return:
>       
>       * Exit code `2` if there is an upgrade to be run, with the diff on the stdout (no need for debug)
>       * Exit code `0` if there is not, with nothing on the stdout except an informational message
> 
>     * `updatecli maniferst upgrade` like you currently have
>       
>       * Exit code `0` if it upgraded the manifest successfully, as well if there isn't anything to do (idempotent operation)
>       * If there is a change, then the diff is printed on  the stdout (no need for a debug flag)
>       * Optionnaly, we can add a `--quiet` flag that would suppress the diff, but I'm not sure of its usefullness
> 
> 
> The rationale is related to the fact that in both case, the "action" has a different intent and is applied on a manifest: using flags is confusing for end users, and I don't see a reason to not showing the diff (when there is one) by default.
> 
> Debug flag should only be for powerful users when they are debugging a weird (buggy?) behavior.
> 
> WDYT?

Thanks for the feedback, that's a great idea, I'llo implement that

pkg/core/pipeline/condition/main.go
```diff
@@ -29,14 +29,13 @@ type Condition struct {
 
 // Config defines conditions input parameters
 type Config struct {
-	resource.ResourceConfig `yaml:",inline"`
+	resource.ResourceConfig `yaml:",inline,omitempty"`
```
lemeurherve, line 32: Why there isn't any type on this line like for ex https://github.com/updatecli/updatecli/pull/671/files#diff-552d39479cf11c87201c084db9cb6ead8c57697852e048e47839addfc64d9400L32-R32?

dduportal: This « form » is a kind of inheritance between structs. It allows the children struct to have the same fields as the parent (parent being the resource.Resource here).
I believe it is named « composition » in Go

dduportal: More references to help you:
- https://go.dev/ref/spec#Struct_types => golang reference
- https://go101.org/article/type-embedding.html

dduportal: Oh, I understand the confusion: this last example is not the best naming 😅

In that case, it is a named field which is expected to hold an object of type `transformer.Transformers`. 
It is not an inheritance of `transformer.Transformers`.

olblak:
After additional though, I find `updatecli manifest check` confusing with validating schema using jsonschema versus validating that manifest can be upgraded

dduportal:
> After additional though, I find `updatecli manifest check` confusing with validating schema using jsonschema versus validating that manifest can be upgraded

WDYT about 
`updatecli manifest validate` to .. validate with the schema ?

olblak:
Well I found the discussion interesting but a bit confusing, what do you think of

* `updatecli show manifest`: Display in the console an updatecli manifest. Similar to the current command `udpatecli show` (and we deprecated `updatecli show`)
* `updatecli validate manifest upgrade`: Test that a manifest can be upgrade 
* `updatecli validate manifest syntax`: Test using jsonschema that a manifest is validate 
* `updatecli upgrade manifest`: Try to upgrade an updatecli manifest to the latest version available based on updatecli version

olblak:
Btw I found the bug in the textdiff and I think I rebased correctly

olblak:
> > putting the verb first is confusing for me
> 
> I am really not attached to the order so we can go with your suggestion
> 
>     * `updatecli manifest show`: Display in the console an updatecli manifest. Similar to the current command `udpatecli show` (and we deprecated `updatecli show`)
> 
>     * `updatecli manifest validate`: Test using jsonschema that a manifest is valid
> 
>     * `updatecli manifest upgrade`: Try to upgrade an updatecli manifest to the latest version available based on updatecli version
> 
> 
> > updatecli manifest upgrade --write
> 
> I like very much your suggestion to display on stdout by default unless we specify a flag I prefer the short version `-i` with the long version `in-place` similarly to "sops"
> 
> > Go back to your initial proposal with the --check flag for updatecli manifest upgrade
> 
> What do you think of using the parameter `--dry-run` such as `updatecli manifest upgrade --dry-run`

I just realise that `--in-place` and `--dry-run` fill the same purpose. I am dropping `--dry-run`

olblak:
I think the PR is ready to be merged

The new command is ``updatecli manifest upgrade` 
```
Usage:
  updatecli manifest upgrade [flags]

Flags:
  -c, --config string   Sets config file or directory. (default: './updatecli.yaml') (default "./updatecli.yaml")
  -h, --help            help for upgrade
  -i, --in-place        Write updated Updatecli manifest back to the same file instead of stdout

Global Flags:
      --debug   Debug Output

```

The dry-run is the default behavior.

I did some testing with the example directory 

```
++++++++++++++++++++
+ MANIFEST UPGRADE +
++++++++++++++++++++

Loading Pipeline "examples/updatecli.d/nginx.yaml"
WARNING: The directive 'scm' for the target["jenkinsioNginxDigest"] is now deprecated. Please use the new top level scms syntax
⚠ Updatecli manifest change required
--- examples/updatecli.d/nginx.yaml(old)
+++ examples/updatecli.d/nginx.yaml(updated)
@@ -1,23 +1,36 @@
+name: NGINX.YAML
+pipelineid: 5d2ef3f73e3c99cb9d6470470fa613b71156b3a9b229e2b62feba3f4f1216559
+pullrequests:
+    target_jenkinsioNginxDigest:
+        kind: github
+        spec: {}
+        scmid: target_jenkinsioNginxDigest
+        targets:
+            - jenkinsioNginxDigest
+scms:
+    target_jenkinsioNginxDigest:
+        kind: github
+        spec:
+            branch: master
+            email: update-bot@olblak.com
+            owner: jenkins-infra
+            repository: charts
+            token: x
+            user: update-bot
+            username: olblak
 sources:
-  lastDockerDigest:
-    kind: dockerdigest
-    name: "Get latest nginx:1.17 from dockerhub"
-    spec:
-      image: "library/nginx"
-      tag: "1.17"
+    lastDockerDigest:
+        name: Get latest nginx:1.17 from dockerhub
+        kind: dockerdigest
+        spec:
+            image: library/nginx
+            tag: "1.17"
 targets:
-  jenkinsioNginxDigest:
-    name: "Jenkins.io nginx"
-    kind: yaml
-    spec:
-      file: "charts/jenkins.io/values.yaml"
-      key: image.tag
-    scm:
-      github:
-        user: "update-bot"
-        email: "update-bot@olblak.com"
-        owner: "jenkins-infra"
-        repository: "charts"
-        token: "x"
-        username: "olblak"
-        branch: "master"
+    jenkinsioNginxDigest:
+        name: Jenkins.io nginx
+        kind: yaml
+        spec:
+            file: charts/jenkins.io/values.yaml
+            key: image.tag
+        scmid: target_jenkinsioNginxDigest
+        sourceid: lastDockerDigest
```

I run the new command in the example directory, and I spot few configuration errors.

examples/updatecli.d/sources/dockerfile_sources.yaml
```diff
@@ -23,9 +23,9 @@ sources:
         kind: latest
 conditions:
   isENVSet:
-    sourceid: helm
-    name: Is the 2nd ENV instruction having a "keyword" set to "HELM_VERSION"
-    kind: dockerfile
+    sourceid: "helm1"
+    name: 'Is the 2nd ENV instruction having a "keyword" set to "HELM_VERSION"'
+    kind: "dockerfile"
```
dduportal, line 28:
```suggestion
    sourceid: helm1
    name: 'Is the 2nd ENV instruction having a "keyword" set to "HELM_VERSION"'
    kind: dockerfile
```

otherwise use the same quoting everywhere instead to be coherent ;)

olblak: That's true, I'll finish the cleanup :-)

olblak: I added quotes to all manifest examples under the examples directory. But I didn't fix the content of those have been unmaintained for a quite long time so maybe we should remove them at some point

olblak: they were just useful to test that `updatecli manifest upgrade` can update them to the latest syntax

dduportal (approved):
Tested on https://github.com/jenkins-infra/kubernetes-management/ with a recursive command update and it seems to work very well:
- Checkouted to an old commit (`dd34d184efea9cf10900019feb78410b9813aa70`)
- Checked the default behavior with `updatecli-next manifest upgrade --config updatecli/updatecli.d` : showed a LOT of possible changes
- Updated with the `--in-place` flag and ran the previous command again, which showed no update required!


I saw 2 things, not blocking the merge (subsequent PRs):

- I saw the `kind: helmChart` instructions not being updated to lower case
- The exit code of `updatecli-next manifest upgrade --config updatecli/updatecli.d` (dry run) is zero wether there are changes or not. A nice improvement would be to add a flag such as `https://www.terraform.io/cli/commands/plan#detailed-exitcode` to indicate through the exit code if a change is required or not (to ease pipeline processes)

Great feature buddy!

olblak:
Thanks very much testing. 

> The exit code of updatecli-next manifest upgrade --config updatecli/updatecli.d (dry run) is zero wether there are changes or not. A nice improvement would be to add a flag such as https://www.terraform.io/cli/commands/plan#detailed-exitcode to indicate through the exit code if a change is required or not (to ease pipeline process

Indeed that's a great suggestion and I think that it should be in a different PR so we can also handle the other commands
Meanwhile I am adding the "report" such 

```
++++++++++++++++++++
+ MANIFEST UPGRADE +
++++++++++++++++++++

Loading Pipeline "examples/updatecli.d/nginx.yaml"
WARNING: The directive 'scm' for the target["jenkinsioNginxDigest"] is now deprecated. Please use the new top level scms syntax
WARNING: The directive 'scm' for the target["jenkinsioNginxDigest"] is now deprecated. Please use the new top level scms syntax
✔ No Updatecli manifest change required



Summary
===========
Manifest(s) upgrade:
  * Changed:	0
  * Failed:	0
  * Skipped:	0
  * Succeeded:	1
  * Total:	1
```
Which should already highlight changes

Before merging, I sport some bugs that I need to fix

olblak:
>  saw the kind: helmChart instructions not being updated to lower case
This is because we don't validate that the kind value is lowercase. I am currently adding this to the current PR as

PR #681, opened 2022-05-06
Fix: scm concurrency issue when running git clone

### LINKED ISSUE #669: scm concurrency issue when running git clone

### Current Behavior

In order to speed up the `updatecli prepare` step, we execute `git clone` in parallel.
To identify unique git repository, we use the checksum of the `scm` configuration such as:

```
scms:
    default:
        kind: github
        owner: updatecli
        repository: updatecli
```

If for some reasons, we rely on two slightly different scm configuration such as different branches, then we clone in parallel the same repository twice to the same temporary directory which result in a incomplete repository on disk.

Instead of scm configuation checksum, we should use the path on disk return by (GetDirectory)[https://github.com/updatecli/updatecli/blob/c11a5b54ffd8d476e51767fce415698e5d0a0208/pkg/core/pipeline/scm/main.go#L34]

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: *
- updatecli: 0.24.0
```

### Anything else?

The code which need to be modified should only be the function (Clone)[https://github.com/updatecli/updatecli/blob/main/pkg/core/engine/main.go#L83]

olblak:
# Fix: scm concurrency issue when running git clone

Fix #669

Use git directory hash instead of scm spec hash.

## Test

Test with whatever manifest using 1 or multiple scm repository

### Tradeoff

/
### Potential improvement
/

### DIFF
--- a/pkg/core/engine/main.go
+++ b/pkg/core/engine/main.go
@@ -86,7 +86,9 @@ func Clone(
 	hashes *[]uint64,
 	wg *sync.WaitGroup) error {
 
-	hash, err := hashstructure.Hash(s, nil)
+	scmhandler := *s
+
+	hash, err := hashstructure.Hash(scmhandler.GetDirectory(), nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


olblak:
@dduportal This change may have side effect, just in case I am missing something

PR #694, opened 2022-05-10
Add option to disable Helm Version Increment

### LINKED ISSUE #691: Feature Request: Allow disabling Helm chart version increment

I am facing a situation where we want to rely on chart releaser to bump the Helm chart version but  I am still looking for the metadata such as appVersion.

### Solution you'd like

We already have a key `versionincrement` to configure how the version should be incremented. 
I would like to add "none" to the accepted values such as 

```
  noVersionInc: 
    name: Test Chart Update
    scmid: default
    kind: helmchart
    spec:
      name: chart/epinio
      file: Chart.yaml
      appversion: true
      key:  dependencies[2].version
      versionincrement: none
```

I considered adding a new boolean field such as `disableversionincrement` but I think it's better to reuse `versionincrement` 

### Anything else?

_No response_

olblak:
# Add option to disable Helm Version Increment

Fix #691 

## Test

```shell
cp <to_package_directory>
go test
```


pkg/plugins/resources/helm/main.go
```diff
@@ -34,19 +34,30 @@ URL:
 	MAJORVERSION = "major"
 	// PATCHVERSION defines patch version identifier
 	PATCHVERSION = "patch"
+	// NOINCREMENT defines if a chart version need to be updated
```
lemeurherve, line 37:
```suggestion
	// NOINCREMENT defines if a chart version doesn't need to be updated
```

pkg/plugins/resources/helm/main.go
```diff
@@ -34,19 +34,30 @@ URL:
...
 type Spec struct {
-	File             string // [target] Define file to update
-	Key              string // [target] Define Key to update
-	Name             string // [source][condition][target] Define Chart name path like "stable/chart"
-	URL              string // [source][condition] Define the chart location
-	Value            string // [target] Define value to set
-	Version          string // [source][condition]
-	VersionIncrement string // [target] Define the rule to incremental the Chart version, accept a list of rules
-	AppVersion       bool   // [target] Boolean that define we must update the App Version
+
```
lemeurherve, line 44:
```suggestion
```

pkg/plugins/resources/helm/main.go
```diff
@@ -34,19 +34,30 @@ URL:
...
+	// Defines the Helm Chart file to update, only for "target"
```
lemeurherve, line 45: I liked your `[target][...] Blabla` notation, any reason not to keep it?

pkg/plugins/resources/helm/main.go
```diff
@@ -34,19 +34,30 @@ URL:
...
+	Key string
+	// Defines the Chart name path like "stable/chart", only for "target"
+	Name string
+	// Defines the chart location URL, only for source and condition
+	URL string
+	// Defines the value to set for a key, only for "target"
+	Value string
+	// Defines the Chart version, only for "source" and "condition", default value set based on sourceinput value
+	Version string // [source][condition]
+	// Defines if a Chart change, triggers a Chart version update, target only, accept values is a comma separated list of "none,major,minor,patch"
```
lemeurherve, line 57:
```suggestion
	// Defines if a Chart changes, triggers (or not) a Chart version update, target only, accepted values is a comma separated list of "none,major,minor,patch"
```

pkg/plugins/resources/helm/utils.go
```diff
@@ -147,12 +150,18 @@ func (c *Chart) MetadataUpdate(chartPath string, dryRun bool) error {
 			md.Version = v.IncMinor().String()
 		case PATCHVERSION:
 			md.Version = v.IncPatch().String()
+		case NOINCREMENT:
+			// Reset Version to its initial value
+			md.Version = oldVersion
```
lemeurherve, line 155: Shouldn't the md.AppVersion also be restored here?

pkg/plugins/resources/helm/utils.go
```diff
@@ -147,12 +150,18 @@ func (c *Chart) MetadataUpdate(chartPath string, dryRun bool) error {
...
+			break forLoop
 		default:
-			logrus.Errorf("Wrong increment rule %q.", inc)
+			logrus.Warningf("Wrong increment rule %q, ignoring", inc)
 		}
 	}
 
-	logrus.Infof("\tChart Version updated from %q to %q\n", oldVersion, md.Version)
+	if oldVersion != md.Version {
+		logrus.Infof("\tChart Version updated from %q to %q\n", oldVersion, md.Version)
```
lemeurherve, line 163: Should a log be also included if md.AppVersion has changed as well?

lemeurherve (changes requested):

pkg/plugins/resources/helm/main.go
olblak: I also prefer it but I thought it would less obvious. Let me rollback this.

pkg/plugins/resources/helm/utils.go
olblak: It's already done on line 127

pkg/plugins/resources/helm/utils.go
olblak: No because it's not needed. The chart version can be updated up to three times if spec.semverinc=[major,minor,path,none]
While normally "none" should be used alone, I decided to handle the case where "none" is accidently used with one of the other "rule".
Let me add warning if it's the case

pkg/plugins/resources/helm/utils.go
lemeurherve: WDYT of putting the block l125-129 below this one?

pkg/plugins/resources/helm/utils.go
lemeurherve: +1 for the warning, I would even throw an error from validate()

olblak:
Something weird is happening with the e2e test. For some reason the manifest `e2e/updatecli.d/success.d/source/gitTag.yaml` succeed as part of a group of manifest but succeed when execute alone.

I could reproduce the issue with v0.25.0

lemeurherve (approved):

PR #713, opened 2022-06-07
feat(SCM) Introduces a "local" SCM with autoguessing

dduportal:
Ref. https://github.com/updatecli/updatecli/issues/716

Depends on #717 

This change request introduces a new feature for Updatecli: an automatic SCM identified as `local`.

The "automatic guess" feature is not complete: for now, it only extracts some of the atrributes, the others have to be specified through manifest or env. vars. I expect more PRs to improve it along the way.
That's why the associated issue won't be closed.

For now:
- URL for git
- Owner/repository for github

Please note that:

- A new go module dependency was added: `github.com/goware/urlx`. Not strictly necessarly, but really useful to parse the hostname of the github SSH addresses. Might be doable with a good regexp, but that was too distracting for me to write such (as it needs tests, etc.).
- The `gitgeneric` package now provides an interface, implemented by the default `GoGit{}` struct. Should not change anything (plugins `git`, `github` and others have been updated) but it allows mocking (a basic mock is provided because used by the code introduced in this PR) and would help in unit testing any SCM-dependent behavior.
- The `config` package was split (`errors.go` to store the errors, `defaults.go` for the logic around setting up default configuration). Be careful with rebases!
- The `scm` package was split to have a clear source file separation between `scm.Scm` behaviors and `scm.Config` behaviors (Validate was confusing before)
- There are 2 `Merge` methods for `git` and `github` SCM plugins. This code could definitivelt be improved in the future, but it's around the whole "configuration definition" + "deep merge tree structures with zero-ed values". Too much compared to writing the 2 raws methods. Can be improved, but since it's tested and since we do not define new SCM plugins everyday, it's worth it for me.
- Unit tests, unit tests and unit tests
- The end to end manifests where modified to disable the `local` SCM as none of them use it (to avoid unplanned info/warning message).

## Test

```shell
make test
make test-e2e
```

## Additional Information

- The end to end tests where modified to make its easier to run locally.
  - A change has been introduced in the `Makefile` to explain the env. var `$VENOM_VAR_binpath` but also to fail fast with a comprehensive message when missing env vars are absent (instead of starting venom which fails)
  - We can still improve the contributor experience for E2E but later (splitting tests for easier "catch which onefailed") :)
  - No more `values.yaml` for now. If we want to end-to-end testing of the templating of values, we have to add a specific test for this. Otherwise it adds too much varying elements. Another idea: we can also make this file back, but the Makefile would parse it to list the required environment variables in a centralized location).
  -


pkg/core/config/main.go
```diff
@@ -45,6 +45,8 @@ var (
 	ErrNotAllowedTemplatedKey = errors.New("not allowed templated key")
 )
 
+const localScmIdentifier = "local"
```
olblak, line 48: This is just a suggestion

```suggestion
const (
    // LOCALSCMIDENTIFIER defines the scm id used to configure the local scm directory  
    LOCALSCMIDENTIFIER string = "local"
    localScmIdentifier = "local"
)
```

pkg/core/pipeline/scm/main.go
```diff
@@ -2,18 +2,24 @@ package scm
...
 	"github.com/updatecli/updatecli/pkg/plugins/scms/github"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/gitgeneric"
 )
 
 type Config struct {
-	Kind string
-	Spec interface{} `jsonschema:"type=object"`
+	Kind     string
+	Spec     interface{} `jsonschema:"type=object"`
+	Disabled bool
```
olblak, line 22:
```suggestion
        // Disabled is a setting used to disable the local git repository auto configuration
	Disabled bool
```

pkg/core/pipeline/scm/main.go
```diff
@@ -115,6 +119,63 @@ func (s *Scm) GenerateSCM() error {
...
+// GetScmConfigFromDir returns the Scm configuration guessed from a directory on the filesystem, using the gitHandler passed as argument
+// An error is return if it fails to recognize the kind of SCM or if something goes wrong
+func GetScmConfigFromDir(workingDir string, gitHandler gitgeneric.GitHandler) (Config, error) {
+	remotes, err := gitHandler.RemoteURLs(workingDir)
+	if err != nil {
+		return Config{}, err
+	}
+	logrus.Debugf("Found the following remotes in %s: %s", workingDir, remotes)
+
+	originRemoteURL, ok := remotes["origin"]
```
olblak, line 131: I am wondering if it's worthwhile to allow the configuration of originremote via the scm spec

pkg/core/pipeline/scm/config.go
```diff
@@ -0,0 +1,154 @@
...
+	case "github.com":
+		slashEverywhereUrl := strings.Split(strings.ReplaceAll(originRemoteURL, ":", "/"), "/")
+		if len(slashEverywhereUrl) < 3 {
+			return fmt.Errorf("unable to parse the following GitHub remote URL: %s", slashEverywhereUrl)
+		}
+
+		autoguessSpec := github.Spec{
+			Repository: strings.TrimSuffix(slashEverywhereUrl[len(slashEverywhereUrl)-1], ".git"),
+			Owner:      slashEverywhereUrl[len(slashEverywhereUrl)-2],
+			Directory:  workingDir,
```
olblak, line 94: What do you think to guess the username and password based on "GITHUB_ACTOR" and "GITHUB_TOKEN" to align with github action behavior

pkg/core/pipeline/scm/config.go
```diff
@@ -0,0 +1,154 @@
...
+			Branch:     "main",
```
olblak, line 95: We should track that we need to automatically detech the branch as well in a separated PR/issue

pkg/core/pipeline/scm/config.go
dduportal: I would prefer to use the pattern proposed in the PR, with explicit UPDATECLI_ vars
- The UPDATECLI_ pattern is more deterministic and void accidental usages
- These 2 variables (`GITHUB_*`) are complicated to overrides (if you want to use another token with more permissions for instance)

e2e/updatecli.d/deprecated.d/transformers.yaml
```diff
@@ -1,5 +1,9 @@
 name: "Test transformers"
 
+scms:
+  local:
+    disabled: true
```
lemeurherve, line 5: How about `enabled: false` instead? Seems more common to have a "positive" expression as field name

pkg/core/config/defaults.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
+)
+
+// GetChangelogTitle try to guess a specific target based on various information available for
+// a specific job
+func (config *Config) GetChangelogTitle(ID string, fallback string) (title string) {
+	if len(config.Title) > 0 {
+		// If a pipeline title has been defined, then use it for pull request title
+		title = fmt.Sprintf("[updatecli] %s",
```
lemeurherve, line 15: The `[updatecli]` prefix is already harcoded like that elsewhere?

dduportal: I've only changed the method from one filte to the other, I won't touch this method on this PR.

But that is a good question that you raise though :)

e2e/updatecli.d/deprecated.d/transformers.yaml
dduportal: Depends on the languages 😅
- Human language: depends on the culture and people profiles and there isn't really a "best approach".
- Technical language (see below) with zero-ed value make it easier to have a "disabled" attribute

The idea is the following:

- Default behavior is "feature" is enabled
- You want to provide a switch to have disable it if required
- The "zero" value for a boolean in Golang (same problem as helm templates or Groovy language) is `false`. 
- If you want an "enable", then you have to write code for all initialization path that would check the inverted value, since it would be "enable:  false` by default. Much more code to write.

lemeurherve: You're lazy, I understand 😄 😉

pkg/core/pipeline/scm/config.go
olblak: Sounds good to me, we can have smaller PR later.

pkg/core/config/defaults.go
olblak: > The `[updatecli]` prefix is already harcoded like that elsewhere?

In the same function yes 😆 , that function could be better written but well 🤷🏾

pkg/core/pipeline/scm/config.go
olblak: That's a good point. I started using `UPDATECLI_GITHUB_TOKEN` so I don't conflict with Github Action environment but at the same time `GITHUB_TOKEN` and `GITHUB_ACTOR` became pretty standard so I am fine with both

olblak:
Overall, it's a great PR. Thanks for paving the way 
Considering the size of the work needed here, I totally support merging this PR even if it doesn't solve the global issue.

But before merging I would like to clarify one thing

## Disable local scm 

I don't understand why and when to use the following block

```
scms:
  local:
    disabled: true
```

pkg/core/config/defaults.go
```diff
@@ -0,0 +1,81 @@
...
+		logrus.Warning("**Fallback** Please add a title to you configuration using the field 'title: <your pipeline>'")
+		title = fmt.Sprintf("[updatecli][%s] Bump version to %s",
+			config.Sources[config.Targets[ID].SourceID].Kind,
+			fallback)
+	}
+	return title
+}
+
+// EnsureLocalScm ensures that the config receiver has an
+// SCM configuration "local" if it is needed
```
lemeurherve, line 34:
```suggestion
// EnsureLocalScm ensures the config receiver has a
// "local" SCM configuration if needed
```

pkg/core/config/errors.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// ErrBadConfig is returned when updatecli try to read
+	// a wrong configuration.
+	ErrBadConfig = errors.New("wrong updatecli configuration")
+
+	// ErrNoEnvironmentVariableSet is returned when during the templating process,
+	// it tries to access en environment variable not set.
+	ErrNoEnvironmentVariableSet = errors.New("environment variable doesn't exist")
+
+	// ErrNoKeyDefined is returned when during the templating process, it tries to
```
lemeurherve, line 18:
```suggestion
	// ErrNoKeyDefined is returned when during the templating process, updatecli tries to
```

pkg/core/config/main.go
```diff
@@ -109,6 +96,11 @@ func New(cfgFile string, valuesFiles, secretsFiles []string) (config Config, err
 	// config.PipelineID is required for config.Validate()
 	config.PipelineID = pipelineID
 
+	// Ensure that there is a local SCM defined as specified
```
lemeurherve, line 99:
```suggestion
	// Ensure there is a local SCM defined as specified
```

pkg/core/pipeline/scm/config.go
```diff
@@ -0,0 +1,154 @@
...
+		return fmt.Errorf("no remote named 'origin' could be found in the repository %s", workingDir)
+	}
+
+	var remoteHostname string
+	httpRemoteUrl, err := url.Parse(originRemoteURL)
+	if err != nil {
+		/**
+		Case of non HTTPS scheme: either git or SSH (github.com for instance).
+		Hostname extraction on the "left" of the colon character (':')
+		**/
```
lemeurherve, line 75:
```suggestion
		// Case of non HTTPS scheme: either git or SSH (github.com for instance).
		// Hostname extraction on the "left" of the colon character (':')
```
I don't see the point of using another type of comment.

pkg/core/pipeline/scm/main.go
```diff
@@ -27,7 +20,7 @@ var (
 	ErrWrongConfig = errors.New("wrong scm configuration")
 )
 
-// Scm is an interface that offers common functions for a source control manager like git or github
+// ScmHandler is an interface that offers common functions for a source control manager like git or github
```
lemeurherve, line 23:
```suggestion
// ScmHandler is an interface offering common functions for a source control manager like git or github
```

pkg/plugins/utils/gitgeneric/main_test.go
```diff
@@ -141,3 +144,45 @@ func TestTagsIntegration(t *testing.T) {
...
+			gotRemotes, gotErr := g.RemoteURLs(tt.workingDir)
+
+			if tt.wantErr {
+				require.Error(t, gotErr)
+				return
+			}
+
+			require.NoError(t, gotErr)
+			// Only testing that there is an "origin" remote with a non empty URL.
+			// Because origin's URL, as well as other remotes, depend on the user or CI configuration and is not deterministic.
```
lemeurherve, line 183:
```suggestion
			// Because origin's URL, as well as other remotes, depends on the user or CI configuration and is not deterministic.
```

pkg/plugins/utils/gitgeneric/mocks.go
```diff
@@ -0,0 +1,12 @@
+package gitgeneric
+
+// MockGit is a stub implementation of the `GitHandler` interface to be used in our unit test suites.
+type MockGit struct {
+	GitHandler // Ensure that any unspecified method from this interface are still declared
```
lemeurherve, line 5:
```suggestion
	GitHandler // Ensure any unspecified method from this interface are still declared
```

lemeurherve (dismissed):
Minor comment nitpicks

Great job! 🎉

e2e/values.yaml
```diff
@@ -1,4 +1,4 @@
-github:
-  token: "GITHUB_TOKEN"
-  username: "GITHUB_ACTOR"
-  branch: main
+# github:
```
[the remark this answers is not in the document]
dduportal: Oh, I thought I removed it. good catch!

lemeurherve (approved):

dduportal:
Rebase in progress (with conflict merge)

dduportal:
@olblak Ready to review (approve/merge if ok).

I've done full testing on jenkins-infra/kubernetes-management as usual with a `updatecli diff` with this PR.
Also tested the full "apply" on https://github.com/dduportal/docker-hashicorp-tools/pulls (3 opened PRs).

olblak (approved):
🎉

PR #715, opened 2022-06-10
pkg/core/config: replace deprecated `strings.Title` with `cases.Title`

### LINKED ISSUE #668: strings.Title is has been deprecated since G1.18

```
strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (SA1019)
```

### Solution you'd like

Look in updatecli codebase to replace `strings.Title`

Maybe it's time to introduce a library to improve updatecli UI such as https://charm.sh/

### Anything else?

_No response_

jlevesy:
# Replace deprecated calls to `strings.Title` with `cases.Title`

Fixes #668

This PR replaces [all the callsites of `strings.Title`](https://github.com/updatecli/updatecli/search?q=strings.Title) with the recommended `cases.Title` method, configured for 'English', with the `cases.NoLower` option to get as close as possible from the behavior of `strings.Title` (https://go.dev/play/p/0lQ9iHPi_Ro)

This change doesn't seem to break the test suite, so that's a first step. I'm unsure how to properly test this though.

## Test

```shell
go test ./pkg/core/config/
```

### Tradeoff

None

### Potential improvement

None

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -13,6 +13,9 @@ import (
 	"text/template"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+	"gopkg.in/yaml.v3"
 
 	"github.com/updatecli/updatecli/pkg/core/pipeline/condition"
 	"github.com/updatecli/updatecli/pkg/core/pipeline/pullrequest"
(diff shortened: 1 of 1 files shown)


olblak:
thanks for the PR, the change should be fine.

olblak (approved):

PR #700, opened 2022-05-21
Fix scm resource initialization

olblak:
Signed-off-by: Olblak <me@olblak.com>

# Fix scm resource initialization

Fix #699 

The code base evolved in a situation where `scm.Init` was called in different places. That led to have some variable not set in some cases. Instead I move that logic in the `New` function that create SCM spec. I also fix a configuration bug where we couldn't configure GitHub commit behavior. 

In summary:
* Fix git/github scm initialization
* Add comment on git spec and GitHub spec to generate description in Jsonschema
* Expose commit configuration for GitHub SCM resource

## Test

To test this pull request, you can run the following commands:

```shell
./bin/updatecli diff --config e2e/updatecli.d/success.d/source --values e2e/values.yaml --debug --clean=false 
ls /tmp/updatecli/github_com_updatecli_updatecli_git
```

PR #717, opened 2022-06-12
fix(github) initialize the headbranch 

dduportal:
Fix the issue introduced in #700 (ref. https://github.com/updatecli/updatecli/pull/700#issuecomment-1153273824)

## Test

```shell
make test
make test-e2e
```

Also tested on jenkins-infra/kubernetes-management

## Additional Information

- Added unit tests to underline the error before fixing
- Fixed a typo
- Removed the pointer/references to the pipelineID string. Not sure the reason why pointer was used?


olblak (approved):
Nice catch, this PR didn't seam to be easy to work on

PR #650, opened 2022-04-14
Add `spec.files` for Condition and Target of `kind` `yaml`

lemeurherve:
# feat(yaml): implement spec.files for kind: yaml

Follow-up of #490
Ref #435

Allow to specify more than one file as target for `kind: yaml`

Like what have been done for `kind: file`, the possibility to specify multiple files is only possible in the `targets` stage.

## Test

To test this pull request, you can run the following commands:

```shell
cp <to_package_directory>
go test
```

PR #729, opened 2022-06-21
feat(yaml): implement `spec.files` for `kind: yaml` conditions too

lemeurherve:
Follow-up of #650, WIP

## Test

```shell
cp <to_package_directory>
go test
```


pkg/plugins/resources/yaml/target.go
```diff
@@ -60,55 +66,89 @@ func (y *Yaml) target(source string, dryRun bool) (bool, []string, string, error
...
 		}
+		message.WriteString(fmt.Sprintf("Update key %q from file %q", y.spec.Key, filePath))
+
 	}
 
-	files = append(files, y.spec.File)
+	// If no file was updated, return an error
+	// TODO: why?
+	if notChanged == len(y.files) {
+		return false, files, message.String(), fmt.Errorf("no file was updated")
```
olblak, line 148: Why are you returning an error if nothing changed? 
Wouldn't it be better to return something like
```suggestion
		return false, files, message.String(), nil
```

lemeurherve: https://github.com/updatecli/updatecli/pull/650/commits/4de4f95c7b4509f1ce1377b4ac2efb1d212f7ec4

Issue #731, opened 2022-06-24
Values file is not accepted


kuisathaverat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

the execution fails with the following output

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "update-test.yml"
ERROR: wrong file extension ".yml" for file "values.yml"
ERROR: failed loading pipeline(s)
        * "update-test.yml" - wrong file extension ".yml" for file "values.yml"
ERROR: ✗ no valid pipeline found
ERROR: command failed
```

### Expected Behavior

the pipeline is executed

### Steps To Reproduce

* Create a pipeline file like this one `update-test.yml`
```yaml
---
title: Update test values
sources:
  dummy:
    name: "Dummy"
    kind: shell
    spec:
      command: echo {{ .Value }}
targets:
  dummy:
    name: "Dummy"
    kind: shell
    spec:
      command: echo {{ .Value }}
```
* Create a values file like this one `values.yml`
```
Value: "foo"
```
* execute the diff command
```
> updatecli diff --config update-test.yml --values values.yml --debug


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "update-test.yml"
ERROR: wrong file extension ".yml" for file "values.yml"
ERROR: failed loading pipeline(s)
        * "update-test.yml" - wrong file extension ".yml" for file "values.yml"
ERROR: ✗ no valid pipeline found
ERROR: command failed
```

I have checked that the pipeline is executed without the values files
```
> updatecli diff --config update-test.yml 


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "update-test.yml"

SCM repository retrieved: 0


++++++++++++
+ PIPELINE +
++++++++++++



######################
# UPDATE TEST VALUES #
######################


SOURCES
=======

dummy
-----
The shell 🐚 command "echo <no value>" ran successfully with the following output:
----
<no value>
----


TARGETS
========

dummy
-----

**Dry Run enabled**

The shell 🐚 command "echo <no value> <no value>" ran successfully with the following output:
----
<no value> <no value>
----

=============================

REPORTS:


⚠ UPDATE-TEST.YML:
        Sources:
                ✔ [dummy] Dummy (kind: shell)
        Target:
                ⚠ [dummy] Dummy (kind: shell)



Run Summary
===========
Pipeline(s) run:
  * Changed:    1
  * Failed:     0
  * Skipped:    0
[...]

PR #732, opened 2022-06-24
fix: check if the extension is not `yaml` and not `yml`

kuisathaverat:
# check if the extension is not `yaml` and not `yml`

Fix https://github.com/updatecli/updatecli/issues/731

## Test

```shell
cat >EOF > pipeline.yml 
---
title: Update test values
sources:
  dummy:
    name: "Dummy"
    kind: shell
    spec:
      command: echo {{ .Value }}
targets:
  dummy:
    name: "Dummy"
    kind: shell
    spec:
      command: echo {{ .Value }}
EOF
cat <EOF > values.yml
Value: "foo"
EOF
cp pipeline.yml pipeline.yaml
cp values.yml values.yaml 
go run main.go diff --config pipeline.yml --values values.yml
go run main.go diff --config pipeline.yaml --values values.yml
go run main.go diff --config pipeline.yml --values values.yaml
go run main.go diff --config pipeline.yaml --values values.yaml
```

### DIFF
--- a/pkg/core/config/template.go
+++ b/pkg/core/config/template.go
@@ -104,9 +104,8 @@ func ReadFile(filename string, values *map[string]interface{}, encrypted bool) (
 
 	baseFilename := filepath.Base(filename)
 
-	if extension := filepath.Ext(baseFilename); strings.Compare(extension, ".yaml") != 0 ||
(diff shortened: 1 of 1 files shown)


olblak (approved):
Oops

PR #744, opened 2022-06-29
feat: Add Gitea support

### LINKED ISSUE #737: Feature Request: Add SCM gittea support

I would like to use updatecli to automate files stored on gittea repositories

### Solution you'd like

```
scms:
  default:
    kind: gittea
    spec:
      user: updatecli
      email: me@olblak.com
      owner: updatecli
      repository: updatecli
      token: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
      branch: main
```

_No response_

### Anything else?

_No response_

olblak:
# Add gitea support

Fix #737 

- [x] Add new resource `gitea/release` to interact with Gitea releases, where:
    - [x]  [source]  can retrieve a release
    - [x] [condition] can test if a release exist
    - [x] [target] can create a release
- [x] Add new resource `gitea/branch` to interact with Gitea branches via the API
    - [x]  [source]  can retrieve a branch
    - [x] [condition] can test if a branch exists
    - [x] [target] creating branches won't be supported in this PR
- [x] Add new resource `gitea/tag` to interact with Gitea tags via the API
    - [x]  [source]  can retrieve a tag
    - [x] [condition] can test if a tag exists
    - [x] [target] creating tags via API won't be supported in this PR
- [x] Add new resource pullrequest resource`gitea` to interact with Gitea pullrequest via the API
- [x] Documentation

## Test

```shell
go build -o ./bin/updatecli . && ./bin/updatecli apply --config e2e/updatecli.d/success.d/gitea/
```

## Additional Information

To avoid collision between gitea and github scm temporary directory, it now contains the service.
So github will clone repositories in `/tmp/updatecli/github/...` and gitea in `/tmp/updatecli/gitea/...` 

### Tradeoff

Now that I am implementing the integration with a second git service, I realize that having a specific scm resource per git service add no additional value.


pkg/plugins/scms/github/main.go
```diff
@@ -70,7 +70,7 @@ func New(s Spec, pipelineID string) (*Github, error) {
 	}
 
 	if s.Directory == "" {
-		s.Directory = path.Join(tmp.Directory, s.Owner, s.Repository)
+		s.Directory = path.Join(tmp.Directory, "github", s.Owner, s.Repository)
```
dduportal, line 73: What is the reason of this change? It feels weird

olblak:
I should move the two following commits to a separated pullrquest 
* https://github.com/updatecli/updatecli/pull/744/commits/7de0e932158305b59ba4eeba7ecb506c0da69e25
* https://github.com/updatecli/updatecli/pull/744/commits/8d02c89e30bc0a5cf0852ea6c2f102a7543460f6

pkg/plugins/scms/github/main.go
olblak: I made a comment in the PR description

_To avoid collision between gitea and github scm temporary directory, it now contains the service.
So github will clone repositories in /tmp/updatecli/github/... and gitea in /tmp/updatecli/gitea/..._

olblak: Another option would be to do like git scm and creating the directory `github_com_updatecli_updatecli` and `try_gitea_io_updatecli_updatecli`

olblak: In facts ,I should have add this comment to the code

dduportal: Why would there be a collision ? I’m not sure to see the difference with current updatecli manifest with different scm defined ?

olblak: But a good refactoring would be to decide the path outside the scm plugin

dduportal: Interesting case: this is a bug and should be treated outside this PR as well honestly.

=> proposal: wdyt about using the SCM identifier as it would NEVER be possible to have the same values (instead of the type that could still lead to collisions?

eg. with your manifest:

```
/tmp/updatecli/example1/updatecli/updatecli
/tmp/updatecli/example2/updatecli/updatecli
```

olblak:
> I should move the two following commits to a separated pullrquest
> 
>     * [7de0e93](https://github.com/updatecli/updatecli/commit/7de0e932158305b59ba4eeba7ecb506c0da69e25)
> 
>     * [8d02c89](https://github.com/updatecli/updatecli/commit/8d02c89e30bc0a5cf0852ea6c2f102a7543460f6)

The pullrequest have been open.

I need to fix the e2e tests and then the PR should be ready

pkg/plugins/scms/github/main.go
olblak: > Interesting case: this is a bug and should be treated outside this PR as well honestly.

IMHO I wouldn't consider it as a bug because it's a side effect introduced by this pullrequest.

While I am open to use a different schema, using the scmid seems wrong to me.
We commonly use the same scmid accross manifest such as default, local,etc
So the risk is higher.

Another option would be to use the same approach than the git repository so for example `https://github.com/updatecli/updatecli.git` becomes `/tmp/updatecli/https_github_com_updatecli_updatecli_git`

dduportal: > IMHO I wouldn't consider it as a bug because it's a side effect introduced by this pullrequest.

Fair point :)

> While I am open to use a different schema, using the scmid seems wrong to me. We commonly use the same scmid accross manifest such as default, local,etc So the risk is higher.

I have to strongly disagree: as we use a map (in the YAML manifest) which keys are the scmids: you cannot have 2 times the same id. If you have,n then something went terribly wrong. So the risk of 2 different SCM colliding is absolute zero (if you name the temp. dir with the scm id).

About the risk of collision with other objects: it means that the temp dir is falt with a list of ALL the temp dir that could be created. If you see a risk of collision due to naming, then it means that you have to namespace, otherwise you'll have a nasty bug hard to debug quite soon (even outisde this PR).
WDYT to namespace with the resource type (either with subdirectories or part of the name)? 

eg. `${TMP_DIR}/updatecli/scms/example1/` and `${TMP_DIR}/updatecli/scms/example2/` in your example?
(and `${TMP_DIR}/updatecli/source/example1/` , `${TMP_DIR}/updatecli/target/example2/`, etc.)?

> 
> Another option would be to use the same approach than the git repository so for example `https://github.com/updatecli/updatecli.git` becomes `/tmp/updatecli/https_github_com_updatecli_updatecli_git`

I find this example pretty hard to understand and read. It does not tell me immediatly what is this folder for. However it is an implementation detail= end user will never check it.

dduportal: > The thing is by design, when scm repositories are cloned locally in the temporary directory, they are reused across the different pipelines. So that we don't have to delete them and re-clone them between each pipeline.
> Which means that using the scmid, doesn't tell if ${TMP_DIR}/updatecli/scms/example1/ is a scm of type github or gitea

I don't see why this behavior requires the "kind" of scm to be know in the path? 

Additionnaly, I consider the "cross pipeline caching" to be both a bug (I can't reliably determine the outcome of updatecli without having read updatecli source code) and a security issue (code injection through deterministic caching). 
I would expect caching to be namespaces per pipeline.

pkg/plugins/resources/gitea/branch/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+	results := []string{}
+	for _, branch := range branches {
+		results = append(results, branch.Name)
+	}
+
+	return results, nil
+}
+
+func (s *Spec) Validate() error {
```
dduportal, line 127:
```suggestion
func (s Spec) Validate() error {
```

:)

dduportal:
As discussed together, here are some resources if you want to improve the e2e harness for gitea (later):

- https://github.com/oufti-playground/lab-vm/blob/master/docker/docker-compose.yml#L58-L86
- https://github.com/oufti-playground/lab-vm/blob/master/docker/gitserver/Dockerfile
- https://github.com/oufti-playground/lab-vm/blob/master/docker/gitserver/setup-gitea.sh

pkg/plugins/resources/gitea/client/main.go
```diff
@@ -0,0 +1,65 @@
...
+				),
+			},
+		}
+	}
+
+	return client, nil
+
+}
+
+func (s *Spec) ValidateClient() (err error) {
```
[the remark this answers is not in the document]
olblak: Same here

olblak:
I spot a few minor bugs which I need to solve

pkg/plugins/resources/gitea/branch/source.go
```diff
@@ -0,0 +1,47 @@
...
+		return "", nil
+	}
+
+	g.foundVersion, err = g.Spec.VersionFilter.Search(versions)
+
+	if err != nil {
+		switch err {
+		case version.ErrNoVersionFound:
+			logrus.Infof("%s No Gitea branches found matching pattern %q", result.FAILURE, g.versionFilter.Pattern)
+			return "", nil
```
olblak, line 29: I am now wondering if I should return an error or not :/ 
If I don't return an error then a pipeline output looks like 

```
#####################
# TEST GITEA BRANCH #
#####################


SOURCES
=======

default
-------
Searching for version matching pattern "~0.99"
✗ No Gitea branches found matching pattern "~0.99"

⚠ empty source returned


CONDITIONS:
===========

default
-------
✔ Gitea branch "main" found

=============================

REPORTS:


✔ Test Gitea branch:
	Source:
		⚠ [default]  (kind: gitea/branch)
	Condition:
		✔ [default]  (kind: gitea/branch)

Run Summary
===========
Pipeline(s) run:
  * Changed:	0
  * Failed:	0
  * Skipped:	0
  * Succeeded:	1
  * Total:	1

```

olblak: If I return an error then the pipeline stop right after the source

pkg/plugins/resources/gitea/client/main.go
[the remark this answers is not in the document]
olblak: done

pkg/plugins/resources/gitea/branch/source.go
olblak: In the case of the github release, it returns an error

olblak: After consideration, I went to the return error like githubreleas

dduportal: I vote for the error as well. I expect my updatecli pipelines to fail if an empty source is returned.

pkg/plugins/resources/gitea/client/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/drone/go-scm/scm/transport/oauth2"
+	"github.com/sirupsen/logrus"
+)
+
+// Spec defines a specification for a "gitea" resource
+// parsed from an updatecli manifest file
+type Spec struct {
+	// [S][C][T] URL specifies the default github url in case of Gitea enterprise
+	URL string `yaml:",omitempty" jsonschema:"required"`
+	// Username specifies the username used to authenticate with Gitea API
```
olblak, line 19:
```suggestion
	// [S][C][T] Username specifies the username used to authenticate with Gitea API
```

PR #745, opened 2022-06-30
By default Use temporary working branch with git scm

olblak:

# By default Use temporary working branch with git scm

Requirement for https://github.com/updatecli/updatecli/pull/744

## Test

To test this pull request, I used the following manifest to manifest that in one case we have a temporary branch and not in the second situation

```shell
title: "Test git scm resource"

scms:  
  tmpDir:
    kind: git 
    spec:
      url: https://github.com/updatecli/updatecli.git
      disableworkingbranch: true
      branch: main
  noneTmpDir:
    kind: git 
    spec:
      url: https://github.com/updatecli/website.git
      branch: master

sources:
  default:
    kind: jenkins

targets: 
  tmpDir:
    name: "tmpdir"
    kind: file
    scmid: tmpDir
    spec:
      file: README.adoc
  noneTmpDir:
    name: "testnonetmpdir"
    kind: file
    scmid: noneTmpDir
    spec:
      file: README.md

```

### Tradeoff

Up to know the default behavior for git scm was to not use a temporary branch such as "updatecli_xxxx". To keep the current manifest working we could provide an ugprade path where automatically set "disableworkingbranch: true" for version older than 0.27, and we had a warning message

### Potential improvement

Now that I am experimentation with gitea support, I realize that scm of type other than "svn, or mercurial" doesn't add any value and quite the opposite make the code more difficult to maintain. On mid term, I am considering deprecating the github scm kind

### DIFF
--- a/pkg/plugins/scms/git/main.go
+++ b/pkg/plugins/scms/git/main.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"strings"
@@ -23,6 +24,8 @@ type Spec struct {
 	Password string `yaml:",omitempty"`
 	// Branch specifies the git branch
 	Branch string `yaml:",omitempty"`
+	// DisableWorkingBranch disables temporary working branch such as updatecli_xxx
+	DisableWorkingBranch bool `yaml:",omitempty"`
 	// User specifies the git commit author
 	User string `yaml:",omitempty"`
 	// Email specifies the git commit email
@@ -38,12 +41,12 @@ type Spec struct {
 }
 
 type Git struct {
-	spec         Spec
-	remoteBranch string
+	spec       Spec
+	HeadBranch string
 }
 
 // New returns a new git object
-func New(s Spec) (*Git, error) {
+func New(s Spec, pipelineID string) (*Git, error) {
 	var err error
 	if len(s.Directory) == 0 {
 		s.Directory, err = newDirectory(s.URL)
@@ -56,10 +59,18 @@ func New(s Spec) (*Git, error) {
 		s.Branch = "main"
 	}
 
-	return &Git{
-		spec:         s,
-		remoteBranch: git.SanitizeBranchName(s.Branch),
-	}, nil
+	g := Git{
+		spec: s,
+	}
+
+	switch s.DisableWorkingBranch {
+	case true:
+		g.HeadBranch = git.SanitizeBranchName(s.Branch)
+	case false:
+		g.HeadBranch = git.SanitizeBranchName(fmt.Sprintf("updatecli_%s", pipelineID))
+	}
+
+	return &g, nil
 
 }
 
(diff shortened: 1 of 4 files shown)


olblak:
Missing the migration warning

olblak:
Missing the warning message

olblak:
> Up to know the default behavior for git scm was to not use a temporary branch such as "updatecli_xxxx". To keep the current manifest working we could provide an ugprade path where automatically set "disableworkingbranch: true" for version older than 0.27, and we had a warning message

Well in fact, I don't see any other way than keeping the current behavior when used without pullrequest as I don't want to introduce breaking changes.

So I am going  to check if a scm of kind "git" is used alone, if it's the case then I just print a warning message saying that the behavior will be modified in coming releases so we give enough time to run the migration tool `updatecli manifest upgrade`

pkg/plugins/scms/git/main.go
```diff
@@ -56,10 +59,18 @@ func New(s Spec) (*Git, error) {
...
+	g := Git{
+		spec: s,
+	}
+
+	switch s.DisableWorkingBranch {
+	case true:
+		g.HeadBranch = git.SanitizeBranchName(s.Branch)
+	case false:
+		g.HeadBranch = git.SanitizeBranchName(fmt.Sprintf("updatecli_%s", pipelineID))
+	}
```
dduportal, line 71: This logic should not be there, as it introduces a lot of coupling. I expect the code, responsible to call the `New` method, to pass the branch name in the provied spec.

pkg/plugins/scms/github/main.go
```diff
@@ -84,8 +86,14 @@ func New(s Spec, pipelineID string) (*Github, error) {
...
-		Spec:       s,
-		HeadBranch: git.SanitizeBranchName(fmt.Sprintf("updatecli_%v", pipelineID)),
+		Spec: s,
+	}
+
+	switch s.DisableWorkingBranch {
+	case true:
+		g.HeadBranch = git.SanitizeBranchName(s.Branch)
+	case false:
+		g.HeadBranch = git.SanitizeBranchName(fmt.Sprintf("updatecli_%s", pipelineID))
```
dduportal, line 96: Same as my other comment: this logic is duplicated and should not be there

dduportal (changes requested):
This PR is rollbacking some major changes on the SCM, and introduces code duplication and increased coupling, so can we discuss (sync or async) which problem should it solve?

olblak:
I am putting this PR on old until we finish the work introduced by @dduportal 

https://github.com/updatecli/updatecli/pull/713

olblak:
I think I'll have another look to this but I am considering changing the parameter name from `disableworkingbranch` to `workflow`
Where workflow could accept the following parameters
* "centralized", means commit directly to the branch defined in the manifest
* "feature-branch", means commits to a temporary branch
* "github/gitea", means commits to a temporary branch and open a pullrequest

In

dduportal:
> 

Can you clarify, it is not really clear?

olblak:
> > 
> 
> Can you clarify, it is not really clear?

No nevermind, I'll try to have a look to your PR and the issue you created.
You better captured and explained the problem.

PR #754, opened 2022-07-05
feat: Allow to set environment variable within shell resources

### LINKED ISSUE #749: Shell resource don't inherit environment variable

### Current Behavior

Scripts called by the shell resource only have access to the environment variable `DRY_RUN`. More context from the two following links

```
	// If Env is nil, the new process uses the current process's
	// environment.
```
=>  https://pkg.go.dev/os/exec#Cmd

But it's set to 
```
Env: []string{fmt.Sprintf("DRY_RUN=%v", dryRun)},
```
=> https://github.com/updatecli/updatecli/blob/main/pkg/plugins/resources/shell/target.go#L40

For example in the context of Github Action we need to inherit the updated env `PATH` based on cached tool so we can use custom command in script s. 

### Expected Behavior

This manifest https://github.com/epinio/docs/blob/main/.github/workflows/updatecli.yml should be working by using the command 

```
act workflow_dispatch -j updatecli --env "UPDATECLI_GITHUB_TOKEN=xxx" --env "UPDATECLI_GITHUB_ACTOR=yyy"
```

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: linux
- updatecli: 0.27.0
```

### Anything else?

I don't see any reason to not inherit all environment variable from the updatecli process, so I would retrieve all of them from `os.Environment` then append `DRY_RUN`

Also I am wondering if we shouldn't replace `DRY_RUN` by `UPDATECLI_DRY_RUN` to reduce the risk of collision

olblak:

# Allow to set environment variable within shell resources

Fix #749

Because target shell resources, set the environment variable `DRY_RUN`, all other environment variable are hidden with the shell script.

Note that by default all environment variables are available for source and condition shell resource.

This affects multiple project relying on Github Action that update the PATH env variable with cached tool

## Test

```shell
cp <to_package_directory>
go test
```


olblak:
Manifest example

```
name: Shell example using environment variable

sources:
  default:
    name: Test source shell
    kind: shell
    spec:
      command: ./examples/scripts.d/env.sh

  default2:
    name: Test source shell
    kind: shell
    spec:
      command: ./examples/scripts.d/env.sh
      environments:
        - name: PATH
          inherit: true

conditions:
  default:
    name: Test source shell
    disablesourceinput: true
    kind: shell
    spec:
      command: ./examples/scripts.d/env.sh

  default2:
    name: Test source shell
    kind: shell
    disablesourceinput: true
    spec:
      command: ./examples/scripts.d/env.sh
      environments:
        - name: PATH
          inherit: true

targets:
  default:
    name: Test source shell
    disablesourceinput: true
    kind: shell
    spec:
      command: ./examples/scripts.d/env.sh

  default2:
    name: Test source shell
    kind: shell
    disablesourceinput: true
    spec:
      command: ./examples/scripts.d/env.sh
      environments:
        - name: PATH
          inherit: true

```

olblak:
And the script example
```
#!/bin/sh 

echo "PATH: $PATH"
echo "DRY_RUN: $DRY_RUN"

```

pkg/plugins/resources/shell/environments.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Environments []Environment
+
+func (e *Environments) ToStringArray() []string {
```
jlevesy, line 12:
```suggestion
func (e Environments) ToStringArray() []string {
```

Slices are already references, you don't need to use a pointer to manipulate them without copy

jlevesy, line 12:
```suggestion
func (e *Environments) ToStringSlice() []string {
```

Nit: Arrays in go are fixed length  arrays, here you're returning a slice of string, so this method should be called `ToStringSlice`.

pkg/plugins/resources/shell/environments.go
```diff
@@ -0,0 +1,70 @@
...
+	result := []string{}
+
+	for _, environment := range *e {
+		result = append(result, environment.String())
+	}
```
jlevesy, line 17:
```suggestion
	result := make([]string, len(*e))

	for i, environment := range *e {
		result[i] = environment.String()
	}
```

Nit: you can pre allocate memory when doing this kind of operation, this saves some costly array copies when calling `append` on a full array. That's performance optimization thought.

pkg/plugins/resources/shell/environments.go
```diff
@@ -0,0 +1,70 @@
...
+		foundName[env.Name] = env.Value
+	}
+
+	if len(foundDuplicatedName) > 0 {
+		logrus.Warningf("duplicated environment variable found: [%q]\n", strings.Join(foundDuplicatedName, ","))
+		return true
+	}
+
+	return false
+
```
jlevesy, line 41:
```suggestion
	foundName := map[string]struct{}{}
	foundDuplicatedName := []string{}

	for _, env := range *e {
		if _, ok := foundName[env.Name]; ok {
			foundDuplicatedName = append(foundDuplicatedName, env.Name)
			continue
		}
		
		foundName[env.Name] = struct{}{}
	}

	if len(foundDuplicatedName) > 0 {
		logrus.Warningf("duplicated environment variable found: [%q]\n", strings.Join(foundDuplicatedName, ","))
		return true
	}

	return false
```

Nit: you can use `map[string]struct{}` when you don't care about the value of a map. This also saves memory.

jlevesy (commented):
👀 

Most of my comments are style nits so feel free to ignore them ^^.

Let me know this this helps or not!

olblak:
> Most of my comments are style nits so feel free to ignore them ^^.

Thanks, I am always open for some learning :)

pkg/plugins/resources/shell/environments.go
olblak: Good point, thanks for the advice, I'll try to remember next time

olblak: That's a very good suggestion, thank you very much

olblak:
I now realize that I didn't address a weird behavior:

* By default `source` and `condition` inehrit from all environment variable from the updatecli process
* By default `target` has almost no variable ["PWD","DRY_RUN","SHLVL"]

@dduportal Do you have any opinions on which behavior should be the default for all stages

dduportal:
> I now realize that I didn't address a weird behavior:
> 
>     * By default `source` and `condition` inehrit from all environment variable from the updatecli process
> 
>     * By default `target` has almost no variable ["PWD","DRY_RUN","SHLVL"]
> 
> 
> @dduportal Do you have any opinions on which behavior should be the default for all stages

I can't remember why I differentiate these behaviors. I fear that I might have been lazy for the target.

Is it ok to keep the behavior that you introduced in this PR also for target (which adding the `DRY_RUN` variable in the list of env to be passed, overriding the value with a warning if it already exists)

dduportal:
> Manifest example
> 
> ```
> name: Shell example using environment variable
> 
> sources:
>   default:
>     name: Test source shell
>     kind: shell
>     spec:
>       command: ./examples/scripts.d/env.sh
> 
>   default2:
>     name: Test source shell
>     kind: shell
>     spec:
>       command: ./examples/scripts.d/env.sh
>       environments:
>         - name: PATH
>           inherit: true
> 
> conditions:
>   default:
>     name: Test source shell
>     disablesourceinput: true
>     kind: shell
>     spec:
>       command: ./examples/scripts.d/env.sh
> 
>   default2:
>     name: Test source shell
>     kind: shell
>     disablesourceinput: true
>     spec:
>       command: ./examples/scripts.d/env.sh
>       environments:
>         - name: PATH
>           inherit: true
> 
> targets:
>   default:
>     name: Test source shell
>     disablesourceinput: true
>     kind: shell
>     spec:
>       command: ./examples/scripts.d/env.sh
> 
>   default2:
>     name: Test source shell
>     kind: shell
>     disablesourceinput: true
>     spec:
>       command: ./examples/scripts.d/env.sh
>       environments:
>         - name: PATH
>           inherit: true
> ```

Looks good!

WDYT to use the same configuration behavior as Docker?

```
environments:
  - name: PATH # Only name, eg. 'if value == ""' means "inherit"
  - name: GITHUB_TOKEN
     value: "SuperSecretTokenExposed" # Custom value that could be templatized from within updatecli
 ```


Also, I wonder if the attribute `name:` could not be `key:`  or `variable:` ? (but nitpicking for this one)
 
 ?

**Comment by olblak** [comment]
> I can't remember why I differentiate these behaviors. I fear that I might have been lazy for the target.
Is it ok to keep the behavior that you introduced in this PR also for target (which adding the DRY_RUN variable in the list of env to be passed, overriding the value with a warning if it already exists)

I guess you mean the source/condition :D I agree it doesn't make sense to us the `DRY_RUN` env variable for source and condition but if we don't do it then by default then the default behavior in the "os" golang package is to inherit all environment variable from the updatecli process. So I am gonna add "DRY_RUN" to source and condition as well  

> Also, I wonder if the attribute name: could not be key: or variable: ? (but nitpicking for this one)

TBH I hesitated implementing the `inherit` key, and I agree with you it doesn't really add value, so I am gonna remove it
Regarding variable `key/variable` instead of `name/value`, It's a personal taste but I prefer the `name/value` so unless addition input, I am gonna keep it :D

File: pkg/plugins/resources/shell/environment.go [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type Environment struct {
+	// Name defines the environment variable name
+	Name string `yaml:",omitempty"`
+	// Value defines the environment variable value
+	Value string `yaml:",omitempty"`
+}
+
+func (e *Environment) String() string {
```
dduportal, line 22:
```suggestion
func (e Environment) String() string {
```


This method does not change the receiver

pkg/plugins/resources/shell/environments.go
```diff
@@ -0,0 +1,70 @@
...
+	if len(foundDuplicatedName) > 0 {
+		logrus.Warningf("duplicated environment variable found: [%q]\n", strings.Join(foundDuplicatedName, ","))
+		return true
+	}
+
+	return false
+}
+
+// Ensures that we don't have duplicated value for a variable and that the user is not attempting to override the DRY_RUN reserved variable.
+func (e *Environments) Validate() error {
```
dduportal, line 45:
```suggestion
func (e Environments) Validate() error {
```

Same as above: a `Validate` method which is able to modify the receiever is really confusing

pkg/plugins/resources/shell/command.go
```diff
@@ -37,6 +39,12 @@ func (nce *nativeCommandExecutor) ExecuteCommand(inputCmd command) (commandResul
 	command.Env = append(command.Env, inputCmd.Env...)
 	err := command.Run()
 
+	// Display environment variable in debug mode
```
lemeurherve, line 42:
```suggestion
	// Display environment variables in debug mode
```

pkg/plugins/resources/shell/environment.go
```diff
@@ -0,0 +1,45 @@
...
+	return fmt.Sprintf("%s=%s", e.Name, e.Value)
+}
+
+func (e *Environment) Validate() error {
+	gotErr := false
+
+	// If a environment variable name specified without value
```
lemeurherve, line 29:
```suggestion
	// If an environment variable name is specified without value
```

pkg/plugins/resources/shell/environment.go
```diff
@@ -0,0 +1,45 @@
...
+	// then inherit the value from Updatecli environment
```
lemeurherve, line 30:
```suggestion
	// then it inherits the value from Updatecli environment
```

pkg/plugins/resources/shell/environments.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Environments []Environment
+
+func (e Environments) ToStringArray() []string {
```
lemeurherve, line 12: I liked the suggestion of `ToStringSlice()` here

pkg/plugins/resources/shell/main.go
```diff
@@ -12,6 +12,8 @@ import (
 type Spec struct {
 	// Specifies the shell command
 	Command string `yaml:",omitempty" jsonschema:"required"`
+	// Environments allows to pass environment variable to the shell script
```
lemeurherve, line 15:
```suggestion
	// Environments allows to pass environment variable(s) to the shell script
```

lemeurherve (approved):
Nit only :)

pkg/plugins/resources/shell/environment.go
```diff
@@ -0,0 +1,45 @@
...
+	Name string `yaml:",omitempty"`
+	// Value defines the environment variable value
+	Value string `yaml:",omitempty"`
+}
+
+func (e *Environment) String() string {
+	return fmt.Sprintf("%s=%s", e.Name, e.Value)
+}
+
+func (e *Environment) Validate() error {
```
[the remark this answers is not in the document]
olblak: I used "Update" which seems more appropriate to me, as we retrieve the environment variable value if needed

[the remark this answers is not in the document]
dduportal: I agree that `Update()` seems more appropriate. However I remember that it is also used for the template processing in updatecli manfiest, which "updates" the `Spec` but should not `Update` the Config objects. WDYT?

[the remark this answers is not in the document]
jlevesy: If you need a replacement name, how about `Read` or `Load`?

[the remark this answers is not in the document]
lemeurherve: My vote would go to `Load`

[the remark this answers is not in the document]
olblak: > If you need a replacement name, how about Read or Load?
I like `load` very much, let me change that

olblak:
After the final evaluation, I would like to trigger a new release, as this PR is blocking several of my pipelines

dduportal (approved):

olblak:
Thanks everybody for the constructive reviews

jetersen:
oh damn, could really use a release of this 😅 

Thought this was just supported.

jetersen:
This is massively missing with variables in pwsh cannot even access $ENV:TEMP

jetersen:
Workaround :( https://github.com/jenkinsci/bom/pull/1290/commits/b359dd5557fa43635327ae393027356a6daaaec4#diff-f3b39a28e1d2b517f4b0d950ceb91c4eadd3d9cf4e8727bb3f2bf1cf9a68792dR127

lemeurherve:
@jetersen  FTR, it's now available in the v0.28.0 release

jetersen:
Thanks for reminding me that I need to setup releases watch on this repo.

Issue #716, opened 2022-06-12
Feature Request: automatic SCM configuration from local environment


dduportal:
### Is your feature request related to a problem?

Ref. https://github.com/updatecli/updatecli/issues/465#issuecomment-1147234388

As an updatecli user, most of my manifests are using the SCM configuration from the environment where it is running.
I would want to avoid having to specify any `scms` configuration at top-level as it could be automatically configured.

As an updatecli developer, I need updatecli to be able to differentiate "current" SCM environment (is it the SCM of the current directory where updatecli is executed?) from external SCMs when running a target or a pull request task for https://github.com/updatecli/updatecli/issues/465.

### Solution you'd like

- A special SCM identified as `local`, that I could specify from my targets/pullrequests, that would autoconfigure itself from the directory where updatecli is executed.
  - An informational message should be written when an autoguessed SCM is used

- User could opt-out of the "autoguessed configuration" by specifying `scms.local.disabled` to `true`
  - `disabled` is `false` by default
  - The informational message should specify this opt-out option to help users

- Manifests without any SCM at all (we have some, at least, in the `e2e/` directory for deprecated cases) would break unless they opt-out (see bullet below). 

- User should be able to complement/override the autoguessed configuration for `local` by specifying its own values, allowing control of the autoguess behavior so that these values are never guessed:
  - By specifying environment variables with the naming pattern `UPDATECLI_SCM_LOCAL_XXX`, where `XXX` (uppercase) can be any of the possible guessed values
  - By specifying custom configuration in the manifest through `scms.local.spec.xxx`,  where `xxx` (lower case) can be any of the possible guessed values

### Alternatives you've considered

- Different naming: As discussed in https://github.com/updatecli/updatecli/issues/465, we could use `current`, `auto`, `manual`

- Opt-out by default: opt-in by default (e.g. having to specify `scm.local.enable: true` would require a bunch of configuration to be written as it is today: I don't see the point of the feature in this case.
  - The amount of "broken" manifest is clearly low
  - There is a message to opt-out so user know what to do when it breaks

dduportal:
List of the autoguessed atrributes:

For`git` :
- [x] URL - https://github.com/updatecli/updatecli/pull/713

For `github`:
- [x] Repository name - https://github.com/updatecli/updatecli/pull/713
- [x] Owner - https://github.com/updatecli/updatecli/pull/713

PR #763, opened 2022-07-14
Automate Helm Chart Update

### LINKED ISSUE #761: Feature Request: Automate Helm Chart Update

Up to now we used Updatecli to define tailored update scenarios. I am now looking to automatically detected those common scenarios. 
One of them is:

1. Search recursively all Chart.yaml in a specific directory/git repository
2. Identify all chart dependencies 
3. Generate appropriate Updatecli manifest and then run them right away 

```
name: <chart_name>-<dependency_name>
sources:
    <dependency_name>:
        name: Get latest "<dependency_name>" Helm Chart Version
        kind: helmchart
        spec:
            name: <dependency_name>
            url: <dependency_url>
conditions:
    <dependency_name>:
        name: Ensure dependency "<dependency_name>" is specified
        kind: yaml
        spec:
            file: .chart/<chart_name>/Chart.yaml
            key: dependencies[2].name
            value: <dependency_name>
        disablesourceinput: true
targets:
    <dependency_name>:
        name: Bump chart dependency "<dependency_name>" in Chart "<chart_name>"
        kind: helmchart
        spec:
            file: Chart.yaml
            key: dependencies[2].version
            name: chart/<dependency_name>
            versionincrement: minor
        sourceid: <dependency_name>
```

### Solution you'd like

I would like to be able to run the following command 

```
# Updatecli should be able to identify helm chart 
updatecli diff

#Updatecli should be able to apply helm chart update
updatecli apply
```

The flag `--disable-auto` would disable the `auto-guessing` feature

_No response_

### Anything else?

For a more global discussion https://github.com/updatecli/updatecli/discussions/755
The auto feature would also greatly benefit from https://github.com/updatecli/updatecli/issues/716

olblak:
# Automate Helm Chart Update

Fix #761 

- [x] Without manifest
- [x] With manifest
- [x] SCM Support
- [x] Pullrequest support   
- [x] [Documentation](https://github.com/updatecli/website/pull/428) 
- [x] Dockerimage should be using semver
- [x] Values.yaml must be in the same directory than the chart.yaml

This pullrequest is an attempt to automatically generate Updatecli manifest based on a local directory.
I envision two different scenarios, with and without manifest

## Without

The default scenario, if we run `updatecli diff` or `updatecli apply`, by default we try to detect available updates from the local directory. Until https://github.com/updatecli/updatecli/issues/716 is done, the default scenario won't support scm

## With manifest

This scenario relies on a new configuration, to customize the way we build manifest.
We use this configuration to provide scm configuration, or for example restrict generated manifest based on rules.

We can run one of the following command

* `updatecli manifest show --config example.yaml`
* `updatecli diff --config example.yaml`
* `updatecli apply --config example.yaml`

.example.yaml
```
# **Optional** Provide generic scm configuration used to generate and/or apply Updatecli manifest
# If an scm configuration is provided then we use that repository to automatically generate manifest 
scms:
  test:
    kind: github
    spec:
      owner: epinio
      repository: helm-charts
      token: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
      branch: main

# Provide generic pullrequest configuraiton used to generate and/or apply Updatecli manifest
pullrequests:
  test:
    kind: github
    scmid: test
    #targets:
    #  - default
    #scmid: default # May conflicts with current code
    #targets:
    #  - goWorkflow # May conflicts with current code
    spec:
      labels:
        - chore
        - skip-changelog

# Provide autodiscovery configuration such as the scm or pullrequest config to use
autodiscovery:
  scmid: test
  pullrequestid: test
  crawlers:
    helm:
      #ignore:
      #  # - path: chart/epinio-ui/*
      #  # - path: ~/Projects/Epinio/helm-charts/chart/epinio/*
      #  # - path: chart/epinio/*
      #only:
      #  # - path: ~/Projects/Epinio/helm-charts/chart/epinio/*
      #  - path: chart/epinio/*

```

## First plugin

As a first working scenario, I am using the Helm chart dependency that needs to be updated, but I am considering to use the same approach than for source/condition/target, where we could add more plugins.

For better visibility, I am planning to put this new plugin type in `pkg/plugins/generics`

So for a specific directory, I am looking for every Chart.yaml or Chart.yml that have dependencies listed. Then I return one manifest per dependency identified. This means that once merged, we would have a new way to specifies Updatecli manifest 

1. Using  yaml configuration
4. Generating them from a directory

## Test

```shell
go build -o bin/updatecli . 
# Run from a directory with helm chart.
./bin/updatecli diff --config dumb.yaml
```

### Potential improvement

When we run updatecli `diff` or `apply` we set the default configuration to "updatecli.yaml" and we trigger an error if no configuration file is available. This is problematic with the builder because we are now in a situation where we may not need any configuration files.

Solution A: We don't trigger an error if "updatecli.yaml" is not available
Solution B: We require a new flag such as `updatecli diff --auto`

I would be more in favor of not triggering an error if `updatecli.yaml is empty or doesn't exist`

PS: Please note that the naming of package, struct, etc. is not definitive and evolve with my understanding of the current problem

another potential improvement is to run `updatecli show` to display detected manifest. Considering that `updatecli show` don't clone scm repositories, it won't work with remote repositories


olblak:
For the same of this feature, I am hesitating between `auto`, `generic`, `builder` which explain why those three names are used in the pullrequest.

olblak:
Please also note that this PR doesn't generate "pipelines" but "pipeline manifest"

olblak:
The auto-discovery feature introduced in this pullrequest , means it should be possible to run updatecli commands without providing any manifest as `updatecli diff` could discovery available pipeline.
But due to some early design decision, by default if no manifests are provided then we fallback to "updatecli.yaml" and fail if it doesn't exist.

This commit, https://github.com/updatecli/updatecli/pull/763/commits/11e468c5fd70aaf540d136884c4b74a9144471a5 refactor a bit so we first detect `updatecli.yaml`, to remain backward compatible, then `updatecli.d` if it's a directory, and finally we just try to autodiscover available pipelines

olblak:
I think it would be better to split this pullrequest in two. One about the autodiscovery and a second one about helm chart dependency update

olblak:
Here is the documentation that I started on the autodiscovery feature https://github.com/updatecli/website/pull/428

olblak:
After additional experimenting, I am starting to wonder when to enable the "local" discovery.
What I mean by "local" autodiscovery is when updatecli tries to identify potential pipeline based on local files.

At the moment it can be done in two cases 

Case 1:
Run `updatecli manifest show`, at this time of writing it would show 6 manifests

Case 2: 
Run `updatecli manifest show --config example.yaml` with the following file 

example.yaml
```
autodiscovery:
  crawlers:
    helm:
      auths:
        # Specifies dockerhub credentials to use if docker image are identified in a helm chart
        docker.io:
          token: {{ requiredEnv "DOCKERHUB_TOKEN" }}
```

The second case,  means we would identify two times the manifest one with the local and the second time with the custom configuration...
So we can run `updatecli manifest show --config example.yaml --disable-local-autodiscovery`

I really struggle to identify how to detect if the local autodiscovery should be execute based on provide configuration

So I am wondering if it wouldn't be better to inverse the logic by having `updatecli manifest show --enable-local-autodiscovery`

olblak:
I added manifest examples in `e2e/updatecli.d/warning.d/autodiscovery`

olblak:
so I found the local autodiscovery quite annoying by default, so I disable it by default

olblak:
I am just dropping this repository https://github.com/rancher/fleet-examples 
To remind myself to use it in my test

PR #798, opened 2022-08-08
Implement the source "DockerImage"

### LINKED ISSUE #512: Feature Request: Implement the source "DockerImage"

I am facing a situation where the source of truth of a version is defined on a docker registry.

I am looking for a way to retrieve the latest docker image tag matching a version such as semantic versioning or regex.
At the moment we have the condition "dockerImage" which allows to test if a docker image tag exist on a docker repository.
Naturally I would use the same resource kind

### Solution you'd like

I would like to reuse the same approach than for the dockerImage condition but with the versionFilter as well to retrieve the latest docker image tag matching a version pattern.
But I am 

**Default**: Retrieve the newest docker image tag
```
    name: "Docker Image Published on Registry"
    kind: dockerImage
    spec:
      image: "updatecli/updatecli"
```

**Semver**: Retrieve the latest version matching pattern 
```
    name: "Docker Image Published on Registry"
    kind: dockerImage
    spec:
      image: "updatecli/updatecli"
      versionFilter:
          kind: semver
          pattern: "~0"
```
**Regex**: Retrieve the latest version matching regex pattern
```
    name: "Docker Image Published on Registry"
    kind: dockerImage
    spec:
      image: "updatecli/updatecli"
      versionFilter:
          kind: regex
          pattern: "0.(\d*)$"
```

I have been wondering using a different resource kind than "dockerImage"

### Anything else?

_No response_

olblak:
Fix #512

Thanks @dduportal for the work you did on the dockerregistry support, it greatly simplified this pullrequest \o/
Especially the UT part :)
 
This pullrequest add "source" support to the dockerimage plugin.

## Test

```shell
# Test a manifest
go build -o bin/updatecli && time ./bin/updatecli diff --config e2e/updatecli.d/success.d/source/dockerimage.yaml

# Test the dockerregistry UT
cd pkg/plugins/utils/docker/dockerregistry
go test .
```

## Additional Information

The "latest" dockerimage means different things depending the registry. For example DockerHub returns a list of images alphabetically order, while ghcr.io return a list by creation time. In both cases the API only return a list of image tag. 
This means that the only reliable way of sorting images is by using semantic versioning assuming that the image implement it.

### Tradeoff

The API used to list all tags for a specific image doesn't return tag metadata such as architecture supported . We need another API call to retrieve those metadata and this for each tag. The consequence of this is that we may exhaust our API limit very quickly.      

### Potential improvement

In the future, we may implement caching for source "dockerimage"


pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
 	"github.com/updatecli/updatecli/pkg/plugins/utils/docker/dockerimage"
 	"github.com/updatecli/updatecli/pkg/plugins/utils/docker/dockerregistry"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
 // Spec defines a specification for a "dockerimage" resource
 // parsed from an updatecli manifest file
 type Spec struct {
-	// Architecture specifies the container image architecture such as `amd64`
+	// [S][C][T]Architecture specifies the container image architecture such as `amd64`
```
hervelemeur, line 16: Shouldn't a space be added between the last `]` and the first comment word on these lines?

pkg/plugins/resources/dockerimage/source.go
```diff
@@ -2,8 +2,92 @@ package dockerimage
...
+
+	}
+
+	if len(tag) == 0 {
+		logrus.Infof("%s No Docker Image Tag found matching pattern %q", result.FAILURE, di.versionFilter.Pattern)
+		return tag, fmt.Errorf("no Docker Image Tag tag found matching pattern %q", di.versionFilter.Pattern)
+	} else if len(tag) > 0 {
+		logrus.Infof("%s Docker Image Tag %q found matching pattern %q", result.SUCCESS, tag, di.versionFilter.Pattern)
+	} else {
+		logrus.Errorf("Something unexpected happened in Github source")
```
hervelemeur, line 71:
```suggestion
		logrus.Errorf("Something unexpected happened in GitHub source")
```
(Nit)

pkg/plugins/utils/docker/dockerregistry/tags.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/docker/dockerimage"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/link"
+)
+
+// Tags retrieves the all tags of the provided docker image from the registry
```
hervelemeur, line 14:
```suggestion
// Tags retrieves all tags of the provided docker image from the registry
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
+	// [S]VersionFilter provides parameters to specify version pattern and its type like regex, semver, or just latest.
+	VersionFilter version.Filter `yaml:",omitempty"`
 }
 
 // DockerImage defines a resource of type "dockerimage"
 type DockerImage struct {
 	spec     Spec
 	registry dockerregistry.Registry
 	image    dockerimage.Image
+	// versionFilter holds the "valid" version.filter, that might be different than the user-specified filter (Spec.VersionFilter)
```
hervelemeur, line 37: In which case(s) are these values different?

pkg/plugins/resources/dockerimage/main.go
olblak: Good point

olblak, line 16:
```suggestion
	// [S][C][T] Architecture specifies the container image architecture such as `amd64`
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
 )
 
 // Spec defines a specification for a "dockerimage" resource
 // parsed from an updatecli manifest file
 type Spec struct {
-	// Architecture specifies the container image architecture such as `amd64`
+	// [S][C][T]Architecture specifies the container image architecture such as `amd64`
 	Architecture string `yaml:",omitempty"`
-	// Image specifies the container image such as `updatecli/updatecli`
+	// [S][C][T]Image specifies the container image such as `updatecli/updatecli`
```
olblak, line 18:
```suggestion
	// [S][C][T] Image specifies the container image such as `updatecli/updatecli`
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
 // parsed from an updatecli manifest file
 type Spec struct {
-	// Architecture specifies the container image architecture such as `amd64`
+	// [S][C][T] Architecture specifies the container image architecture such as `amd64`
 	Architecture string `yaml:",omitempty"`
-	// Image specifies the container image such as `updatecli/updatecli`
+	// [S][C][T]Image specifies the container image such as `updatecli/updatecli`
 	Image string `yaml:",omitempty"`
-	// Tag specifies the container image tag such as `latest`
+	// [C][T]Tag specifies the container image tag such as `latest`
```
olblak, line 20:
```suggestion
	// [C][T] Tag specifies the container image tag such as `latest`
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
 	Tag string `yaml:",omitempty"`
-	// Username specifies the container registry username to use for authentication. Not compatible with token
+	// [S][C][T]Username specifies the container registry username to use for authentication. Not compatible with token
```
olblak, line 22:
```suggestion
	// [S][C][T] Username specifies the container registry username to use for authentication. Not compatible with token
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
 	Username string `yaml:",omitempty"`
-	// Password specifies the container registry password to use for authentication. Not compatible with token
+	// [S][C][T]Password specifies the container registry password to use for authentication. Not compatible with token
 	Password string `yaml:",omitempty"`
-	// Token specifies the container registry token to use for authentication. Not compatible with username/password
+	// [S][C][T]Token specifies the container registry token to use for authentication. Not compatible with username/password
```
olblak, line 26:
```suggestion
	// [S][C][T] Token specifies the container registry token to use for authentication. Not compatible with username/password
```

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -7,30 +7,36 @@ import (
...
-	// Username specifies the container registry username to use for authentication. Not compatible with token
+	// [S][C][T] Username specifies the container registry username to use for authentication. Not compatible with token
 	Username string `yaml:",omitempty"`
-	// Password specifies the container registry password to use for authentication. Not compatible with token
+	// [S][C][T] Password specifies the container registry password to use for authentication. Not compatible with token
 	Password string `yaml:",omitempty"`
-	// Token specifies the container registry token to use for authentication. Not compatible with username/password
+	// [S][C][T] Token specifies the container registry token to use for authentication. Not compatible with username/password
 	Token string `yaml:",omitempty"`
+	// [S]VersionFilter provides parameters to specify version pattern and its type like regex, semver, or just latest.
```
olblak, line 28:
```suggestion
	// [S] VersionFilter provides parameters to specify version pattern and its type like regex, semver, or just latest.
```

pkg/plugins/resources/dockerimage/main.go
olblak: Due to Updatecli manifest templating, we avoid modifying spec data as it leads to unexpected data. So if we need to modify the object for some reason then we copy and work on the copied one.

olblak:
Considering that dockerhub and ghcr.io do not return tags in the same order, ghrc sort by creation time while dockerhub sort by alphabetical order. The "latest" tag doesn't mean anything in this plugin so I am wondering if by default I should use soft semantic versioning

@hervelemeur  @dduportal  do you have any opinion?


Also while being in the scope of this PR, we could also provide an option to enforce "strict" semver

pkg/plugins/resources/dockerimage/main.go
dduportal: As a rule of thumb:
- Spec is a data specified by the end user through YAML (and eventually something else). It can be refreshed at any time and templatized.
- "Config" (naming is not always the same) is the transient data (in memory) with default configuration (default values, "Spec" converted to a custom model, and contextual data for dependency injection such as clients objects, etc.)

dduportal:
> Considering that dockerhub and ghcr.io do not return tags in the same order, ghrc sort by creation time while dockerhub sort by alphabetical order. The "latest" tag doesn't mean anything in this plugin so I am wondering if by default I should use soft semantic versioning
> 
> @hervelemeur @dduportal do you have any opinion?
> 
> Also while being in the scope of this PR, we could also provide an option to enforce "strict" semver

Could you explain with use cases (and associated YAML manifests)? 

I think I understand but I want to avoid any confusion (I don't want you having to change code once merged :D ).
Also I don't really understand the use case for "latest" outside retrieveing the full image name?

At first glance (.e.g without thinking too much) I would keep the Docker image pattern:

- If no tag specified, then assume it is `:latest`
- If the tag is `latest` (explicit tag or no tag at all), then retrieve the full image name including the `:latest` tag (low value for use case)
- If semver filter, then retrieve the most "recent" semver (based on semver sorting rules)
- If regex filter, then apply filtering, sort the result alphabetically (lexicographic?) in memory and get the last one.

=> If we need more advanced filtering/sorting, it should be a separated feature scoped to `versionFilter`

olblak:
> > Could you explain with use cases (and associated YAML manifests)?

> I think I understand but I want to avoid any confusion (I don't want you having to change code once merged :D ).

To be honest, after some additional thought I am not sure see to a usecase in the current context..

>  Also I don't really understand the use case for "latest" outside retrieveing the full image name?
 
Me neither :p I can't think of a good usage of "latest" :/
semver and regex are the only one that interest me...

> => If we need more advanced filtering/sorting, it should be a separated feature scoped to versionFilter

I totally agree, I am just wondering if by default we should use semver or regex versionFilter, as the "latest" is pointless

olblak:
The PR is ready as is and we can still improve it later once we really start using it :)

PR #804, opened 2022-08-12
Update jsonschema library to 0.6.0

olblak:
Superseed https://github.com/updatecli/updatecli/pull/799

The jsonschema library used in Updatecli removed YAML tag support to focus on JSON tag. As far as I could see it doesn't impact updatecli because we didn't leverage YAML tag but it's hard to say for sure. The generated jsonschema is pretty big, 

## Test

/


dduportal (approved):
👍

Issue #803, opened 2022-08-12
Semantic Version filter drop "v" prefix


olblak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When parsing version using semantic version, updatecli drop the v prefix such as 

```
######################
# BUMP VENOM VERSION #
######################


SOURCES
=======

latestVersion
-------------
Searching for version matching pattern "*"
✔ Github Release version "1.0.1" found matching pattern "*"
```

### Expected Behavior

It should return 

```
######################
# BUMP VENOM VERSION #
######################


SOURCES
=======

latestVersion
-------------
Searching for version matching pattern "*"
✔ Github Release version "v1.0.1" found matching pattern "*"

```

### Steps To Reproduce

`updatecli diff --confeig updatecli/updatecli.d/venom.yaml `

### Environment

```markdown
- OS: linux
- updatecli: v0.29.0
```

hervelemeur:
Note, we're using semver filter specially for dripping this "v" in some cases

dduportal:
> ### Is there an existing issue for this?
> 
>     * [x]  I have searched the existing issues
> 
> 
> ### Current Behavior
> 
> When parsing version using semantic version, updatecli drop the v prefix such as
> 
> ```
> ######################
> # BUMP VENOM VERSION #
> ######################
> 
> 
> SOURCES
> =======
> 
> latestVersion
> -------------
> Searching for version matching pattern "*"
> ✔ Github Release version "1.0.1" found matching pattern "*"
> ```
> 
> ### Expected Behavior
> 
> It should return
> 
> ```
> ######################
> # BUMP VENOM VERSION #
> ######################
> 
> 
> SOURCES
> =======
> 
> latestVersion
> -------------
> Searching for version matching pattern "*"
> ✔ Github Release version "v1.0.1" found matching pattern "*"
> ```
> 
> ### Steps To Reproduce
> 
> `updatecli diff --confeig updatecli/updatecli.d/venom.yaml `
> ### Environment
> 
> ```gfm
> - OS: linux
> - updatecli: v0.29.0
> ```
> 
> ### Anything else?
> 
> _No response_

I've always thought it was expected ?

olblak:
It wasn't initially intended and to be honest I thought it was more a technical limitation of the semver library that we are using.
I am open to suggestion here

olblak:
> Note, we're using semver filter specially for dropping this "v" in some cases

[...]

PR #813, opened 2022-08-21
Feat: add an experimental flag and use it to define the versionFilter behavior

dduportal:
This PR introduces a new flag `--experimental` for all subcommands, which sets a global boolean variable `experimental` in a new package `cmdoptions`.

It also reverts indirectly #802 but allows the new behavior (e.g. returning the original version) for #803 if the `--experimental` flag is used.

The goal is to avoid a breaking change while allowing the new behavior to be used for testing autodiscovery. This feature flag would allows us to ship feature without enabling it.

## Test

```shell
cp <to_package_directory>
go test
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/cmdoptions"
 	"github.com/updatecli/updatecli/pkg/core/log"
 
 	"github.com/updatecli/updatecli/pkg/core/engine"
@@ -18,6 +19,7 @@ var (
 	secretsFiles []string
 	e            engine.Engine
 	verbose      bool
+	experimental bool
 
 	rootCmd = &cobra.Command{
 		Use:   "updatecli",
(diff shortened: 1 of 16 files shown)


olblak (approved):
Tested and it works great

olblak:
I guess before merging this PR we need a new issue to remind us to move this from experimental to GA then assign that issue to a futur milestone such as 0.32.0

olblak:
Also I would add this documentation to the coming release note:

If the Github Release you are interacting with, prepend retrieved versions with the character "v" such as "v1.0.0",
then please read this message carefully.

In Updatecli, we consider that information retrieved from a "Source" shouldn't be altered.
Instead, it should be the role of a "Transformer" rule to modify it.

The Github Release version retrieve using the semantic version rule will stop dropping the "v" if one is specified in the GitHub release tag such as v1.0.0.
This bug has been around for so long that we saw a significant amount of Updatecli manifest written in a way that we can't automatically fix by running "updatecli manifest upgrade"
 
 As we take seriously backward compatibility for Updatecli manifest, we introduce the parameter "--experimental" to allows us to test such new behaviors before making them general availability.
 
 Please be ready to adapt your manifest to this new behavior.
 
 Before:
 ```
  sources:
      example:
          name: Get latest Updatecli release
          kind: githubrelease
          spec:
              owner: updatecli
              repository: updatecli
              username: '{{ requiredEnv "GITHUB_ACTOR" }}'
              versionfilter:
                  kind: semver
 ```

 After:
 ```
  sources:
      example:
          name: Get latest Updatecli release
          kind: githubrelease
          spec:
              owner: updatecli
              repository: updatecli
              username: '{{ requiredEnv "GITHUB_ACTOR" }}'
              keeporiginalversion: true
              versionfilter:
                  kind: semver
              transformers:
                  - trimprefix: v
 ```

dduportal:
Wdyt to split the PR here?

Current PR should only introduce the flag.
New pr should use the flat for the « version » behavior (and would add the explanation your shared library nd the warning message).

rationale is that the pr was initially a proof of concept

olblak:
This sounds good to me

dduportal:
> This sounds good to me

ok, then let me split the PR into 2 news then (would be clearer for readers)

dduportal:
Introduced the new flag alone: https://github.com/updatecli/updatecli/pull/815

dduportal:
Aaaand https://github.com/updatecli/updatecli/pull/816 for the "drop prefix with warning"

PR #838, opened 2022-08-31
pullrequest correctly search linked targets based on pullrequestID

olblak:
Signed-off-by: Olblak <me@olblak.com>

Fix #836 
In release 0.30.0, Updatecli automatically detects which targets are related to a pullrequest spec, for generating the pullrequest body. 

So 
```
pullrequests:
  jenkins:
    title: Bump Jenkins Version to {{ source "jenkins" }}
    kind: github
    scmid: github2
    targets:
      - jenkins
```

becomes 
```
pullrequests:
  jenkins:
    title: Bump Jenkins Version to {{ source "jenkins" }}
    kind: github
    scmid: github2
```

But in the code I passed the wrong id value the pullrequest scmID instead of the pullrequets ID
I didn't detect this issue because in my tests environment, it's usually the same value.

## Test

To test this pull request, you can run the following commands:

```shell
cp <to_package_directory>
go test
```

PR #839, opened 2022-09-01
Fix: yaml source workdir relying on scm

olblak:

Spot while testing https://github.com/updatecli/updatecli/pull/838
This appears to be a regression where the source yaml didn't use the correct scmID

## Test

```shell
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/githubPullrequest.yaml
```

### Tradeoff

/

### Potential improvement

This part should deserve better UT, but I don't really have the time atm

### DIFF
--- a/e2e/updatecli.d/success.d/githubPullrequest.yaml
+++ b/e2e/updatecli.d/success.d/githubPullrequest.yaml
@@ -13,13 +13,16 @@ scms:
 sources:
   jenkins:
     name: Get updatecli config title
-    kind: jenkins
+    kind: yaml
     scmid: default
+    spec:
+      file: "updatecli/updatecli.d/updatecli.yaml"
+      key: "title"
 
 targets:
-  jenkins:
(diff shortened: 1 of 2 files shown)


lemeurherve:
Oh! I'm wondering if it's not the same type of bug I've got in my pending PR...

olblak:
> Oh! I'm wondering if it's not the same type of bug I've got in my pending PR...

Which one? I am curious about this bug because I am pretty sure we use this workflow but can't remember where

lemeurherve:
https://github.com/updatecli/updatecli/pull/650

I'm struggling to debug e2e tests, but I think the errors I'm encountering could be related to this workdir.

olblak:
well try.gitea.io seems to be dead.
I'll switch to [codeberg](https://codeberg.org/) when I have some time but for now I am gonna merge this PR and trigger a release

PR #853, opened 2022-09-08
Bump to Golangci 1.49 and fix linter issues

olblak:
Supersede https://github.com/updatecli/updatecli/pull/848
And fix new errors


olblak:
Force merge as test won't pass anyway due to https://github.com/updatecli/updatecli/pull/851

PR #858, opened 2022-09-09
fix: Correctly retrieve GitHub enterprise url

olblak:
Signed-off-by: Olblak <me@olblak.com>

Fix #854 

## Test

To test this pull request, you can run the following commands:

```shell
# Build binary using Golang 1.19
go build -o bin/updatecli .
# Test using a manifest
./bin/updatecli diff --config updatecli/updatecli.d
```

### Tradeoff

I don't have access to a GitHub enterprise service to test. But we probably just read to quickly this doc https://github.com/shurcooL/githubv4#usage

PR #859, opened 2022-09-09
Fix custom github url and remove default url protocol

gerardsegarra:
Fix 854
Fix a bug where custom github urls are not considered. Also move the protocol to be specified in the url. 

### Tradeoff

If you think moving the protocol to be specified by the user is too of a breaking change I can remove it. My rationale is that naming it URL is misleading if the protocol is not added. [ref](https://www.guru99.com/url-vs-uri-difference.html)


pkg/plugins/scms/github/scm.go
```diff
@@ -21,11 +23,13 @@ func (g *Github) Clean() error {
...
 // Clone run `git clone`.
 func (g *Github) Clone() (string, error) {
-
-	URL := fmt.Sprintf("https://github.com/%v/%v.git",
+	URL := fmt.Sprintf("%v/%v/%v.git",
+		g.Spec.URL,
 		g.Spec.Owner,
 		g.Spec.Repository)
 
+	logrus.Debug("Cloning repository", URL)
```
olblak, line 31: This debug message is already displayed in the the gitgeneric library
```suggestion
```

So if you had the parameter --debug to your command you'll see the message

olblak:
Hey @gerardsegarra 
In the end this PR https://github.com/updatecli/updatecli/pull/858 is ready to be merge IMHO
And it covers your suggestion to allow specifying http protocol

olblak:
I am closing this PR in favor of https://github.com/updatecli/updatecli/pull/858
Thanks again for looking into this

PR #860, opened 2022-09-11
e2e test should be executed from repository root directory

olblak:
e2e tests were executed from e2e directory instead of the git root directory. 
This messed up manifest with relative path.

## Test

```shell
export GITHUB_ACTOR=xxx
export GITHUB_TOKEN=yyy
make build test-e2e
```


lemeurherve (approved):
Good one on the "ERROR" :)

Issue #862, opened 2022-09-12
Creating pull request for GitHub custom URLs is not working.


gerardsegarra:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Updatecli cannot create Pull Requests for custom github enterprise.

The error message is:
```
Skipping due to:
        pull Request stage:     "non-200 OK status code: 404 Not Found body: \"\""
```

After investigating for a while, I found that the problem is that `shurcooL/githubv4` [expects a graphql url](https://github.com/shurcooL/githubv4/blob/234843c633fadff9694210ce7ab97948c9148a14/githubv4.go#L22).

Just adding the graphql path to the URL parameter doesn't work (as that same url is used to pull the repo), but just by appending `/api/graphql` to [the s.URL](https://github.com/updatecli/updatecli/blob/main/pkg/plugins/scms/github/main.go#L101) when initiating the client, fixes it.

### Environment

```markdown
- OS: Linux and macos
- updatecli: 0.33.1
```

PR #863, opened 2022-09-12
Add graphql path when creating client

gerardsegarra:
Fix [862](https://github.com/updatecli/updatecli/issues/862)

After investigating for a while, I found that the problem is that shurcooL/githubv4 [expects a graphql url](https://github.com/shurcooL/githubv4/blob/234843c633fadff9694210ce7ab97948c9148a14/githubv4.go#L22).

Just adding the graphql path to the URL parameter doesn't work (as that same url is used to pull the repo), but just by appending /api/graphql to [the s.URL](https://github.com/updatecli/updatecli/blob/main/pkg/plugins/scms/github/main.go#L101) when initiating the client, fixes it.

Based on official [documentation](https://docs.github.com/en/enterprise-cloud@latest/graphql/guides/managing-enterprise-accounts#3-setting-up-insomnia-to-use-the-github-graphql-api-with-enterprise-accounts), the enterprise url for GraphQL is `https://<HOST>/api/graphql`.

### Potential improvement

Tests are missing for all the custom URLs.

### DIFF
--- a/pkg/plugins/scms/github/main.go
+++ b/pkg/plugins/scms/github/main.go
@@ -3,6 +3,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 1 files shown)


olblak (commented):
Thanks for looking into this. 
I don't have access to a GitHub enterprise so my knowledge is purely theorical

Could I ask you to use the standard url library to do a clean url join
https://pkg.go.dev/net/url@go1.19beta1#JoinPath

something like 
```

graphqlURL , err := url.JoinPath(s.URL, "/api/graphql")
if err != nil { 
    return nil, err
}
g.client = githubv4.NewEnterpriseClient(graphqlURL, httpClient)
```

gerardsegarra:
Done :D
Thank you for being so quick reviewing this.

olblak:
> Done :D Thank you for being so quick reviewing this.

Thanks for fixing this part, if all tests are green, I can cut a release

lemeurherve (approved):

PR #868, opened 2022-09-15
fix: Gitea pullrequest shouldn't try to be opened if missing remote branches

### LINKED ISSUE #867: Updatecli fails if no Gitea Pullrequest need to be created

### Current Behavior

Updatecli fails if no Gitea Pullrequest need to be created

```
TARGETS
========

ueadme
------
✔ Content from file "/tmp/updatecli/gitea/olblak/updatecli-test/README.md" already up to date
✔ All contents from 'file' and 'files' combined already up to date


PULL REQUESTS
=============

Pipeline "Test Gitea tag" failed
Skipping due to:
	pull Request stage:	"Not Found"

=============================
```

### Expected Behavior

I wouldn't expect Updatecli to fail if something has to be changed

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:Linux
- updatecli: 0.33.2
```

### Anything else?

_No response_

olblak:
Fix #867 

## Test

```shell
cp <to_package_directory>
go test
```


pkg/plugins/resources/gitea/pullrequest/create_test.go
```diff
@@ -0,0 +1,72 @@
...
+			name: "pullrequest shouldn't be created on olblak/updatecli should not exist",
+			manifest: struct {
+				URL          string
+				Token        string
+				Owner        string
+				Repository   string
+				SourceBranch string
+				TargetBranch string
+			}{
+				URL:          "try.gitea.io",
```
lafriks, line 40: try.gitea.io should not be used for tests as it's not guaranteed to be working at all times or even preserve repositories there. Please use something more stable like codeberg.org

olblak: Thanks for the feedback, `try.gitea.io` was a "temporary" solution until I find a better solution and then I discovered codeberg.org

I recently created an account over ther and now I have to migrate the tests

PR #880, opened 2022-09-22
Remove example directory

olblak:

Many examples are broken or outdated, 
I am not planning to fix them anytime soon.

The manifest in the e2e directory are regularly tested

## Test

/

### Tradeoff

I considered renaming the directory from `examples` to `examples.deprecated`, but then I realized that I would probably never fix them anyway. 
We already have many public examples.

### Potential improvement
/


lemeurherve (approved):
Good riddance I guess? 🤷

PR #883, opened 2022-09-22
chore: optimize steps order in `go` GitHub Action

lemeurherve:
No need to install and check version of tools before they're really used. If a previous step fails, it avoid unnecessary downloads and executions.


olblak (approved):

olblak:
It makes sense

PR #886, opened 2022-09-23
Deprecate depends_on in favor of dependson

olblak:

Align depends_on syntax with the current manifest convention.
This also remove the confusion introduce by jsonschema

## Test

```shell
cp <to_package_directory>
go test
```

### DIFF
--- a/pkg/core/pipeline/condition/main.go
+++ b/pkg/core/pipeline/condition/main.go
@@ -138,11 +138,11 @@ func (c *Config) Validate() error {
 	if len(c.DeprecatedDependsOn) > 0 {
 		switch len(c.DependsOn) == 0 {
 		case true:
-			logrus.Warningf("%q is deprecated in favor of %q.", "depends_on", "dependson")
+			logrus.Warningln("\"depends_on\" is deprecated in favor of \"dependson\".")
 			c.DependsOn = c.DeprecatedDependsOn
 			c.DeprecatedDependsOn = []string{}
 		case false:
-			logrus.Warningf("%q is ignored in favor of %q.", "depends_on", "dependson")
+			logrus.Warningln("\"depends_on\" is ignored in favor of \"dependson\".")
 			c.DeprecatedDependsOn = []string{}
 		}
 	}
--- a/pkg/core/pipeline/condition/main.go
+++ b/pkg/core/pipeline/condition/main.go
@@ -134,6 +134,19 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	// Handle depends_on deprecation
+	if len(c.DeprecatedDependsOn) > 0 {
+		switch len(c.DependsOn) == 0 {
+		case true:
+			logrus.Warningf("%q is deprecated in favor of %q.", "depends_on", "dependson")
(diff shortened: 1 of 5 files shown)


pkg/core/pipeline/resource/main.go
```diff
@@ -42,6 +42,8 @@ type ResourceConfig struct {
 	SCMID string `yaml:",omitempty"` // SCMID references a uniq scm configuration
 	// !deprecated, please use scmid
 	DeprecatedSCMID string `yaml:"scmID,omitempty" jsonschema:"-"` // SCMID references a uniq scm configuration
+	// !deprecated, please use scmid
```
lemeurherve, line 45:
```suggestion
	// !deprecated, please use dependson
```

pkg/core/pipeline/resource/main.go
```diff
@@ -27,7 +27,7 @@ import (
 
 type ResourceConfig struct {
 	// depends_on specifies which resources must be executed before the current one
```
lemeurherve, line 29:
```suggestion
	// dependson specifies which resources must be executed before the current one
```

pkg/core/pipeline/resource/main.go
```diff
@@ -42,6 +42,8 @@ type ResourceConfig struct {
 	SCMID string `yaml:",omitempty"` // SCMID references a uniq scm configuration
 	// !deprecated, please use scmid
 	DeprecatedSCMID string `yaml:"scmID,omitempty" jsonschema:"-"` // SCMID references a uniq scm configuration
+	// !deprecated, please use scmid
+	DeprecatedDependsOn []string `yaml:"depends_on,omitempty" jsonschema:"-"` // SCMID references a uniq scm configuration
```
lemeurherve, line 46:
```suggestion
	DeprecatedDependsOn []string `yaml:"depends_on,omitempty" jsonschema:"-"` // depends_on specifies which resources must be executed before the current one
```

pkg/core/pipeline/condition/main.go
```diff
@@ -134,6 +134,19 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	// Handle depends_on deprecation
+	if len(c.DeprecatedDependsOn) > 0 {
+		switch len(c.DependsOn) == 0 {
+		case true:
+			logrus.Warningf("%q is deprecated in favor of %q.", "depends_on", "dependson")
```
lemeurherve, line 141: Why the interpreted string here? I would have used a simple string, it won't change.
(Same below and in the other files)

lemeurherve (commented):
Can the depends_on deprecation be handled in a single place instead of being duplicated 3 times?

pkg/core/pipeline/condition/main.go
olblak: It's more a matter of taste, I prefer the readability this way than something like 

`logrus.Warningln("\"depends_on\" is deprecated in favor of \"dependson\".")`

olblak:
> Can the depends_on deprecation be handled in a single place instead of being duplicated 3 times?

Well we could refactor the code to do so but I don't think it's worth doing now.

PR #887, opened 2022-09-23
chore: don't make `sources.name` mandatory in JSON Schema

lemeurherve:
Fix #XXX

## Test

```shell
cp <to_package_directory>
go test
```


olblak (approved):
While I would have preferred to keep it mandatory as I think it improve manifest readability. I know that @dduportal was also in favor of removing this

PR #885, opened 2022-09-23
Add Json support

olblak:
Signed-off-by: Olblak <me@olblak.com>

https://github.com/updatecli/updatecli/issues/870

Add json support

## Test

To test this pull request, you can run the following commands:

```shell
cp pkg/plugins/resources/json
go test
```

### Tradeoff

I decided to use https://github.com/TomWright/dasel to manipulate Json 

### Potential improvement

We can get some inspiration from this PR to implement toml and csv support

Issue #870, opened 2022-09-17
Feature Request: Improve support for json, toml, xml, yaml and csv


samj1912:
### Is your feature request related to a problem?

Currently it is not possible to use toml or csv files as source/condition/targets.

The filter/search capabilities for yaml are also not as expressive and do not allow specifying filter expressions to target keys.

### Solution you'd like

My proposal is to incorporate dasel - https://daseldocs.tomwright.me/use-as-a-go-package

a popular configuration parsing library into updatecli as a plugin to improve its capabilities.

It supports quite a few config file types https://daseldocs.tomwright.me/usage/supported-file-types

additionally it also supports more powerful search/query/put selector expressions -

- https://daseldocs.tomwright.me/selectors/next-available-index
- https://daseldocs.tomwright.me/selectors/all-indexes
- https://daseldocs.tomwright.me/selectors/dynamic
- https://daseldocs.tomwright.me/selectors/search
- https://daseldocs.tomwright.me/selectors/search-1

### Alternatives you've considered

Use libraries like yq but although they are powerful for yaml, dasel presents a better alternative since it supports more file types.

### Anything else?

Thank you for this amazing tool!

samj1912:
Happy to work on this!

olblak:
Hi @samj1912 ,

Thank you for the detailed issue.
First of all, I would definitely be interested to support standard file such as json,toml,csv,etc.
Having them would allow to easily implement "autodiscovery" for scenarios such as automatically bumping javascript dependencies specified in packages.json, or terraform,etc. 

Toml has been recurring demand, and I also need it in my projects. But due to lack of time, I didn't go further. So we went with the "file" resource + regex. It's not ideal and not easy to maintain but it works.
You can find more context on https://github.com/updatecli/updatecli/issues/534

Regarding Dasel, I was really enthusiast with when I discovered that project. I thought I would use it and contribute some of the the logic from Updatecli to it. hence I implemented the XML resource here 
-> https://github.com/updatecli/updatecli/pull/336/files

But then I realized they are many edge cases that I needed to fix first.
The XML part drops content.
The YAML drop comments and still use goyaml v2 so I would expect other unsupported edge cases. At this time, YAML support would reintroduce issues that I fixed in Updatecli but the query mechanism is more advanced so I am still keeping an eye on it.
The Toml and Json seems to be working fine but I didn't have the time to test further. In long term I wish I had more time to contribute to Dasel as it definitely makes sense to move some logic over there.

[...]

PR #889, opened 2022-09-24
Add Toml support

olblak:

Add support to toml for source/condition/target

https://github.com/updatecli/updatecli/issues/870

Fix #XXX

## Test

```shell
cp pkg/plugins/resources/toml
go test
```

### Tradeoff

Same approach than https://github.com/updatecli/updatecli/pull/885
I am leveraging Dasel


olblak:
I need to test if Dasel preserve Toml comments

olblak:
> I need to test if Dasel preserve Toml comments

Well it doesn't

e2e/updatecli.d/success.d/toml.yaml
```diff
@@ -0,0 +1,27 @@
+name: Basic Json Example
```
olblak, line 1:
```suggestion
name: Basic Toml Example
```

pkg/plugins/resources/toml/spec.go
```diff
@@ -0,0 +1,29 @@
+package toml
+
+import "errors"
+
+type Spec struct {
+	// [s][c][t] File specifies the csv file to manipulate
```
olblak, line 6:
```suggestion
	// [s][c][t] File specifies the toml file to manipulate
```

pkg/plugins/resources/toml/spec.go
```diff
@@ -0,0 +1,29 @@
...
+	File string `yaml:",omitempty"`
+	// [s][c][t] Key specifies the query to retrieve an information from a csv file
```
olblak, line 8:
```suggestion
	// [s][c][t] Key specifies the query to retrieve an information from a toml file
```

olblak:
I played a bit with the multiple update which is super powerfull. I already works for target, I'll add support for condition but I don't think of a usage for source

olblak:
Well after additional thought, I think I need a new boolean to enable/disable the multiple query because in the case of the following query, we can't guess automatically the meaning 

Should `.employees.[*].role`

1. Return the value for key `.employees.[*].role`
2. Return "role" value for all employees

I am considering adding a spec parameter named "multipe" to enable multiple query, with default set to false

olblak:
So I decided to add the flag `multiple` to enable multiple queries at once

PR #893, opened 2022-09-26
Override error for json with failing query

olblak:
Override the error message from dasel, as it's too verbose


olblak:
Additionnal work is needed on the json resource to add multiple query which will deprecated the current PR

Issue #942, opened 2022-10-19
command execution doesn't seem to work for complex commands


JayDoubleu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Command either doesn't execute properly or executes in an odd way

### Expected Behavior

Complex commands should execute properly

### Steps To Reproduce

```
version: 0.33.3

sources:
  latestVersion:
    kind: shell
    name: Get latest version
    spec:
      command: >-
        curl -s https://api.github.com/repos/updatecli/updatecli/releases | grep tag_name | cut -d "\"" -f 4| sort -V | tail -n1
      versionfilter:
        kind: semver
    transformers:
      - findsubmatch:
          pattern: v((\d+\.)?(\d+))?(\*|\.\d+)
          captureindex: 1
```

Output:

```
+ PIPELINE +
++++++++++++



#############
# TEST.YAML #
#############


SOURCES
=======

latestVersion
-------------
The shell 🐚 command "curl -s https://api.github.com/repos/updatecli/updatecli/releases | grep tag_name | cut -d \"\\\"\" -f 4| sort -V | tail -n1" ran successfully with the following output:
----
curl 7.79.1 (x86_64-redhat-linux-gnu) libcurl/7.79.1 OpenSSL/1.1.1n-fips zlib/1.2.11 brotli/1.0.9 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.45.1 OpenLDAP/2.4.59
Release-Date: 2021-09-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL TLS-SRP UnixSockets
----

⚠ empty source returned

=============================

REPORTS:


- TEST.YAML:
        Source:
                ⚠ [latestVersion] Get latest version (kind: shell)
```

### Environment

```markdown
- OS:
- updatecli:
```

olblak:
Thanks for opening this issue as it triggered use cases that I didn't think about initially.

[...]

PR #944, opened 2022-10-19
Allow multiple query combined with versionfilter to json source

olblak:

As discussed on https://github.com/updatecli/updatecli/issues/942#issuecomment-1284561424
I realized that it would be super usefull to be able to do a multiple query combined with versionfilter on the json resource

This pullrequest allows the following manifest to work 

```
sources:
  semverVersion:
    kind: json
    name: Get latest version from remote json file
    spec:
      file: https://api.github.com/repos/updatecli/updatecli/releases
      key: ".[*].tag_name"
      multiple: true
      versionfilter:
        kind: semver
        pattern: "v0.35"
```

## Example 

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "/tmp/updatelci.yaml"

SCM repository retrieved: 0


++++++++++++++++++
+ AUTO DISCOVERY +
++++++++++++++++++



++++++++++++
+ PIPELINE +
++++++++++++



##################
# UPDATELCI.YAML #
##################


SOURCES
=======

semverVersion
-------------
Searching for version matching pattern "v0.35"
✔ Value "v0.35.1", found in file "https://api.github.com/repos/updatecli/updatecli/releases", for key ".[*].tag_name"'

=============================

REPORTS:


- UPDATELCI.YAML:
	Source:
		✔ [semverVersion] Get latest version from remote json file (kind: json)

Run Summary
===========
Pipeline(s) run:
  * Changed:	0
  * Failed:	0
  * Skipped:	1
  * Succeeded:	0
  * Total:	1

```

## Test

```shell
go build -o bin/updatecli .
./bin/updatecli --config manifest_mentioned_here.yaml
go test
```

### Potential improvement

I'll try to find some time to open a similar pullrquest for toml and csv as they also use Dasel
I need to investigate for the xml resource but it should be possible
Unfortunately, it won't work for the yaml resource as it doesn't support multiple query


dduportal:
Like in https://github.com/updatecli/updatecli/issues/945#issuecomment-1285044879, it seems that this would be a broader change as it implies having a multi-valued source.

How is the order managed?

dduportal:
Can you update the PR description to show the result? The example is not clear on what is it doing exactly?

olblak:
> Like in [#945 (comment)](https://github.com/updatecli/updatecli/issues/945#issuecomment-1285044879), it seems that this would be a broader change as it implies having a multi-valued source.

The source output doesn't change as we still return one single value. Here `versionfilter` works the same way than for plugins

- githubrelease
- giteaTag
- gittag
- dockerimage

Basically we retrieve a list of information and we delegate the responsability to the versionfilter to identify the value we are looking for

So this pullrequest introduced two parameters

1. `multiple`: It's used to describe that we are looking for multiple values from the json query and not only one. Because `.[*].tag_name` is a valid json key identifer
1. `versionfilter`: It's used to filter the result from the multiple query 

In the context of a source, both must be used together

olblak:
> Can you update the PR description to show the result? The example is not clear on what is it doing exactly?

I added an example in the PR body

olblak:
I am open to change if it makes things clearier. Ideally I would like to use the same wording for csv/toml/xml/json/yaml
Maybe @hervelemeur has some opinion too

olblak:
Before merging I need to correctly deprecate the key `multiple` previously used for conditoin and target

PR #946, opened 2022-10-20
Fix e2e tests

olblak:
I noticed a minor wrong test


olblak:
Well the test won't work

PR #955, opened 2022-10-27
fix(pullrequest): default title with the source output instead of its result '✔'

lemeurherve:
Fix 
image

## Test

```shell
cp <to_package_directory>
go test
```

### Potential improvement

I intend to improve this pull request default title by including the name of what's updated, and maybe also the type.

Something like

> [updatecli] Bump `<name>` <helm chart|docker image|...> version to <source-output>

The easiest way I found to retrieve the name would be to use sources[0].name, implying the change of (jenkins infra) manifests (they almost all have a description in the `name` field)

### DIFF
--- a/pkg/core/pipeline/pullRequests.go
+++ b/pkg/core/pipeline/pullRequests.go
@@ -58,7 +58,7 @@ func (p *Pipeline) RunPullRequests() error {
 		if len(pr.Title) == 0 {
 			pr.Title = p.Config.GetChangelogTitle(
 				firstTargetID,
-				p.Sources[firstTargetSourceID].Result,
+				p.Sources[firstTargetSourceID].Output,
 			)
 		}
 


olblak:
> I intend to improve this pull request default title by including the name of what's updated, and maybe also the type.
I think it would only work if the pullrquest only depends on one source. Otherwise it can be confusing.

olblak (approved):
Nice catch, I am surprised that this bug lived for so long :o

lemeurherve:
> I think it would only work if the pullrquest only depends on one source. Otherwise it can be confusing.

Yes there will be special cases, I intend to respond to the common cases mainly, like one source, and fallback to a default title like the current one otherwise.

olblak:
> > I think it would only work if the pullrquest only depends on one source. Otherwise it can be confusing.
> 
> Yes there will be special cases. I intend to respond to the common cases mainly (like one source) and fallback to a default title like the current one otherwise.

Is this something you want to implement in this PR?
Imho the current one is ready to be ship, I am planning to trigger a release next week

lemeurherve:
> Is this something you want to implement in this PR?

No. 🚢

PR #1003, opened 2022-11-26
Add Dockerfile Autodiscovery

olblak:
Add Dockerfile Autodiscovery.

- [x] Add [documentation](https://github.com/updatecli/website/pull/594) 

The purpose of the Dockerfile autodiscovery is to update docker image tag used in FROM instruction.
If a FROM instruction relies on ARG then we try to update the ARG instead. 
The audiscovery covers the two following scenarios

```
# Simple case to update the FROM
FROM jenkins/jenkins:2.35.1-alpine

# From instruction with a platform 
FROM --platform="linux/arch64" jenkins/jenkins:2.35.1-alpine

# ARG case where we update the ARG instead of the FROM
ARG jenkins_version 2.235.1
FROM jenkins/jenkins:${jenkins_version}-alpine
```

Fix #XXX

## Usage 

This crawler can be used with and without manifest 
Without manifest, you can run `updatecli diff --experimental`

With a manifest such as the following one, `updatecli diff --config manifest.yaml --experimental`

```
name: "[asciidoctor/docker-asciidoctor] Docker compose autodiscovery using git scm"
scms:
  asciidoctor:
    kind: git 
    spec:
      url: https://github.com/asciidoctor/docker-asciidoctor.git
      branch: main
    
autodiscovery:
  # scmid is applied to all crawlers
  scmid: asciidoctor
  crawlers:
    dockerfile:
      # Override the default root directory where to look for Dockerfile
      rootdir: 
      ignore:
        - archs:
          - amd64
          path: Dockerfile.dapper
          images:
            - "rancher/k3d:1.0.0"
            - "rancher/fleet"
            - "nginx"
      only:
        - archs:
            - amd64
          path: Dockerfile.dapper
          images:
            - "rancher/k3d:*"
            - "nginx"
      filematch:
        - "Dockerfile.*"
```

## Test

```shell
pkg/plugins/autodiscovery/dockerfile/
go test
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/warning.d/autodiscovery/dockerfile --experimental
```

### Potential improvement

The ignore image rule could benefits from a regex


macedogm:
@olblak thanks for working on this PR. I haven't read it entirely, but is the plan to also update the image tag in these situations?

1. Both image and tag are referenced in the `ARG`:
```Dockerfile
ARG ALPINE=alpine:3.16
FROM ${ALPINE}
```

2. The image tag has two parts that can be updated independently and can target multiple versions - `1.18.8-alpine3.17`, `1.19.3-alpine3.16` and `1.19.3-alpine3.17`.
```Dockerfile
FROM golang:1.18.8-alpine3.16
```

For the second case I had to use two `sources`, one to get the latest golang version and the second to get the latest alpine variant based on the latest golang version. You can see in https://github.com/k3s-io/k3s/pull/6583/files#diff-110e31729b81049475207c4c2765d5dd2d682899b1a3d1a0e29a13ccc1587f7aR23-R42 and https://github.com/macedogm/test-k3s/pull/16/files.

olblak:
Thanks for the feedback.

> 1. Both image and tag are referenced in the `ARG`:

```Dockerfile
ARG ALPINE=alpine:3.16
FROM ${ALPINE}
```

While it's not a use case that I initially thought about, I could easily support it.

> 2. The image tag has two parts that can be updated independently and can target multiple versions - `1.18.8-alpine3.17`, `1.19.3-alpine3.16` and `1.19.3-alpine3.17`.

To be honest I don't see how I could automatically detect such values.
The purpose of the autodiscovery is to catch common pattern and to rely on the "classical" approach to handle this scenario

olblak:
> While it's not a use case that I initially thought about, I could easily support it.
In fact it's already supported but I realize that I didn't considered the case where we have multiple arg such as

```
ARG image=alpine
ARG tag=3
FROM ${image}:${tag}
```

olblak:
> While it's not a use case that I initially thought about, I could easily support it.
After additional thought, I don't think I'll be able to easily handle it :D

macedogm:
Thanks for the update. I'll see if there is anything that I can do to help. Those are not the trivial cases.

pkg/plugins/autodiscovery/dockerfile/matchingRule.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// MatchingRule allows to specifies rules to identify manifest
+type MatchingRule struct {
+	// Arch specifies a list of docker image architecture
+	Archs []string
+	// Path specifies a Helm chart path pattern, the pattern requires to match all of name, not just a substring.
```
olblak, line 14:
```suggestion
	// Path specifies a Dockerfile path pattern, the pattern requires to match all of name, not just a substring.
```

olblak:
After looking more into this, supporting multiple ARG in a FROM instruction is definitely something more difficult and I won't support that case. Instead I trigger a warning when the situation appear.

The current implement works this way.

Dockerfile
```
ARG version 0.38.0
FROM updatecli/updatecli:${version}
```

1. Identify a FROM instruction relying on a variable such as `FROM updatecli/updatecli:${version}`
2. Extract the substring before and after the variable. Prefix is `updatecli/updatecli:` and suffix is empty
3. Replace the variable content by an ARG value
4. So we now have a FROM instruction such as `FROM updatecli/updatecli:0.38.0`
5. Look for the next version matching pattern `x.y.x` 
6. Found 0.39.0, so `updatecli/updatecli:0.39.0`
7. Trim prefix (updatecli/updatecli:) and suffix (empty) retrieved in step 2
8. The ARG  value is updated to 0.39.0

If we have more than one ARG then we have more than one unknown information which increase the complexity to retrieve the information...

Issue #1004, opened 2022-11-27
Commit message's title is not well formatted when it's longer than 72 characters


macedogm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The current logic  to break the commit message's title when it's longer than 72 characters is not properly working, because it doesn't count the characters `(`, `)`, `:` and ` ` used in the template message.

https://github.com/updatecli/updatecli/blob/0a4f057cb7732f26feab143e7c85791c174e753b/pkg/plugins/scms/git/commit/main.go#L82

https://github.com/updatecli/updatecli/blob/0a4f057cb7732f26feab143e7c85791c174e753b/pkg/plugins/scms/git/commit/main.go#L13

This makes the title appear broken. See example in https://github.com/rancher/rancher/pull/39696/commits/53f7944f751afe2fe25e4d32139f2ed782729adf.

```
chore: Updated the content of the file "/tmp/updatecli/github/rancher…
…/r...

... ancher/tests/v2/validation/charts/monitoring.go"

Made with ❤️️ by updatecli
```

### Expected Behavior

The logic should properly take into account those characters and properly break the title. For example:

```
chore: changed lines [43] of file "/var/folders/gs/8nqtp16979lbq32ycb...
... cgnpcm0000gp/T/updatecli/github/macedogm/rancher/package/Dockerfile"

Made with ❤️️ by updatecli
```

### Steps To Reproduce

Use a config that has the resource `pullrequests` to create a PR. See that the commit message's title is not properly broken, as presented above.

### Environment

```markdown
- OS: Mac (arm64) and Ubuntu 22.04 (amd64)
- updatecli: v0.38.0
```

Issue #1005, opened 2022-11-27
Wrong commit message when a PR from resource `pullrequests` has 2 or more commits


macedogm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a PR is made from a resource `pullrequests` and it has 2 or more commits from multiple `targets`, the commit message for the second and afterwards commits is ignored and it's always equal as the first commit.

For example, see PR https://github.com/rancher/rancher/pull/39696 and the 3 commits:
1. https://github.com/rancher/rancher/pull/39696/commits/53f7944f751afe2fe25e4d32139f2ed782729adf
2. https://github.com/rancher/rancher/pull/39696/commits/08380e994f72cf9987365fb4044b9562500c9667
3. https://github.com/rancher/rancher/pull/39696/commits/48c7559204b23eeedbf342d11b9af3a0562d6e21

Each commit updates a different file, but the message for commits 2 and 3 is the same as commit 1.

This happens, because the commit message is not a new variable for each commit. It's currently being update as a global variable.

https://github.com/updatecli/updatecli/blob/0a4f057cb7732f26feab143e7c85791c174e753b/pkg/plugins/scms/git/commit/main.go#L104-L113

Additionally, if the commit title is set in the configuration, but body is not set, then the commit message will be broken. It will use the configured commit title, but the body from the generated commit itself. For example:

```
chore: Bump rancher/shell image version
... ash.sh e7a1cc6a9b414787b3dc8937d614dd30649864ec"

Made with ❤️️ by updatecli
```

### Expected Behavior

Each commit must have its own commit message.

### Steps To Reproduce

See PR https://github.com/rancher/rancher/pull/39696 and the used configuration in https://github.com/rancherlabs/updatecli-automation/blob/34a0cc4687c8c069feaed457cf843c6cbd6a8507/updatecli/updatecli.d/rancher/shell/rancher/v2.7/shell.yaml#L49-L76.

### Environment

```markdown
- OS: Mac (arm64) and Ubuntu 22.04 (amd64)
- updatecli: v0.38.0
```

macedogm:
@olblak this can also be closed with https://github.com/updatecli/updatecli/pull/1006.

dduportal:
Closing as requested. Please note that 0.38.1 is currently being released with your fixes. Many thanks for the contribution!

PR #1006, opened 2022-11-27
Fix commit message when there is 2 or more commits for the same PR

macedogm:
Fixes https://github.com/updatecli/updatecli/issues/1004 and https://github.com/updatecli/updatecli/issues/1005

## Test

See the steps to reproduce described in each issue.

### Potential improvement

Perhaps the variable `placeholders` can have a better name.

Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

### DIFF
--- a/pkg/plugins/scms/git/commit/main.go
+++ b/pkg/plugins/scms/git/commit/main.go
@@ -51,7 +51,7 @@ func (c *Commit) Generate(raw string) (string, error) {
 
 	t := template.Must(template.New("commit").Parse(commitTpl))
 
-	err = c.ParseMessage(raw)
+	commit, err := c.ParseMessage(raw)
 
 	if err != nil {
 		logrus.Errorf(err.Error())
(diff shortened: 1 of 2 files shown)


olblak:
Thanks for the PR, I am just waiting @dduportal review as he's currently looking into the scm part :)

olblak:
I suspect that the function is call in a test file hence the linting error

macedogm:
I totally missed the test file. I've updated it now and tests are passing.

However, I see that it's still throwing an error, which also happens before the changes made.

```
% go test                                                                                 
ERRO[0000] error: empty commmit messager                
PASS
ok  	github.com/updatecli/updatecli/pkg/plugins/scms/git/commit	0.617s
```

Unfortunately I wasn't able to clean this error, I might have missed something simple.

dduportal (approved):
LGTM thanks a lot!

PR #1029, opened 2022-12-10
Add Helm Chart OCI support

### LINKED ISSUE #811: Feature Request: Add OCI registry support to helm resource

Helm now support OCI registry as a way to store helm package.
Considering that Updatecli already support OCI registry in the dockerimage resource. I should be straightforward to add OCI registry support to the helm chart as well

### Solution you'd like

```
sources:
  kubed:
    name: "Get Latest Helm Chart version"
    kind: "helmchart"
    spec:
      url: "oci://localhost:5000/helm-charts/mychart"
```

So if url start with "oci//" then we know we are facing an oci registry

_No response_

### Anything else?

https://helm.sh/docs/topics/registries/

olblak:
Fix #811 

- [ ] Add OCI support to Helm Resource
- [ ] Add OCI support for Helmfile autodiscovery
- [ ] Add OCI support for Helm autodiscovery

## Test

```shell
cp <to_package_directory>
go test
```

### DIFF
--- a/e2e/updatecli.d/success.d/helmchart.yaml
+++ b/e2e/updatecli.d/success.d/helmchart.yaml
@@ -47,6 +47,13 @@ sources:
     spec:
       name: epinio
 
+  fromOCI:
+    name: Get upgrade-responder
+    kind: helmchart
+    spec:
+      url: oci://ghcr.io/olblak/charts/
+      name: upgrade-responder
+
 conditions:
   default1:
     name: Retrieve Version v1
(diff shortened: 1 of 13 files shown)


olblak:
I temporary commented the OCI integration tests.
ghcr.io doesn't allow anonymous query to ghcr.io and I don't have yet image available on dockerhub

since I am planning to build and publish an oci helm chart for updatemonitor which requires a bit of work.
I am planning to merge this PR as is then come back later to fix the tests.

Issue #465, opened 2022-01-14
Feature Request: Introduces the concept of "actions"


dduportal:
### Is your feature request related to a problem?

This is an important feature request, discussed with @lemeurherve, to upgrade the object of type `pullRequest` to meet the following goals for end users:

- Solve the "SCM/Local" problem (#262) once for all
- Make the `pullRequest` concept not tied to GitHub (e.g. allows to use "pure Git", but would also make it compliant with GitLab or Mercurial or any other tool not using the wording "Pull Request")
- Allowing to choose wether or not to squash the commits resulting of multiple targets

We would want these 3 elements solved without too much major breaking change of course, in an expressive way allowing to keep all existing use cases.

### Solution you'd like

Example of syntax that we are proposing to evolve the actual (0.18.x):

```yaml
## No changes for scms/sources/conditions
title: "Bump LDAP Helm Chart Version"
scms:
  default:
    kind: github
    spec:
      user: "{{ .github.user }}"
      email: "{{ .github.email }}"
      owner: "{{ .github.owner }}"
      repository: "{{ .github.repository }}"
      token: "{{ requiredEnv .github.token }}"
      username: "{{ .github.username }}"
      branch: "{{ .github.branch }}"
sources:
  lastChartVersion:
    kind: helmChart
    spec:
      url: https://jenkins-infra.github.io/helm-charts
      name: ldap

## Changes for targets: no more scm/scmID concept by default, see below
targets:
  updateChartVersion:
    name: "Update the chart version for ldap"
    kind: file
    spec:
      file: clusters/prodpublick8s.yaml
      matchPattern: 'chart: jenkins-infra\/ldap((\r\n|\r|\n)(\s+))version: .*'
      replacePattern: 'chart: jenkins-infra/ldap${1}version: {{ source `lastChartVersion` }}'

## Changes: pullrequest renamed to "actions" (naming is subject to debate, see below)
actions:
  default:
    # New kind of plugin, extending actions, see below
    kind: github-pullrequest
    scmID: default
    targets:
      - updateChartVersion
    spec:
      labels:
        - dependencies
        - ldap
```

```shell
## No change: test source -> condition -> target, with dry run for targets
$ updatecli diff --config manifest.yaml --values values.yaml

## New flag that ensures that the target is applied locally but the "actions" are not applied
$ updatecli apply --config manifest.yaml --values values.yaml --no-actions # Or --actions=false

## No change: Apply all, as before
[...]

PR #1030, opened 2022-12-11
wip: remove concept of SCM from target package

dduportal:
Related to https://github.com/updatecli/updatecli/issues/465

## Test

```shell
cp <to_package_directory>
go test
```


dduportal:
I'm closing this PR. It was an experiement and it allowed to draw the following conclusions on our codebase:

- Removing the `scmid` field for targets breaks some use cases:
  - Using updatecli manifest in a different repository than the files changed by the targets
  - Opening PR on a different repository than the current repository

- Resources plugins should be simplified:
  - The methods `TargetFromSCM()`, `ConditionFromSCM()` and `SourceFromScm()` must die, because they make plugins code difficult to write and maintain (and some plugins still have different behaviors wether you set `scmid` or not...).
  - The target object should provide the "working directory" to the plugins `Target()` and abstract away the whole "temp dir, git clone, etc." logic from plugin
  - The action objects should abstract the "list modified files/stage changed content/git add / git commit/git push/etc." logic away from plugins (a plugin should NOT specify a commit message) and from target

PR #990, opened 2022-11-19
breaking: rename 'pullrequests' to 'actions'

dduportal:
Related to #465.

This is the first "atomic step" inspired by the explorations done in #575 .

This PR changes the concept of "Pull Request" in the manifests to "Actions".

WiP.

Todo list:

- [x] rename code reference to "action"
- [ ] Add code to deprecate "pullrequests" (both user messages and conversion to action under the hood)
- [ ] Create a package for the plugin github/pullrequest of kind actions
- [ ] Create a package for the plugin gitea/pullrequest of kind actions
- [ ] update examples 
- [ ] update tests
- [ ] update documentation (PR on updatecli/website)

## Test

To test this pull request, you can run the following commands:

```shell
cp <to_package_directory>
go test
```

PR #1037, opened 2022-12-12
fix(actions) print the deprecation action's kind warning only once

### LINKED ISSUE #1035: Wrong warning message sent from Github resource of type source,condition,target

### Current Behavior

Current Updatecli run show the wrong warning message for source/condition/target which shouldn't be updated

```
#######################
# BUMP GOLANG VERSION #
#######################


SOURCES
=======
WARNING: The kind "github" for actions is deprecated in favor of 'github/pullrequest'
```

### Expected Behavior

No message should be displayed

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- updatecli:
```

### Anything else?

_No response_

dduportal:
Fix #1035

When writing #990 , I introduced a minor bug where the warning message telling the user about using a kind of action that is deprecated.

I forgot to update the internal config state, so the processing of template triggered the message at least a 2nd time (or more...).

This PR adds the required unit test that fails with the `0.40.0` behavior.

## Test

```shell
# Build a local updatecli binary
go build -o ./dist/updatecli

# Remove the `version:` top level directive to avoid a failure, as local binary builds does not have a version
sed '/version:/d' ./updatecli/updatecli.d/golang.yaml | wc -l

# Execute the golang manifest, with only 1 warning this time
./dist/updatecli diff --config ./updatecli/updatecli.d/golang.yaml
```

result:

```text
./dist/updatecli diff --config ./updatecli/updatecli.d/golang.yaml


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "./updatecli/updatecli.d/golang.yaml"
WARNING: The `pullrequests` keyword is deprecated in favor of `actions`, please update this manifest. Updatecli will continue the execution while trying to translate `pullrequests` to `actions`.
WARNING: The kind "github" for actions is deprecated in favor of 'github/pullrequest'

SCM repository retrieved: 1

# ...
```

## Additional Information

- Using the `version` top-level directive is blocking local builds and debug. We should set a default "dev" or "debug" version by default, that would be accepted. It can be cumbersome when debugging.
  - Should we remove it from the manifests? I can't remember the use case of specifying it?

- The unit test does not cover the whole package code: room for improvement :)
  - There is a (minor) bug caught by unit tests (I did not added the failing unit tests here as it's another scope) that I'll try to fix tomorrow.

### DIFF
--- a/pkg/core/pipeline/action/main.go
+++ b/pkg/core/pipeline/action/main.go
@@ -72,9 +72,11 @@ func (c *Config) Validate() (err error) {
 	/** Deprecated items **/
 	if c.Kind == githubIdentifier {
 		logrus.Warnf("The kind %q for actions is deprecated in favor of '%s/pullrequest'", githubIdentifier, githubIdentifier)
+		c.Kind = "github/pullrequest"
 	}
 	if c.Kind == giteaIdentifier {
 		logrus.Warnf("The kind %q for actions is deprecated in favor of '%s/pullrequest'", giteaIdentifier, giteaIdentifier)
+		c.Kind = "gitea/pullrequest"
 	}
 	if c.DeprecatedScmID != "" {
 		switch len(c.ScmID) {
--- a/pkg/core/pipeline/action/main_test.go
+++ b/pkg/core/pipeline/action/main_test.go
@@ -1,0 +1,81 @@
+package action
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Validate(t *testing.T) {
+	tests := []struct {
+		name           string
+		config         Config
+		wantErrMessage string
+		wantConfig     Config
+	}{
+		{
+			name:           "Failing case with missing 'Kind' and 'scmid'",
+			config:         Config{},
+			wantErrMessage: `missing value for parameter(s) ["kind,scmid"]`,
+		},
+		{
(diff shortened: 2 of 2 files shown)


olblak:
I was just about to start debugging this issue and I saw your PR ❤️ 

> I've run the case with a debugger: the double validation message appears because the manifest is templated.
Each time the template is rendered (e.g. twice), the validation happens on the newly update template. 

Nice catch

> Using the version top-level directive is blocking local builds and debug. We should set a default "dev" or "debug" version by default, that would be accepted. It can be cumbersome when debugging.

I agree, I also face the issue. The version is used by the `updatecli manifest upgrade` to check that we can upgrade a new version syntax and not downgrade a manifest. The problem here the variable "version.Version" is not defined at build time

olblak:
I confirm that your PR also fix the `updatecli manifest upgrade`

```
⚠ Updatecli manifest change required
--- updatecli/updatecli.d/venom.yaml(old)
+++ updatecli/updatecli.d/venom.yaml(updated)
@@ -1,9 +1,9 @@
 name: Bump venom version
 pipelineid: 6a1c726e25648f93a45aa3b540ba8fd6307a62f92830404f55259bfbbc1b3c6f
-pullrequests:
+actions:
     default:
         title: '[updatecli] Bump Venom version to {{ source "latestVersion" }}'
-        kind: github
+        kind: github/pullrequest

```

olblak (approved):

PR #1057, opened 2022-12-20
Allows to reverse expected condition result

### LINKED ISSUE #1053: Feature Request: Allow to define expected Condition result

Sometimes I would like my condition to return `false` as the expected result instead of checking for a result to be `true`

### Solution you'd like

I would like to use the key "false" such as 

```
conditions:
    dockerTag:
        name: Is docker image golang:{{ source "latestGoVersion" }} published
        kind: dockerimage
        spec:
            image: golang
            tag: '{{ source "latestGoVersion" }}'
        scmid: default
        sourceid: latestGoVersion
        false: true
```

_No response_

### Anything else?

I also would like to be able to combine condition result but I don't we could express with the current syntax

olblak:
Fix #1053 

## Test

```shell
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/gitea/tag.yaml
```

### DIFF
--- a/e2e/updatecli.d/success.d/gitea/tag.yaml
+++ b/e2e/updatecli.d/success.d/gitea/tag.yaml
@@ -32,7 +32,8 @@ conditions:
       url: "try.gitea.io"
       owner: "olblak"
       repository: "updatecli-test"
-      tag: "0.0.2"
+      tag: "99.0.2"
+    failwhen: true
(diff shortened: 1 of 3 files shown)


lemeurherve (approved):

olblak:
Thanks @lemeurherve

PR #1071, opened 2023-01-05
Allow to define `changedif` behavior for the shell resource

olblak:
Opening for visibility

This pullrequest allows to have different success criteria for the shell resource.
It's an initial implementation based on this [discussion](https://github.com/updatecli/updatecli/discussions)
I am introduce three  new success criteria which are:
*  `console/output` - (current shell behavior)
* `exitcode` where we can specify which exit code to map to.
* `file/checksum`: that monitor file checksum to determine success

While the "console/output" introduce no change as it requires no configuration, we can use the following snippet to customize the exit code

```
sources
  echo:  
    name: Should be succeeding
    kind: shell
    spec:
      command: "echo 1.0.0"
      changedif:
        kind: exitcode
        spec:
          warning: 2
          failure: 1
          success: 0
``` 

```
conditions:
  success:
    disablesourceinput: true
    name: Test the condition succeed according success criteria despite command failing
    kind: shell
    spec:
      command: "false"
      changedif:
        kind: file/checksum
        spec:
          files:
            - e2e/updatecli.d/success.d/shell/checksum.yaml
```

## Test

```shell
cp pkg/plugins/resources/shell/success/console/
go test 
pkg/plugins/resources/shell/success/exitcode/
go test
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/command.yaml
./bin/updatecli diff --config e2e/updatecli.d/success.d/shell
```

### Tradeoff

I must admit that I am not a big fan of the word "outcome" another one that I could think of is "successcriteria" but I find it a bit long...

**EDIT**
I decided to go with `success` as I think it's pretty self explanatory
**EDIT**
Finally I leaning toward `changedif`
### Potential improvement

As suggested in the discussion, I could imagine introducing more possible outcome rules such as `file/exist` when we need to monitor a lot of files


olblak:
I think instead of outcome, 'success' would better capture the meaning

olblak:
Damien suggestion would be to replace `success` to `changed`

dduportal:
> Damien suggestion would be to replace `success` to `changedif`

`changed_if` to be precise :)

olblak:
> > Damien suggestion would be to replace `success` to `changedif`
> 
> `changed_if` to be precise :)

After some experimentation with the jsonschema `changed_if` sounds fine so I'll proceed with it

olblak:
For a reason that I ignore the following line 

```
ChangedIf SpecChangedIf `yaml:"changed_if,omitempty" json:"changed_if,omitempty"`
```

Do no extract the `changed_if` but `changedif`, without the lowercase

olblak:
But the json tag works correctly as the generated jsonschema has the property `changed_if`

olblak:
Even thought I prefer the lowercase syntax, I can't figure out why the yaml tag is not correctly used. So I am just going to move forward without the underscore

olblak:
Once the tests are passing here, I'll merge this pullrequest so I can finalize the npm autodiscovery one

olblak:
Same remark than [here](https://github.com/updatecli/updatecli/pull/1112#issuecomment-1405416144) until we release 0.44.0 after FOSDEM. I am open to change the name. Once release it will have to be deprecated

PR #1076, opened 2023-01-09
chore(e2e tests) Comment e2e `json` test to avoid GitHub api rate limit

olblak:
Comment some e2e tests that rely on GitHub api which is rate limited

## Test
/


dduportal (approved):

olblak:
As suggested by @dduportal I'll publish a dummy json (or reuse an existing one such as

olblak:
I know realize that I should be using the jsonschema published on [www](https://www.updatecli.io/schema/latest/config.json)

PR #1079, opened 2023-01-10
Only set git auth if both username && password are specified

olblak:

Gitea e2e started failing a week or two ago. 
I first noticed it on `try.gitea.io`.
I assumed it was due to that service being an unstable one so I moved the test to codeberg but the same issue appeared yesterday.
After some quick testing, I noticed that configuring http auth with no username/password would trigger errors
This PR only set http auth if both username&password are defined

## Test

```shell
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/gitea/scm.yaml --debug --clean=true
```

Also did some testing with private repo and it works as expected

### Potential improvement

Similarly to the Github scm, I am wondering if the username is needed

### DIFF
--- a/pkg/plugins/utils/gitgeneric/main.go
+++ b/pkg/plugins/utils/gitgeneric/main.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


olblak:
Merging for now so e2e pass.

PR #1081, opened 2023-01-11
[source][condition] Add Cargo Package support

loispostula:
Implement a new plugin to support cargo packages

This PR allows using Cargo Package as a source and a condition.

By default, it will use `crates.io` index repository, but a private repository can be used as well

## Test

```shell
cd pkg/plugins/resources/cargopackage
go test
```


olblak:
Waow nice, I am glad to see you manage to create a new resource without reaching out for help :D
I'll find some time later today or this week to review your PR

loispostula:
No problem, I've remove all the git handling, I had not understood the scm concept fully. I'm adapting the description as well to remove the tradeoff

pkg/core/pipeline/resource/main.go
```diff
@@ -118,6 +121,7 @@ func GetResourceMapping() map[string]interface{} {
 
 	return map[string]interface{}{
 		"aws/ami":       &awsami.Spec{},
+		"cargopackage":  &cargopackage.Spec{},
```
olblak, line 124: I am wondering if it's not better to just use "cargo" instead of "cargopackage"
@hervelemeur is more picky than me on this :)

pkg/plugins/resources/cargopackage/main.go
```diff
@@ -0,0 +1,159 @@
...
+	newResource := &CargoPackage{
+		spec:          newSpec,
+		versionFilter: newFilter,
+	}
+
+	return newResource, nil
+}
+
+// Validate tests that tag struct is correctly configured
+func (cp *CargoPackage) Validate() error {
```
olblak, line 68:
```
func (s *Spec) Validate() error {
```

IHMO this function should be moved to the file `spec.go` and validate the spec object only

It should only be called from the function `New(spec interface{}) (*CargoPackage, error) {`

pkg/core/pipeline/resource/main.go
loispostula: I'm open to change, I used `cargopackage` to be consistent with `dockerdigest`, `dockerfile`, `dockerimage`, ...

pkg/plugins/resources/cargopackage/main.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// Get package data from Json API
+func (cp *CargoPackage) getPackageData() (PackageData, error) {
+	var pd PackageData
+
+	pd.Name = cp.spec.Package
+
+	packageDir, err := getPackageFileDir(cp.spec.Package)
```
olblak, line 132: You don't handle the error

pkg/plugins/resources/cargopackage/source.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/updatecli/updatecli/pkg/core/result"
+)
+
+// Source returns the latest npm package version
+func (cp CargoPackage) Source(workingDir string) (string, error) {
+	if len(cp.spec.IndexDir) == 0 && len(workingDir) > 0 {
+		cp.spec.IndexDir = workingDir
+	}
+
+	err := cp.Validate()
```
olblak, line 15: This function could be moved to the function New() so we also validate the spec for condition and target

olblak:
Coud I also ask you to add an manifest to `e2e/updatecli.d/success.d/crate.yaml`?
The manifest cannot return any error or warning message as define on https://github.com/updatecli/updatecli/blob/main/e2e/venom.d/test_diff.yaml

olblak:
I think it's missing the part where you dowload the index from the crate.io repository.

I just tested the following manifest

```
sources:
  test:
    kind: cargopackage
    spec:
      package: syn
```

and I get 

```
SOURCES
=======

test1
-----
ERROR: something went wrong while opening the package file "open /home/olblak/Projects/Updatecli/updatecli/3/s/syn: no such file or directory"
ERROR: ✗ open /home/olblak/Projects/Updatecli/updatecli/3/s/syn: no such file or directory
Pipeline "CRATE.YAML" failed
Skipping due to:
	sources stage:	"open /home/olblak/Projects/Updatecli/updatecli/3/s/syn: no such file or directory"
```

This makes me think you may want to download the repository in the temporary directory created by Updatecli
https://github.com/updatecli/updatecli/blob/main/pkg/core/tmp/main.go#L25

loispostula:
You mean that we should default to use and clone the `crates.io` repository, if no `IndexUrl` or `scmId` is defined?

pkg/plugins/resources/cargopackage/source.go
```diff
@@ -0,0 +1,44 @@
...
+			indexDir, err := loadDefaultCrateIndex()
+			if err != nil {
+				return "", err
+			}
+			cp.spec.IndexDir = indexDir
+		} else {
+			cp.spec.IndexDir = workingDir
+		}
+	}
+
```
loispostula, line 30: This is in order to detect wether or not we are getting the source in the context of `scm` or not. Is there a better way to do it?

pkg/plugins/resources/cargopackage/main.go
```diff
@@ -0,0 +1,198 @@
...
+			err != git.NoErrAlreadyUpToDate {
+			logrus.Errorf("something went wrong when pulling the existing repo %q\n", err)
+			return "", err
+		}
+	}
+	if err != nil {
+		return "", err
+	}
+	return indexPath, nil
+}
```
loispostula, line 198: Is there a way to inject a `scm` entry from a condition or source context in order to prevent reimplementing `git clone` as it's a feature handled by `scm`?

pkg/plugins/resources/cargopackage/source.go
olblak: I didn't test it but something like 

```
package cargopackage

import (
	"github.com/updatecli/updatecli/pkg/plugins/scms/git"
)

func cloneCrateGitRepository() (string, error) {

	s, err := git.New(
		git.Spec{
			URL: "https://github.com/rust-lang/crates.io-index.git",
		},
	)

	if err != nil {
		return "", err
	}

	return s.Clone()
}
```

pkg/plugins/resources/cargopackage/condition.go
```diff
@@ -0,0 +1,66 @@
...
+// ConditionFromSCM test if a tag exists from a git repository specific from SCM
+func (cp *CargoPackage) ConditionFromSCM(source string, scm scm.ScmHandler) (bool, error) {
+	path := scm.GetDirectory()
+
+	if len(cp.spec.IndexDir) > 0 {
+		logrus.Warningf("IndexDir is defined and set to %q but is overridden by the scm definition %q",
+			cp.spec.IndexDir,
+			path)
+	}
+	cp.spec.IndexDir = path
```
olblak, line 35: We try to avoid modifying the spec parameter.
The reason is at any moment in time we want to be able to identify the user input. It's more a convention in the code base 

Instead we copy the value to a new variable and then that variable can be change. If needed we still have the initial value in the spec

olblak: In this case that would look like `cp.IndexDir = path` where `cp.IndexDir` is init in the New function

pkg/plugins/resources/cargopackage/source.go
olblak: I just did some testing and while it works, it just super slow like a minute
I didn't try your approach

loispostula: I agree, the crates.io repo is huge, so when we pull it it takes time. Hence why I wanted to let the user use a scm resource to do it.

if `go-git` would allow sparse checkout we could only download the crate file, but the issue is pending there

olblak: What surprise me is even a repository that has already been 

First time repository download 
- `./bin/updatecli diff --config e2e/updatecli.d/success.d/crate.yaml --debug  289.71s user 22.11s system 128% cpu 4:03.55 total`

Second time once the repository is already cloned
- `./bin/updatecli diff --config e2e/updatecli.d/success.d/crate.yaml --debug  173.48s user 20.45s system 145% cpu 2:13.04 total
`

I am bit surprised that it doesn't go much faster once already downloaded.  Usually repositories that I work on are relatively small so I am never  affected by such slowness maybe I could do some optimization on Updatecli side.

I am testing a very simple config file

```
sources:
  test1:
    kind: cargopackage
    spec:
      package: syn 
  test2:
    kind: cargopackage
    spec:
      package: rand
```

olblak:
Excepted a few minor remark, I think this pullrequest is really great.
Once the default "clone" operation uses the Updatecli scm snippet so it clones the repository in the Updatecli temporary location and we ensure that we do not modify the spec data, I think it can be merged

Not in this pull request but a follow up effort would be to improve the git clone operation . I think some benchmarking is needed to speed up. I never realize how slow it could be for large repositories

pkg/plugins/resources/cargopackage/source.go
olblak: I am wondering if we can also leverage `$CARGO_HOME/registry` 
Something like:

1. If scm then use `scm.getDirectory()` if scm is specified
2. If indexDir use `index.Dir` if it's specified
3. If `$CARGO_HOME/registry` if it exists
4. Clone `$CARGO_HOME/registry` in `/tmp/updatecli`

olblak: I don't see how to improve the clone operation but the crates.io repository is constantly updated

pkg/core/pipeline/resource/main.go
```diff
@@ -60,6 +61,9 @@ func New(rs ResourceConfig) (resource Resource, err error) {
 	switch kind {
 	case "aws/ami":
 		return awsami.New(rs.Spec)
+	case "cargopackage":
+
+		return cargopackage.New(rs.Spec, rs.SCMID)
```
loispostula, line 66: @olblak I don't know if you are ok with using a `New` signature that is different than the other resource definition? By having `scmid` here, I can infer the proper index to use in the `source` method without having to rely on comparing the `workingdir` to the current dir

olblak: Changing this new signature is fine that being said I don't understand why there is a different behavior if it's a scm or not.

In the case of the source 
```func (cp CargoPackage) Source(workingDir string) (string, error) {```
workingDir is already correctly configured either to the scm directory or local path and it should only be used fi `cp.spec.IndexDir` and `cp.spec.IndexURL` is not set

Comment (loispostula) [reply pkg/core/pipeline/resource/main.go:66]: If we are in the case of using an `scm` then the `Source` function get the scm directory as input, `cp.spec.IndexDir` and `cp.spec.IndexURL` will not be set and we should use the `workingDir`.

If we are in the case of **not** using an `scm` but we want to use the default cargo index, then we will get the local path as input and `cp.spec.IndexURL` will not be set . 

The only way to differentiate the two scenario from within the `Source` function is then to compare that the `workingDir` is not the `localPath` and it seemed more hacky than to give the `scm` information in the `New` function

Comment (olblak) [reply pkg/core/pipeline/resource/main.go:66]: Thanks for the information, then I guess a boolean `isSCM` could make sense in the New signature

Comment (loispostula) [reply pkg/core/pipeline/resource/main.go:66]: So you would pefere `rs.SCMID != ""` than passing the scmId?

Comment (loispostula) [reply pkg/core/pipeline/resource/main.go]: done

File: pkg/plugins/resources/cargopackage/condition.go [hunk 9]
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	_, versions, err := cp.getVersions()
+	if err != nil {
+		return false, err
+	}
+
+	for _, v := range versions {
+		if v == versionToCheck {
+			fmt.Printf("%s release version '%s' available\n", result.SUCCESS, versionToCheck)
```
olblak, line 54:
```suggestion
			logrus.Infof("%s release version '%s' available\n", result.SUCCESS, versionToCheck)
```

pkg/plugins/resources/cargopackage/condition.go
```diff
@@ -0,0 +1,62 @@
...
+			return true, nil
+		}
+	}
+
+	fmt.Printf("%s Version %q doesn't exist\n", result.FAILURE, versionToCheck)
```
olblak, line 59:
```suggestion
	logrus.Infof("%s Version %q doesn't exist\n", result.FAILURE, versionToCheck)
```

pkg/plugins/resources/cargopackage/source.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/result"
+)
+
+// Source returns the latest npm package version
+func (cp CargoPackage) Source(workingDir string) (string, error) {
+	logrus.Infof("indexDir: %s, workingDir: %s", cp.indexDir, workingDir)
```
olblak, line 12:
```suggestion
	logrus.Debugf("indexDir: %s, workingDir: %s", cp.indexDir, workingDir)
```

olblak: Imho it's a debut information but I am open to suggestion

pkg/plugins/resources/cargopackage/condition.go
```diff
@@ -0,0 +1,62 @@
+package cargopackage
+
+import (
+	"errors"
+	"fmt"
```
olblak, line 5:
```suggestion
```
I don't think we need fmt in this file if we convert ftm.Printf to logrus.Printf

olblak:
Super excited to see this pullrequest.
I did some testing and it works great.
I am waiting for the tests to pass before merging

olblak (approved):

PR #1093, opened 2023-01-17
Support for npm private registry

loispostula:
Fix [1012](https://github.com/updatecli/updatecli/issues/1012)

Implement support of private npm registry

## Test

```shell
cp pkg/plugins/resources/npm
go test
```

## Additional Information

When working with a private registry, it is necessary to specify the registry URL and an authorization token. These details can be provided in the resource specification and will be used as the default registry for connecting.

The resource code also checks for the presence of an `npmrc` file. This file contains information about scoped packages and can be specified in the `spec.NpmrcPath` field or defaults to `$HOME/.npmrc`.

The code will first establish the default registry by using the provided URL and token or defaulting to the npm registry. It then parses the `npmrc` file to identify additional registries and determine which scoped packages should use them.

When searching for a package, the resource code checks if it is a scoped package and if the scope has a custom registry configured. If so, it uses that registry, otherwise, it uses the default registry.

I've also fleshed out the test to introduced a mocked registry client

### DIFF
--- a/pkg/plugins/resources/npm/condition_test.go
+++ b/pkg/plugins/resources/npm/condition_test.go
@@ -1,7 +1,6 @@
 package npm
 
 import (
-	"log"
 	"os"
 	"path/filepath"
 	"testing"
@@ -13,7 +12,7 @@ import (
 func TestCondition(t *testing.T) {
 	dir, err := CreateDummyRc()
 	if err != nil {
-		log.Fatal(err)
+		require.NoError(t, err)
 	}
 	defer os.RemoveAll(dir)
 	tests := []struct {
--- a/pkg/plugins/resources/npm/condition_test.go
+++ b/pkg/plugins/resources/npm/condition_test.go
@@ -112,10 +112,6 @@ func TestCondition(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := New(tt.spec)
-			if tt.expectedError {
-				assert.Error(t, err)
-				return
-			}
 			require.NoError(t, err)
 			if tt.mockedResponse {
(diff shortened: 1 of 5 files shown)


pkg/plugins/resources/npm/condition_test.go
```diff
@@ -1,19 +1,32 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestCondition(t *testing.T) {
+	dir, err := CreateDummyRc()
+	if err != nil {
+		log.Fatal(err)
```
olblak, line 16: Using log.Fatal is a bit brutal :D 
I would prefere `require.NoError(t, err)`

pkg/plugins/resources/npm/main.go
```diff
@@ -20,10 +27,14 @@ type Spec struct {
...
 	Version string `yaml:",omitempty"`
-	// Defines registry url
+	// URL defines the registry url (defaults to `https://registry.npmjs.org/`)
 	URL string `yaml:",omitempty"`
+	// RegistryToken defines the token to use when connection to the registry
+	RegistryToken string `yaml:",omitempty"`
 	// VersionFilter provides parameters to specify version pattern and its type like regex, semver, or just latest.
 	VersionFilter version.Filter `yaml:",omitempty"`
+	// NpmrcPath defines the path to the .npmrc file
+	NpmrcPath string `yaml:"npmrcPath,omitempty"`
```
olblak, line 37:
```suggestion
	NpmrcPath string `yaml:"npmrcpath,omitempty"`
```

We decided to only use lowercase for the spec.
this is enforced by the generated jsonschema that will convert any uppercase to lowercase.

olblak (approved):
I don't have access to a private NPM registry so I can't really test it but the code part looks great.
Thanks for the pullrequest.

PR #1065, opened 2023-01-01
Add NPM autodiscovery

olblak:
Signed-off-by: Olblak <me@olblak.com>

Dependson:
- [x]  Allows to specify complex shell https://github.com/updatecli/updatecli/pull/1067
- [x] Allows to restrict files monitored by the shell resource 

Add npm autodiscovery to bump any version specified as `dependencies` or `devDependencies` specified in a `package.json` 

Updatecli ignores version bump for the following scenarios
* any `package.json` in `node_modules` directories
* Any version constraint handled by NPM starting with "*", ">","<","~" 

### Manifest
<details><summary>Click to expand</summary>

```

########################################
# [NPM] BUMP @MDI/FONT PACKAGE VERSION #
########################################

name: '[NPM] Bump @mdi/font package version'
pipelineid: b1f2a3e1656341a003847d1b725c14081e2152f27166a31bd668902bc7814d15
sources:
    npm:
        name: Get @mdi/font package version
        kind: npm
        spec:
            name: '@mdi/font'
            versionfilter:
                kind: semver
                pattern: '>=5.9.55'
targets:
    npm:
        name: Bump @mdi/font package version
        kind: json
        spec:
            file: package.json
            key: dependencies.@mdi/font
        sourceid: npm

```

</details>

### Pipeline Run

<details><summary>Click to expand</summary>

```
 ~/Projects/Updatecli/app-dashboard │ ../updatecli/bin/updatecli diff --experimental                                                                                       
Experimental Mode Enabled


+++++++++++
+ PREPARE +
+++++++++++


SCM repository retrieved: 0


++++++++++++++++++
+ AUTO DISCOVERY +
++++++++++++++++++



#######################
# LOCAL AUTODISCOVERY #
#######################

NPM
====
1 manifests identified
Manifest detected: 1

---

=> Total manifest detected: 1



++++++++++++
+ PIPELINE +
++++++++++++



#######################
# LOCAL AUTODISCOVERY #
#######################


########################################
# [NPM] BUMP @MDI/FONT PACKAGE VERSION #
########################################


SOURCES
=======

npm
---
Searching for version matching pattern ">=5.9.55"
✔ Version 7.1.96 found for package name "@mdi/font"


TARGETS
========

npm
---

**Dry Run enabled**

⚠ Key 'dependencies.@mdi/font', from file 'package.json', is incorrectly set to 5.9.55 and should be 7.1.96'

=============================

REPORTS:


✔ Local AutoDiscovery:

[...]

PR #1094, opened 2023-01-17
Add Cargo package autodiscovery

loispostula:
Implement cargo package auto-discovery

Dependson:

- [x] Support autodiscovery of private crates
- [x] Support autodiscovery of crate specifying their version inline

### Support autodiscovery of crate specifying their version inline

Cargo dependencies can be specified in two ways:
```toml
[dependencies]
rand = "0.1.0"
``` 
or 
```toml
[dependencies]
[dependencies.rand]
version = "0.1.0"
```

Due to limitation in the current `toml` parsing capabilities, it's not possible to perform such complex queries on the `Cargo.toml ` files. This means that autodiscovery will not work for dependencies not specifies using the full spec

## Test

```shell
cd pkg/plugins/autodiscovery/cargo
go test
```

### Potential improvement

* Allow to restrict version update to a specific type such as only major/minor/patch update e if semver is used
* Target executing ``cargo generate-lockfile` in dry-run mode shouldn't not change the file `Cargo.lock`
* Leverage [google/osv.dev](https://github.com/google/osv.dev) to detect CVEs


olblak:
> Due to limitation in the current toml parsing capabilities, it's not possible to perform such complex queries on the Cargo.toml files. This means that autodiscovery will not work for dependencies not specifies using the full spec

Do you have example of queries that do no work?
Because the two example that you provide are working.

```
sources:
  example1:
    kind: toml
    spec:
      file: '/tmp/example2.toml'
      key: 'dependencies.rand'

  example2:
    kind: toml
    spec:
      file: '/tmp/example.toml'
      key: 'dependencies.rand.version'
```

pkg/core/engine/main.go
```diff
@@ -504,9 +504,9 @@ func (e *Engine) LoadAutoDiscovery(defaultEnabled bool) error {
 
 			manifest.PipelineID = pipelineID
 
-			if sc != nil {
-				manifest.SCMs = make(map[string]scm.Config)
-				manifest.SCMs[p.Config.Spec.AutoDiscovery.ScmId] = *sc
+			manifest.SCMs = make(map[string]scm.Config)
+			for scmId, sc := range p.SCMs {
+				manifest.SCMs[scmId] = *sc.Config
```
loispostula, line 509: @olblak This is the biggest change in this pr in term of complete workflow, as instead of only adding the initial manifest scm to the generate manifests, I forward all scm spec to the generated manifests

e2e/updatecli.d/warning.d/autodiscovery/cargo/example.1.yaml
```diff
@@ -0,0 +1,23 @@
+name: "Cargo compose autodiscovery using git scm"
+scms:
+  default:
+    kind: git
+    spec:
+      url: https://github.com/loispostula/cargo-lab.git
```
olblak, line 6: I am starting to wonder if we shouldn't create a new Github organization named `updatecli-test` to store those kind of repsitories.

pkg/plugins/autodiscovery/cargo/dependencies.go
```diff
@@ -0,0 +1,207 @@
...
+		TargetKey                  string
+		ScmID                      string
+		WithRegistry               bool
+		RegistrySCMID              string
+		RegistryRootDir            string
+		RegistryURL                string
+		RegistryAuthToken          string
+		RegistryHeaderFormat       string
+	}{
+		ManifestName:               fmt.Sprintf("Bump %s %q for \"%s\" crate", dependencyType, dependency.Name, crateName),
```
olblak, line 89:
```suggestion
		ManifestName:               fmt.Sprintf("Bump %s %q for %q crate", dependencyType, dependency.Name, crateName),
```

%q surround the string value with quotes

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -0,0 +1,59 @@
+package cargo
+
+const (
+	// dependencyManifest is the Go template used to generate
+	// the Helm chart manifests specific for Helm dependencies
```
olblak, line 5:
```suggestion
	// the manifests to update the cargo file
```

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -0,0 +1,59 @@
...
+conditions:
+  {{ .ConditionID }}:
+    name: 'Ensure Cargo chart named "{{ .DependencyName }}" is specified'
+    kind: 'toml'
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      file: '{{ .File }}'
+      Query: '{{ .ConditionQuery }}'
```
olblak, line 45:
```suggestion
      query: '{{ .ConditionQuery }}'
```

keys should all be lowercase

olblak: while the uppercase key should work. The jsonschema generated would display would consider it as an error

pkg/plugins/autodiscovery/cargo/utils.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+	"io/fs"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/text"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/dasel"
+)
+
+// searchChartFiles search, recursively, for every files named Cargo.toml from a root directory.
```
olblak, line 13:
```suggestion
// findCargoFiles search, recursively, for every files named Cargo.toml from a root directory.
```

pkg/plugins/autodiscovery/cargo/utils.go
```diff
@@ -0,0 +1,118 @@
...
+func findCargoFiles(rootDir string, validFiles []string) ([]string, error) {
+	manifestsFiles := []string{}
```
olblak, line 15:
```suggestion
	cargoFiles := []string{}
```

olblak: Maybe just nitpicking

pkg/plugins/autodiscovery/cargo/utils.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Debugf("%d chart(s) found", len(manifestsFiles))
+	for _, foundFile := range manifestsFiles {
+		chartName := filepath.Base(filepath.Dir(foundFile))
```
olblak, line 41:
```suggestion
		cargoFile := filepath.Base(filepath.Dir(foundFile))
```

pkg/plugins/autodiscovery/cargo/utils.go
```diff
@@ -0,0 +1,118 @@
...
+		logrus.Debugf("    * %q", chartName)
```
olblak, line 42:
```suggestion
		logrus.Debugf("    * %q", cargoFile)
```

pkg/plugins/autodiscovery/cargo/utils.go
olblak: Then you would have to change all occurrence of `manifestsFile` to `cargoFiles`

e2e/updatecli.d/warning.d/autodiscovery/cargo/example.1.yaml
loispostula: That would work well I think

pkg/plugins/resources/cargopackage/spec.go
```diff
@@ -1,18 +1,15 @@
...
+	"github.com/updatecli/updatecli/pkg/plugins/utils/cargo"
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
 // Spec defines a specification for a "dockerimage" resource
 // parsed from an updatecli manifest file
 type Spec struct {
-	// [S][C] IndexUrl specifies the url of the index to use to check version
-	// If set, package version will be checked against the API
-	IndexUrl string `yaml:",omitempty"`
```
olblak, line 12: This is a breaking change that could easily be avoided with a deprecation approach
We try to carefully introduce new "user" facing parameters as we want to avoid as much as possible breaking changes.

Something like 

```
`IndexUrl string `yaml:",omitempty" jsonschema:"-"``
```
Means that the filed won't appear in the generated jsonschema and so will return an error if the indexurl parameter is used.

Then in the new object you could have something like 

```
if spec.IndexURL != "" {
  switch spec.Registry.URL != {
    case true:
      logrus.Warningf("Registry.URL and IndexURL are mutually exclusive, unset indexurl")
      spec.IndexURL = ""
    case false:
      logrus.Warningf("indexurl is deprecated in favor of registry.url")
      spec.IndexURL = ""
  }
}

By emptying the deprecated field in favor of the new field, the following command will automatically update the old manifest

`updatecli manifest upgrade`

File: pkg/core/engine/main.go (hunk 1 above)
Comment (olblak) [reply pkg/core/engine/main.go:509]: Interesting, I am curious what's the motivation?

Comment (loispostula) [reply pkg/core/engine/main.go:509]: The recomended way to parse cargo registries is to pull git repositories containing the index. (In https://github.com/updatecli/updatecli/pull/1081 I've implemented API support because the official cargo registry is big and the clone operation takes about 1 minute).

This means that if we want to run autodiscovery in a git repo (leveraging the current scmid process of autodiscovery), I cannot use a cargo registry in a git scm without this changes

**Review by olblak — CHANGES_REQUESTED** [review]
I made some comments and spot some breaking changes that should be avoided

File: pkg/plugins/autodiscovery/cargo/main.go [hunk 11]
```diff
@@ -0,0 +1,71 @@
...
+// Spec defines the Cargo parameters.
+type Spec struct {
+	// RootDir defines the root directory used to recursively search for Cargo.toml
+	RootDir string `yaml:",omitempty"`
+	// Ignore specifies rule to ignore Cargo.toml update.
+	Ignore MatchingRules `yaml:",omitempty"`
+	// Only specify required rule to restrict Cargo.toml update.
+	Only MatchingRules `yaml:",omitempty"`
+	// Auths provides a map of registry credentials where the key is the registry URL without scheme
+	Registries map[string]cargo.Registry `yaml:",omitempty"`
```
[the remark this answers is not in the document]
loispostula: Yes it is used here: https://github.com/updatecli/updatecli/blob/cargo-autodiscovery/pkg/plugins/autodiscovery/cargo/dependencies.go#L59-L63

This is the way to specify custom registries (private) for crates

loispostula:
I've update according to the review, I think the remaining items are:

- [ ] Is it ok to provide all scmid
- [ ] Should we move the test repos to an `updateclitest` org
- [ ] Deprecation and `cargo manifest upgrade` changes. I've implemented those in the New resource code.

pkg/core/engine/main.go
olblak: Thanks, the change is legit and I can't think of a limitation excepted being sure that there is no collision between the scmid used by the registry versus the one specified in the autodiscovery manifest

e2e/updatecli.d/warning.d/autodiscovery/cargo/example.1.yaml
olblak: I just created https://github.com/orgs/updatecli-test 
That would be great if you could migrate the repository over there as it will be more maintainable

pkg/plugins/utils/cargo/main.go
```diff
@@ -0,0 +1,19 @@
+package cargo
+
+type InlineKeyChain struct {
+	// [A][S][C] Token specifies the cargo registry token to use for authentication.
+	Token string `yaml:",omitempty"`
+	// [A][S][C] HeaderFormat specifies the cargo registry header format to use for authentication (defaults to `Bearer`).
+	HeaderFormat string `yaml:"headerFormat,omitempty"`
+}
+
+type Registry struct {
```
olblak, line 10: I like this approach but maybe, we miss a validation function that check that we didn't specified at the same time an URL, a rootdir, or a scmid.

loispostula: Would that be in a `func (s Spec) Validate() (bool)` function?

olblak:
Congrats on the pullrequest, I tried it on `github.com/kubewarden/kwctl` and I only faced one issue.
It couldn't handle this line https://github.com/kubewarden/kwctl/blob/ad831c472db07385e304e0338b159741da5b9a1d/Cargo.toml#L25

We should find a way to skip dependencies that can't be handle such as if version is empty then skip the dependency.

Regarding the failing dependency, we could generate an traditional type of manifest that rely on a gittag for the source and the target would have to update the tag.
That being said, it's ok to handle this scenario later on, in a separated pullrequest. 

I think the current pullrequest face the same issue than my PR on https://github.com/updatecli/updatecli/pull/1065#issuecomment-1370197660
It needs a way to update the file `Cargo.lock`
This will probably be simplified once my shell refactoring is merged

pkg/plugins/utils/cargo/main.go
olblak: I think I would create a function 
`func (r Registry) Validate() (bool)` here
Then call that function in the `New` function

loispostula:
Thanks for the extensive review, I had not foreseen the `Cargo.lock` challenge, I've added the mechanism to generate it, in the same way as in https://github.com/updatecli/updatecli/pull/1065 .

Regarding the issue for the git dependency, I've also added a placeholder for git and local path dependency and am currently ignore them.

olblak:
> Thanks for the extensive review, I had not foreseen the `Cargo.lock` challenge, I've added the mechanism to generate it, in the same way as in #1065 .

I didn't foreseen neither on the NPM pullrequest until very late.
I need to finalize https://github.com/updatecli/updatecli/pull/1071 which shouldn't take long now  

> Regarding the issue for the git dependency, I've also added a placeholder for git and local path dependency and am currently ignore them.

Make sense once thing at a time

loispostula:
@olblak Thanks for #1071 and #1067 

I've added thoses key to the manifest, I also add to make sure that the shell resource inherited the path environment variable. And it's now working fine

olblak:
I'll run some tests over the weekend

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -0,0 +1,81 @@
...
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      file: '{{ .File }}'
+      query: '{{ .ConditionQuery }}'
+    sourceid: '{{ .ExistingSourceID }}'
+targets:
+  {{ .TargetID }}:
+    name: 'Bump crate dependency "{{ .DependencyName }}" for crate "{{ .CrateName }}"'
```
olblak, line 49: Having the version information in the title would be usefull when creating the pullrequest.
Something like 

```suggestion
    name: 'Bump crate dependency "{{ .DependencyName }}" to {{ source ".SourceID" }}"'
```

`{{ source ".SourceID" }}` must be correctly escaped as it needs to be update at run time.
In the NPM pullrequest, I generated the name from the golang code instead of the go template to make it more readable

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -0,0 +1,81 @@
...
+    name: Update Cargo lockfile Cargo.lock
+    dependson:
+      - {{ .TargetID }}
+    disablesourceinput: true
+    kind: shell
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{ end }}
+    spec:
+      command: cargo generate-lockfile
```
olblak, line 69: Is there a way to run this command in dry-run mode?

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -0,0 +1,81 @@
...
+targets:
+  {{ .TargetID }}:
+    name: 'Bump crate dependency "{{ .DependencyName }}" for crate "{{ .CrateName }}"'
+    kind: 'toml'
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      file: '{{ .TargetFile }}'
+      key: '{{ .TargetKey }}'
```
olblak, line 56: An issue I faced on the npm pullrequest is the situation where I want the dependency "~4.0". The lockfile contains "4.0.0" but "4.0.10" is available. I don't want to update the manifest from "~4.0" to "4.0.10", I only update the lockfile.

I am wondering how it works with cargo

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
olblak: Maybe something like this  cargo update --dry-run $DRY_RUN <package_name>

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
olblak: It appears you initial approach was already good

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
olblak: You initial approach by using `cargo generate-lockfile` seems to be the best one.
My only drawback is when running updatecli in diff mode but I don't think there is anything we could do :(

pkg/plugins/utils/cargo/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"os/exec"
+
+	"github.com/sirupsen/logrus"
+)
+
+type InlineKeyChain struct {
+	// [A][S][C] Token specifies the cargo registry token to use for authentication.
+	Token string `yaml:",omitempty"`
+	// [A][S][C] HeaderFormat specifies the cargo registry header format to use for authentication (defaults to `Bearer`).
+	HeaderFormat string `yaml:"headerFormat,omitempty"`
```
olblak, line 14:
```suggestion
	HeaderFormat string `yaml:"headerformat,omitempty"`
```
All user facing configuration must be lowercase. It's a jsonschema constraint otherwise the IDE returns an error.

olblak:
@loispostula In my last commit, I disable updating the Cargo.toml if it's not strict semver. So 1.0.0 should be updated to >=1.0.0 while "1.0" should only update "Cargo.lock" by running `cargo generate-lockfile`

I have mixed feeling about overidding version constraint in Cargo.toml (and I am open for discussion)

olblak (approved):

PR #1095, opened 2023-01-17
Test updatecli on windows && macos runner

olblak:

Add Windows runner to test Golang

## Test
/

### Tradeoff

/

### Potential improvement
/

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -20,6 +20,7 @@ jobs:
         os: [ubuntu-20.04, windows-latest, macos-latest]
 
     steps:
+
       # https://github.com/actions/setup-go
       - name: Set up Go
         uses: actions/setup-go@v3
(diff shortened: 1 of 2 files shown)


olblak:
Well as I initially thought, it will take more time to fix the tests on Windows.
The good thing I now have access to a windows machine

dduportal:
> Well as I initially thought, it will take more time to fix the tests on Windows. The good thing I now have access to a windows machine

Important note about this change:

- Unit test does NOT need to be run on all platforms. If it fails on a platform, then it has to be either refactored to NOT depend on a platform or moved as an end to end test. Even integration tests should not be dependent otherwise it means both them and the code they test is fragile.
- WDYT about running only end to end tests on different paltforms, and staging the built `updatecli` binary between the build machine and the others
  - No need to setup golang and build on each machine
  - spend less time and machine cycle on tests that would have low value

olblak:
Thanks for your feedback 

> WDYT about running only end to end tests on different paltforms, and staging the built updatecli binary between the build machine and the others

Yep that's something which I am considering and **should** be easier to implement.

> Unit test does NOT need to be run on all platforms. If it fails on a platform, then it has to be either refactored to NOT depend on a platform or moved as an end to end test. Even integration tests should not be dependent otherwise it means both them and the code they test is fragile.
 
> If it fails on a platform, then it has to be either refactored to NOT depend on a platform or moved as an end to end test.

I agree with you, but running the tests on all platform is a good way to catch tests that should be refactored to not be platform specific.

I now have access to Windows machine  so I am planning to do some experimentation

olblak:
Closing for now, I managed to build a Dev environment on a windows machine so I'll rework this effort from there

PR #1102, opened 2023-01-18
[gittag] Use `spec.path` instead of working dir

### LINKED ISSUE #1101: Gittag spec.path is not working

### Current Behavior

When the `spec.path` is specified the current working directory is still used.

```yaml
sources:
  gitExample1:
    name: "Get Latest updatecli test"
    kind: gittag
    spec:
      path: "./git_repository_path/"
```

### Expected Behavior

The tags of the specified repository should be used.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- updatecli:
```

### Anything else?

_No response_

enrichman:
Fix #1101 

The `spec.path` value was updated with the `workingDir` if not specified, but still the working directory was used.

### DIFF
--- a/pkg/plugins/resources/gittag/source.go
+++ b/pkg/plugins/resources/gittag/source.go
@@ -19,7 +19,7 @@ func (gt *GitTag) Source(workingDir string) (string, error) {
 		return "", err
 	}
 
-	tags, err := gt.nativeGitHandler.Tags(workingDir)
+	tags, err := gt.nativeGitHandler.Tags(gt.spec.Path)
 
 	if err != nil {
 		return "", err


dduportal (approved):
Good catch, thanks for the fix!

loispostula:
Would not that break `scm` capability of `gittag`, as in that case `gt.spec.Path` will be empty whereas `workingDir` will be the path of the scm repo?

enrichman:
> Would not that break `scm` capability of `gittag`, as in that case `gt.spec.Path` will be empty whereas `workingDir` will be the path of the scm repo?

The check done before will prevent this, because is updating the `gt.spec.Path` as the `workingDir` if empty:`

```go
if len(gt.spec.Path) == 0 && len(workingDir) > 0 {
	gt.spec.Path = workingDir
}
```

enrichman:
@dduportal do you know when this could be released? :slightly_smiling_face: 

I would like to try it out, and close an issue that we have!

Issue #1133, opened 2023-02-02
When using `git` scm with `push=false` only the last target is left committed 


mcwarman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using multiple targets the changes get lost as the clone is reset to `.spec.branch` every time, when `--push=false` is set.

### Expected Behavior

All pipeline changes should be present.

#### Showing `git` not working

1. Apply the following patch:

```diff
diff --git a/e2e/updatecli.d/success.d/json.yaml b/e2e/updatecli.d/success.d/json.yaml
index b784b97..a00af52 100644
--- a/e2e/updatecli.d/success.d/json.yaml
+++ b/e2e/updatecli.d/success.d/json.yaml
@@ -1,5 +1,12 @@
 name: Basic Json Example
 
+scms:
+  default:
+    kind: git
+    spec:
+      url: https://github.com/updatecli/updatecli.git
+      branch: main
+
 sources:
   local:
     name: Get value from json
@@ -56,6 +63,7 @@ targets:
     name: Test value from json
     kind: json
     sourceid: local
+    scmid: default
     spec:
       file: pkg/plugins/resources/json/testdata/data.json
       key: firstName
@@ -65,9 +73,9 @@ targets:
     name: Test value from json
     kind: json
     sourceid: local
+    scmid: default
     spec:
       files:
-        - pkg/plugins/resources/json/testdata/data.json
         - pkg/plugins/resources/json/testdata/data.2.json
       key: firstName
       value: John

```

2. Execute test

```shell
updatecli apply --config ./e2e/updatecli.d/success.d/json.yaml --push=false
```

3. Check results:

```output
git log --oneline -n 3
c48ce4d0 (HEAD -> main) chore: Update key "firstName" from file "pkg/plugins/resources/json/t...
9765d1db (origin/main) fix: jenkins plugin comment
58c43036 Add `spec.files` for Condition and Target of `kind` `yaml` (#650)
```

### Environment

```markdown
- OS: Ubuntu-20.04 (WSL)
- updatecli: v0.43.0
```

olblak:
Very early in the lifecycle of Updatecli we needed both workflow

1. Being able to commit directly to a head branch 
2. Being able to commit to a working branch

Back then, we decided to use the git scm type for the case one and github for the case two.
That decision didn't age very well.

Both @dduportal and I want to solve this but we took two different approaches 
Mine was to add a new boolean `disableworkingbranch` to all scm
[...]

PR #1170, opened 2023-02-19
Target using SCM should be indempotent

### LINKED ISSUE #1162: github/pullrequest Auth failure poisons retries

### Current Behavior

When the `github/pullrequest` action is run with an incorrect username, the action will fail with the message

```
ERROR: something went wrong in target "bumpKo" : "authorization failed"
```

If the username is fixed, and the action rerun, the action will fail with the message:

```
Pipeline ... failed
Skipping due to:
        action stage:   "Head sha can't be blank, Base sha can't be blank, No commits between BRANCH and updatecli_SOME_SHA, Head ref must be a branch"
```

To address this, users must delete a temp folder that updateCLI has created locally. This
1. is suprising
2. requires a stronger mental model of the internal operation of updateCLI than users are likely to have
3. is not idempotent: reruns of the action with the incorrect username will yield different error messages; first about auth and then about no commits of difference

### Expected Behavior

On fixing the username and rerunning the update, the update should succeed.

### Steps To Reproduce

- Define a manifest with an action with `kind: github/pullrequest`
- In scms, define a `username` which does not match the (valid) `token`
- Run the action.
- Fix the `username`.
- Rerun the action

### Environment

```markdown
- OS: OSX 13.1 (ARM) 
- updatecli:
  - Application:    0.44.2
  - Golang     :    1.19.5 linux/amd64
  - Build Time :    2023-02-13T18:01:23Z
```

### Anything else?

_No response_

olblak:
## Requirement

- [x] https://github.com/updatecli/updatecli/issues/1133

## Description
Fix #1162 

At the moment the target execution combined with scm is not idempotent. 

Running first `updatecli apply --config manifest --push=false` will clone the repository, update the files, but not publish the changes, as `--push=false` is specified.
If we run a second time the same command without `--push=false`, Updatecli will not change any files as local files are already up to date and won't push the changes as nothing change in the current pipeline execution.

They are a few workaround which must result in deleting the local directory.  

I propose in this pullrequest to check for each target execution if the current working branch such as `updatecli_xxx` has the same commit hash than the source branch such as `main` and if the remote of `updatecli_xxx` exist.

The goal is to identify if the current working breach has left over changes that need to be publish

## Test

```shell
cp <to_package_directory>
go test
```

### Potential improvement

* Add a flag to disable scm behavior
* https://github.com/updatecli/updatecli/pull/1174 
* --debug should list commits that need to be push

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -38,7 +38,7 @@ body:
     description: |
       examples:
         - **OS**: Ubuntu 20.04
-        - **updatecli**: v0.44.2
+        - **updatecli**: v0.44.4
     value: |
         - OS:
         - updatecli:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,6 +56,7 @@ jobs:
             - name: GoReleaser
               if: success()
               env:
+                AUR_KEY: ${{ secrets.AUR_KEY }}
                 COSIGN_PWD: ${{ secrets.COSIGN_PWD }}
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                 HOMEBREW_TAP_GITHUB_TOKEN: ${{ steps.generate_homebrew_token.outputs.token }}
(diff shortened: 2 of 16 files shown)


olblak:
After additional thought, this https://github.com/updatecli/updatecli/issues/1133 should be fixed by removing the reset operation done during the git checkout

PR #1177, opened 2023-02-26
feat(core/config): make Sprig functions available in config template

Croissong:
This change makes all functions from the [Sprig template library](https://masterminds.github.io/sprig/) available in the manifest.

Coming from Helm/Helmfile, I was missing the various utility functions provided by Sprig while building my config. 
For example, my personal config I am iterating over a  `packages` map and would like to *merge* each package config with a default (showing the use of `mergeOverwrite` from Sprig):
```yaml
# values.yml

defaults:
  trimPrefix: true

packages:
  sheldon:
    owner: rossmacarthur
    trimPrefix: false

  sttr:
    owner: abhimanyu003

  updatecli:
    owner: updatecli
```

```yaml
# updatecli.yaml

sources:
  {{ range $name, $package := . packages }}

  # set defaults
  {{ $package = mergeOverwrite .defaults $package }}
  ...
```
## Test
I slightly refactored `pkg/core/config/template.go` to be able to mock the filesystem interactions, and I added some basic tests for the templating functionality.

## Docs
I created https://github.com/updatecli/website/pull/672 to mention Sprig in the docs.


olblak:
Waouw that's something I wanted so much \o/
Let me run some tests

pkg/core/config/template.go
```diff
@@ -59,7 +62,10 @@ func (t *Template) New(content []byte) ([]byte, error) {
 	// Deepmerge is not supported so a secrets override unencrypted values
 	templateValues := Merge(t.Values, t.Secrets)
 
-	tmpl, err := template.New("cfg").Funcs(funcMap).Parse(string(content))
+	tmpl, err := template.New("cfg").
+		Funcs(funcMap).
```
olblak, line 66: I guess here order matter. I was wondering which one between sprig and updatecli map should have the priority. I am thinking in a long period where sprig introduce a function map that conflict with Updatecli.

Imho it's better to have updatecli function coming first and if a user in the futur want to use a new sprig function, then we'll revisit that decision.

Maybe all we need here is a comment

Croissong: Good catch. Thinking about it, I would actually put `Funcs(funcMap)` second, so that is always takes precedence if Sprig adds a function with the same name.

olblak:
Thanks @Croissong for the pullrequest. It looks great. Once merged I'll take a few days to test the main branch, before releasing 0.45 as I did a few changes around git operation which may introduce regression and I want to be extra safe

pkg/core/config/template.go
olblak: Do you know how it is interpreted? Is the first or last one considered? 

The way I see it is if later on, sprig add a new function that conflict with Updatecli, then we can control how we deprecate adapt without breaking existing manifest

Because if sprig is used first then all existing manifest will stop working if sprig introduce a function that conflict with Updatecli while on the other side, we can plan so transition without breaking existing manifest.

Or we consider we have good test coverage for updatecli function and we'll catch the error if we bump the sprig library version :trollface:

Croissong: Subsequent `Funcs` calls will override any existing functions:
https://github.com/golang/go/blob/master/src/text/template/template.go#L182

So if we add `Funcs(funcMap)` last, it will always take precedence. I agree that this is the safest option.

olblak: Thanks for your reactivity

PR #1187, opened 2023-02-28
fix yaml resource source filepath

olblak:
Fix #XXX

While automating a file update, I spot a regression where the yaml resource wouldn't use the correct filepath when used with scm.

## Test

I updated the e2e tests to catch this issue but it's worth mentioning they will only pass once the current pullrequest is merged on the main branch

### Potential improvement

* A source pipeline shouldn't set by default the current working directory. It's very confusing in resource plugins. https://github.com/updatecli/updatecli/blob/427cb7795834ddd3684d8061df81c6035de0906b/pkg/core/pipeline/source/main.go#L52 I would need to investigate the initial intend and then review all plugins as I first noticed this behavior in the cargo package plugin and now in the yaml

### DIFF
--- a/e2e/updatecli.d/success.d/yaml/duplicate.yaml
+++ b/e2e/updatecli.d/success.d/yaml/duplicate.yaml
@@ -1,31 +1,25 @@
-name: Test YAML resource
-
-scms:
-  default:
-    kind: git
-    spec:
-      url: https://github.com/updatecli/updatecli.git
+name: Test duplicated YAML resource
 
 sources:
   scenario1:
     name: Basic yaml source
     kind: yaml
     spec:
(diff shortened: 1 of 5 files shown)


olblak:
Also while running some tests, I noticed that a key not found in a yaml will not return an error

lemeurherve:
> Also while running some tests, I noticed that a key not found in a yaml will not return an error
[main/pkg/plugins/resources/yaml/source.go#L63L67](https://github.com/updatecli/updatecli/blob/main/pkg/plugins/resources/yaml/source.go?rgh-link-date=2023-02-28T12%3A03%3A18Z#L63L67)
>
> I can't remember if it's on purpose. @dduportal @hervelemeur maybe you recall some decision around that

No idea, sorry.

Issue #1190, opened 2023-02-28
No changelog in the PR request


alexagriffith:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

(redacting some stuff with xx since private) 
Right now, when I run updatecli it prints out the changes that need to be made with a changelog like 

```
################################
# ROLES-XX-BASE-QA.TPL #
################################


SOURCES
=======

roles-xx
----------------
Searching for version matching pattern "latest"
✔ Github Release version "1.22.16.7" found matching pattern "latest"


CHANGELOG:
----------

Release published on the 2023-02-15 00:13:43 +0000 UTC at the url https://github.com/xx/roles-xx/releases/tag/1.22.16.7

View the [build log on Jenkins](https://xx-build.jaas.dev.xx.com/job/xxx).
## Changelog since 1.22.16.6

- [`acd41c6`](https:/github.com/xx): [ENG3DISRE-2672] update configs (#1087) (#1089) by xx[bot] (20983+xx[bot]@users.github)


---

*The above changelog was generated automatically by [Jaazy-Jeff](https://github.com/ds/jaazy-jeff).*


TARGETS
========

roles-kubernetes-qa
-------------------
✔ Content from file "/var/folders/gl/9dxcv29j0f781mb8dmjhjg2h0000gn/T/updatecli/github/qa_base/xx.yaml" already up to date
✔ All contents from 'file' and 'files' combined already up to date


ACTIONS
========


Pull Request available at:

        https://github.com/xx/pull/277
```

so I expect to see this changelog in the changelog field

but instead every PR has a changelog that is empty like 

![Screen Shot 2023-02-28 at 1 00 43 PM](https://user-images.githubusercontent.com/43711982/221941133-ed39afcd-6388-47f4-bdbe-8abc311bdeeb.png)

### Expected Behavior

changelog populated in the auto PR 

### Steps To Reproduce

my action setup is like 

```
actions:
  default:
    title: Bump {{$name}} to {{ source $name }} on QA
    kind: github/pullrequest
    spec:
      automerge: false
      draft: false
      mergemethod: squash
      labels:
        - "chore"
    scmid: default
 ```

### Environment

```markdown
- OS:
- updatecli: v0.36.0
```

lemeurherve:
Related/duplicate of https://github.com/updatecli/updatecli/issues/610

Also related: https://github.com/updatecli/updatecli/issues/159

alexagriffith:
[...]

Issue #914, opened 2022-10-06
Unable to create separate branches/PRs for different targets/scms/PRs


JayDoubleu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Updatecli will push each target changes as commits into a single branch and open single pull request.

### Expected Behavior

Updatecli should push each target changes as single commit into a separate branch and open a pull request for each of the targets.

### Steps To Reproduce

1. `git clone git@github.com:JayDoubleu/updatecli_pr_test.git`
2. `cd updatecli_pr_test`
3. `export UPDATECLI_GITHUB_ACTOR="xxx"`
4. `export UPDATECLI_GITHUB_TOKEN="ghp_xxx"`
5. `updatecli apply --debug updatecli.yaml --clean --debug`

### Environment

```markdown
- OS: golang 1.19.1 linux/amd64
- updatecli: 0.33.3
```

olblak:
Thanks for the reporting this issue, your feedback is very valuable.
We are currently thinking about refactoring the SCM part but I must admit we don't have any ETA
We gathered our thoughts in https://github.com/updatecli/updatecli/issues/465

For now , we use one manifest per pipeline such as in [updatecli.d](https://github.com/updatecli/updatecli/tree/main/updatecli/updatecli.d)

alexagriffith:
just ran into this issue - would be great to have this feature

alexagriffith:
any suggested workaround that doesn't involved many files? I have a list of about 20 version bumps I do in one repo for a YAML file to deploy clusters - its not very helpful for me to have one PR for many updates, I need them to be separate because they need to be merged one at a time

alexagriffith:
> 

this is at the top of your .tpl? or is this a bash file that generates multiple tpl files?

alexagriffith:
> 

can you show me what your bash script looks like? that would help me a lot. thank you for helping me on this!

olblak:
Thanks @JayDoubleu for sharing your workaround. As far as I know, at the moment it's the only one that I am aware of. So using something like 

```
name: Example
pipelineid: '{{ requiredEnv "CURRENT_ITER_ENVIRONMENT" }}'
scms:
   ...
sources:
   ...
```

Then you could have a bash script such as 

```
  export CURRENT_ITER_ENVIRONMENT=xxx
  updatecli apply --config updatecli.d/manifest.yaml

  export CURRENT_ITER_ENVIRONMENT=yyy
[...]

PR #1201, opened 2023-03-02
add more information for disableinputsource

alexagriffith:
Adding a little more information so users understand the impact of not being able to retrieve a source value in the target. 

relates to #1190

### DIFF
--- a/pkg/core/pipeline/target/main.go
+++ b/pkg/core/pipeline/target/main.go
@@ -39,7 +39,6 @@ type Config struct {
 	ReportBody string `yaml:",omitempty"`
 	// ! Deprecated - please use all lowercase `sourceid`
 	DeprecatedSourceID string `yaml:"sourceID,omitempty" jsonschema:"-"`
-	// disablesourceinput disables the mechanism to retrieve a default value from a source. For example, if true, 
 	// disablesourceinput disables the mechanism to retrieve a default value from a source. For example, if true, source information like changelog will not be accessible for a github/pullrequest action.
 	DisableSourceInput bool `yaml:",omitempty"`
 	// sourceid specifies where retrieving the default value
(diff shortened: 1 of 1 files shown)


pkg/core/pipeline/target/main.go
```diff
@@ -39,11 +39,11 @@ type Config struct {
 	ReportBody string `yaml:",omitempty"`
 	// ! Deprecated - please use all lowercase `sourceid`
 	DeprecatedSourceID string `yaml:"sourceID,omitempty" jsonschema:"-"`
-	// disablesourceinput disables the mechanism to retrieve a default value from a source.
+	// disablesourceinput disables the mechanism to retrieve a default value from a source. For example, if true, 
+	// source information like changelog will not be accessible for a github/pullrequest action.
```
olblak, line 43:
```suggestion
	// disablesourceinput disables the mechanism to retrieve a default value from a source. For example, if true, source information like changelog will not be accessible for a github/pullrequest action.
```

olblak: According the library used to generate the jsonschema (https://github.com/invopop/jsonschema)
Description is generated based on the line before

pkg/core/pipeline/target/main.go
```diff
@@ -39,11 +39,11 @@ type Config struct {
 	ReportBody string `yaml:",omitempty"`
 	// ! Deprecated - please use all lowercase `sourceid`
 	DeprecatedSourceID string `yaml:"sourceID,omitempty" jsonschema:"-"`
-	// disablesourceinput disables the mechanism to retrieve a default value from a source.
+	// disablesourceinput disables the mechanism to retrieve a default value from a source. For example, if true,
```
olblak, line 42:
```suggestion
```

olblak:
Thank you @alexagriffith , I am currently looking to https://github.com/updatecli/updatecli/issues/914 and your comment resonate very well with what I am currently implementing

PR #1234, opened 2023-03-19
fix: gitlab url with https prefix

vbehar:
don't prefix the gitlab url with the `https://` prefix if it is already there...

### DIFF
--- a/pkg/plugins/resources/gitlab/client/main.go
+++ b/pkg/plugins/resources/gitlab/client/main.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/plugins/resources/gitlab/client/main.go
```diff
@@ -13,7 +12,7 @@ import (
 // Spec defines a specification for a "Gitlab" resource
 // parsed from an updatecli manifest file
 type Spec struct {
-	// [S][C][T] URL specifies the default github url in case of Gitlab enterprise
+	// [S][C][T] URL specifies the default Gitlab url in case of Gitlab enterprise
```
oxr463, line 15:
```suggestion
	// [S][C][T] URL specifies the default Gitlab Url
```

There is no difference between GitLab.com or any self-hosted instances. I think the "enterprise" bit was also left over from GitHub.

oxr463 (approved):

olblak:
Oups :p

pkg/plugins/resources/gitlab/client/main.go
olblak: Thanks for cleaning this up

olblak (approved):

PR #1233, opened 2023-03-19
Add Bitbucket Server (Stash)

naimo84:
Hi 😉

This is my first post to a golang repo, so please forgive me if everything isn't correct yet.  I left out the tests because there is no free version of bitbucket server.  there are apparently only on premise solutions.  I tested everything with our company bitbucket server.  if you want me to change anything, let me know. Or just delete the PR, If it's completely wrong.

Greets,
Benjamin

PR #1246, opened 2023-03-26
Add Bitbucket Server (Stash)

naimo84:
follow-up of https://github.com/updatecli/updatecli/pull/1233

👍

### DIFF
--- a/pkg/core/pipeline/action/main.go
+++ b/pkg/core/pipeline/action/main.go
@@ -12,15 +12,18 @@ import (
 	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
 	gitea "github.com/updatecli/updatecli/pkg/plugins/resources/gitea/pullrequest"
 	gitlab "github.com/updatecli/updatecli/pkg/plugins/resources/gitlab/mergerequest"
+	stash "github.com/updatecli/updatecli/pkg/plugins/resources/stash/pullrequest"
 	giteascm "github.com/updatecli/updatecli/pkg/plugins/scms/gitea"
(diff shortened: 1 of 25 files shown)


olblak:
Thanks for the clean pullrequest. 

I'll do a final review over the week but I just noticed a few improvement that are not part of the initial scope (to add stash scm support) which is why I don't consider them blocker and I can remove them in a later iteration.

The package `pkg/plugins/resources/stash/release` could probably be deleted as it relies on  
https://github.com/drone/go-scm/blob/master/scm/driver/stash/release.go which do not work

Similarly to Gitlab/Gitea/GitHub, it could have been used to search the latest publish release as a source or condition but I don't even know if stash support the concept of "release" but the go-scm library do not, so I would just remove that code.

The package `pkg/plugins/resources/stash/branch` could be used as a source of condition if it is correctly 
registered in pkg/core/pipeline/resource https://github.com/updatecli/updatecli/blob/main/pkg/core/pipeline/resource/main.go

Since It will be difficult to test it myself, that would be great if you could give it a try. Once registered, you should be able to have something like 

```
sources:
  default:
    kind: stash/branch
    spec:
      owner: Xxx 
      repository: xxx 
      url: xxx 
```

something similar to giteabranch should be  

While being able to use a source of type `stash/tag` vs `gittag` seems duplicated, the `stash/tag` queries an API, while `gittag` clones locally the full repository and then search for tags which can have performance consequences on very large git repositories with a lot of tags. I said it's out of the scope so you could also remove it 

And same thing for pkg/plugins/resources/stash/tag

olblak:
I opened this pullrequest on your fork https://github.com/naimo84/updatecli/pull/1

Issue #1088, opened 2023-01-16
Feature Request: Add Golang package support


olblak:
### Is your feature request related to a problem?

I would like to be able to use Go package version from source and condition.
There is an API documented on https://go.dev/ref/mod#goproxy-protocol

The following endpoint can be used 

List all versions for a specific package
`curl https://proxy.golang.org/github.com/updatecli/updatecli/@v/list`

Get package version information
`curl https://proxy.golang.org/github.com/updatecli/updatecli/@v/v0.40.1.info`

### Solution you'd like

I would like to use the following manifest

```
sources:
  example1:
    name: Get github.com/updatecli/updatecli package version
    kind: gopackage
    spec:
      package: github.com/updatecli/updatecli
      versionfilter:
        kind: semver
        pattern: "~0"

conditions:
  example1:
    name: Get github.com/updatecli/updatecli package version
    kind: gopackage
    disablesourceinput
    spec:
      package: github.com/updatecli/updatecli
      version: 0.30.0
      url: <differentProxy than proxy.golang.org>
```

### Alternatives you've considered

Instead of `package` we could use `name` such as 

I am expecting the following manifest to work

```
sources:
  example1:
    name: Get github.com/updatecli/updatecli package version
    kind: gopackage
    spec:
      name: github.com/updatecli/updatecli
      versionfilter:
        kind: semver
        pattern: "~0"

conditions:
  example1:
    name: Get github.com/updatecli/updatecli package version
    kind: gopackage
    disablesourceinput
    spec:
      name: github.com/updatecli/updatecli
      version: 0.30.0
```

### Anything else?

A similar pullrequest that can be used for inspiration https://github.com/updatecli/updatecli/pull/1081

olblak:
https://pkg.go.dev/golang.org/x/mod/modfile

olblak:
Saving for later, but here is an endpoint to retrieve all Golang version
https://go.dev/dl/?mode=json&include=all

PR #1258, opened 2023-03-31
Add Golang support

olblak:
Also related to https://github.com/updatecli/updatecli/issues/1088

This pullrequest introduces four golang related plugins.

* `golang` to retrieve the latest Golang version and its changelog
  * The source retrieve a Golang version and a http link to the changelog
  * The condition checks that a specific version exists 
* `golang/gomod` to manipulate go.mod file (retr)
  * The source retrieves the version for a specific go module path or the go version specified in the go.mod
  * The condition checks that a version is used for a specific go module path or if a specific go version is specified
  * The target allows to update module version or go version
*   `golang/module` to retrieve go module version from the registry
  * The source retrieves a version from the registry
  * The condition checks that a version exists on a registry  
* `Golang` autodiscovery plugin

* Changelog function querying github uses `UPDATECLI_GITHUB_TOKEN` or `GITHUB_TOKEN` to do authenticated call
* Golang autodiscovery only update module using valid sementic versioning

The reason why the module `golang` is useful is because the Golang project publish version such as `go1.20` where the prefix go is annoying as it prevent us to sort versions.  Also all new major version, do not contain the patch number, so I am adding the zero to it. Such as go1.20 would become 1.20.0

## Test

```shell
cd pkg/plugins/resources/go/language/
go test
cd pkg/plugins/resources/go/gomod/
go test
cd pkg/plugins/resources/go/module/
go test
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/golang/
```

### Tradeoff

/

### Potential improvement

* Golang proxy allows the ability to retrieve the source repository. This is usefull for go module that uses a different name than the source repository. It would retrieve that information


e2e/updatecli.d/success.d/golang/golang.yaml
```diff
@@ -0,0 +1,17 @@
+name: "Test Golanng version"
```
enrichman, line 1: Typo

e2e/updatecli.d/success.d/golang/golang.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+sources:
+  default:
+    name: Get Golang version
+    kind: golang
```
enrichman, line 6: I would consider also a `lang` category and having `lang/go`

e2e/updatecli.d/success.d/golang/gomod.yaml
```diff
@@ -0,0 +1,37 @@
...
+    spec:
+      file: pkg/plugins/resources/go/gomod/testdata/go.mod
+      module: "github.com/Masterminds/semver/v3"
+
+  golang:
+    name: Get version
+    kind: golang/gomod
+    spec:
+      file: pkg/plugins/resources/go/gomod/testdata/go.mod
+      kind: golang
```
enrichman, line 16: Is this `kind` needed since it's already a `golang/mod` kind?

e2e/updatecli.d/success.d/golang/gomod.yaml
```diff
@@ -0,0 +1,37 @@
...
+      file: pkg/plugins/resources/go/gomod/testdata/go.mod
+      version: "v3.2.0"
+
+  goversion:
+    disablesourceinput: true
+    name: Get version
+    kind: golang/gomod
+    spec:
+      module: "github.com/Masterminds/semver/v3"
+      kind: golang
```
enrichman, line 35: Same

e2e/updatecli.d/success.d/golang/module.yaml
```diff
@@ -0,0 +1,19 @@
+name: "Test gomod"
```
enrichman, line 1: Module?

e2e/updatecli.d/success.d/golang/module.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+sources:
+  default:
+    name: Get version
+    kind: golang/module
+    spec:
+      path: "github.com/Masterminds/semver/v3"
```
enrichman, line 8: I would go for `name` or `module` but not sure. :shrug:

e2e/updatecli.d/success.d/golang/gomod.yaml
olblak: This kind is used to define if we want to retrieve the Golang version or a module version but I can remove it. 
If a module is specified we assume we want to retrieve a module version otherwise we retrieve the golang  version specified in the go.mod file

e2e/updatecli.d/success.d/golang/gomod.yaml
olblak: same here :)

e2e/updatecli.d/success.d/golang/module.yaml
olblak: Thanks I am reviewing the  title of those e2e tests

e2e/updatecli.d/success.d/golang/golang.yaml
olblak: After additional though 
I find it weird to have 
* `lang/golang`
* `golang/module`
* `golang/gomod`

I am considering either `golang/lang` or simply `golang`

PR #1275, opened 2023-04-13
Add hash return to gittag plugin

matttrach:
Sometimes there are go repos that don't follow semver (or any known) naming scheme when it comes to release names or tag names, this breaks the 'go get' tool (specifically the version query to go proxy). I would like to keep things up to date without losing the go proxy benefits, the cleanest way I can see to do this is to give go proxy the specific hash of the tag to get.

Here is an example of the problem: https://github.com/k3s-io/kine/blob/master/Dockerfile.dapper#L6
  - you can see the tag is 'gopls/v0.9.5'
  - we have to use the "GOPROXY=direct" env variable to bypass the proxy completely to get this module

What I would like to do:
  - move the tag name to a comment
  - have updatecli get the last tag created with the 'gopls/' prefix
  - have updatecli generate a commit to update the comment with the tag name
  - have updatecli generate a commit to update the 'go install' line with the corresponding hash tag
  - have updatecli generate a PR for the commits

This change adds a "key" to the gittag plugin spec
  - It defaults to the old behavior
  - You can specify "", "name", or "hash", anything else throws an error
  - If you specify the "hash" key, then gittag will return with the tag's commit hash rather than the tag's name

This change adds some functionality to the native git handler
  - It extracts the functionality of getting the tag objects from git as "gitTagRefs".
  - It updates "gitTags" to use the new extracted function, but leaves the behavior alone.
  - It adds a new "gitTagHashes" function which retrieves the hashes from "gitTagRefs"

## Test

```shell
cd pkg/plugins/utils/gitgeneric
go test -v
```

```shell
cd pkg/plugins/resources/gittag
go test -v
```

### Tradeoff

The version filter assumes it is dealing with an array of strings, and I didn't want to change anything around the filtering, so rather than attempting to switch the behavior before filtering I just filtered on the tag names as normal, but then looped on the refs to find the ref which matches the name found. This could have negative side effects if there are more than one tag ref with the same name, but I think tag names must be distinct.

The end result is that the refs slice is looped twice; once to collect the tag names, and once to find the hash for the name found. This _could_ be a non-trivial loop if the repo has _lots_ of tags, but it seems like that is unlikely.

I decided to use the spec for gittag to indicate the desired output, which could be confusing to some users since it is the only plugin which can have different outputs.

### Potential improvement

I am challenging the design assumption for filters to only work with string arrays, and it feels sort of like a hack. I can't think of a way to do it better, but I would like to be able to parse complex sources without undermining the design (maybe some transformer plugin?).


pkg/plugins/resources/gittag/main_test.go
```diff
@@ -0,0 +1,163 @@
...
+			if tt.wantErr {
+				assert.Error(t, gotErr)
+				return
+			}
+			require.NoError(t, gotErr)
+		})
+	}
+}
+
+// spec: Spec{
```
matttrach, line 155: remove

olblak:
Thanks @matttrach for the well explained pullrequest. You pullrequest seems well tested.

pkg/plugins/resources/gittag/main.go
```diff
@@ -18,6 +18,10 @@ type Spec struct {
 	VersionFilter version.Filter `yaml:",omitempty"`
 	// Message associated to the git tag
 	Message string `yaml:",omitempty"`
+	// Key of the tag object to retrieve, default is tag "name"
+	//  filters are always against tag name, this only controls the output
+	//  current values are 'name' and 'hash'
```
olblak, line 23: Technical limitation of the library that I am relying on for generating the jsonschema (https://github.com/invopop/jsonschema#using-go-comments). The comment must be on one line.

Otherwise we'll just see "current values are 'name' and ''hash" both from IDE auto completation and documentation website

That being said, I don't see that constrain anymore so maybe I need to give it another try

olblak:
They are  a few file formatting issues  identified by `golanci-lint`

pkg/plugins/resources/gittag/main.go
matttrach: I went ahead and moved this into a single line.

matttrach:
sorry about that, they are now properly formatted.

PR #1303, opened 2023-04-24
perf: Add mutex guard to github functions for concurrent safety

iamrajiv:
This PR adds synchronization using **`sync.Mutex`** to **`Github.Changelog`** and **`Github.ChangelogV3`** methods to avoid concurrent access issues. In addition, a **`sync.Mutex`** is added to **`getTokenFromEnv`** function as well. These changes ensure that the shared state in the package is accessed safely by multiple goroutines.

## Test

```shell
cp <to_package_directory>
go test
```

### Tradeoff

There might be some overhead due to the added synchronization. However, the benefit of safe concurrent access justifies this tradeoff.

### Potential improvement

There could be potential improvement by using **`sync.RWMutex`** instead of **`sync.Mutex`** in the **`Github.Changelog`** method to allow multiple readers to access the shared state at the same time. However, since the current implementation only allows for a single reader at a time, the **`sync.Mutex`** is sufficient for now.


olblak:
Thanks for the contribution.

At this time, Updatecli only leverage go routines when it comes to cloning git repositories.
This means that all GitHub queries are done in sequence, I may improve the situation if that bothers me enough :) 

So base on my understanding, avoiding the concurrent access issues is not a something that I am facing but I may face it in the futur when I start using more goroutines

iamrajiv:
Thank you for the feedback, @olblak. I understand that Updatecli currently only uses go routines for cloning git repositories, and therefore concurrent access issues are not a problem at this time. However, I appreciate your insight that as I start using more goroutines in the future, I may face such issues. Thank you again for taking the time to review my PR.

PR #1330, opened 2023-05-17
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

## Test

The misspellings have been reported at https://github.com/jsoref/updatecli/actions/runs/5002389671#summary-13547422360

The action reports that the changes in this PR would make it _relatively_ happy: https://github.com/jsoref/updatecli/actions/runs/5002389949#summary-13547422914

### Tradeoff

* If there are library consumers and the changes break apis (I tried to avoid this), there could be some pain.
* A nontrivial number of files have been changed. From my perspective, this is a relatively small set of changes, and it's possible to split them and review them by kind, directory, or incrementally

PR #1340, opened 2023-05-22
Add check-spelling

jsoref:
This adds the [check-spelling action](https://github.com/marketplace/actions/check-spelling) based on [spell-check-this](
https://github.com/check-spelling/spell-check-this/blob/bbf1e22c6da7858723531a218c749cd1a75149d6/.github/workflows/spelling.yml).

## Test

To test this pull request:
1. Apply this commit to a fork
2. Apply a commit with a typo and push it to a branch in GitHub.
3. Afterwards, you can create a pull request into a branch with this commit (note that the behavior for push and pull request are not precisely identical).

### Tradeoff

There is some amount of work required to keep this workflow happy, generally this involves adding items to `expect.txt` (or removing them as they disappear). `expect.txt` has some case folding operations. You can, e.g. tell it that you tolerate `MICROSOFT` and then it will tolerate `MICROSOFT`, but it won't tolerate `Microsoft` nor `microsoft`. If you instead add `Microsoft`, it will tolerate both `MICROSOFT` and `Microsoft`, but not `microsoft`. If you add `microsoft`, it will tolerate `MICROSOFT`, `Microsoft`, and `microsoft` -- this follows how programmers interact with English. check-spelling wants `expect.txt` to closely match your word use, so if you stop using `microsoft` but still have `Microsoft`, it will suggest switching back to `Microsoft`.

You can instead add words to your custom dictionary (`allow.txt`) and check-spelling will treat them like any other word in its dictionary (no pruning involved).

You can also exclude files using `excludes.txt` or (portions of) lines using `patterns.txt`.

If `candidate.patterns` happens to match a line, it may make a suggestion of an additional item to add to `patterns.txt` -- If you don't like a suggestion, you can comment it out (or simply remove it from the file).

The contents of `advice.md` are included in reports when there are failures. -- You can/should adjust it based on the needs of your users.

As configured here, reporting is mostly via the job summary (e.g. [✅]( https://github.com/jsoref/updatecli/actions/runs/5002389671#summary-13547422360) or [❌](https://github.com/jsoref/updatecli/actions/runs/5002389949#summary-13547422914)). This is more secure than a form that involves adding a comment and it supports longer reports. There is support for sarif reporting, but sarif visibility is quite restricted, so generally speaking only repository owners will see them (this does include fork owners when fork owners push to their forks).

If you have questions, feel free to tag me at any time.

### Potential improvement

I have something of a roadmap: https://docs.check-spelling.dev/Possible-features.html, PR  https://github.com/updatecli/updatecli/pull/1330?notification_referrer_id=NT_kwDOACBWLLI2NDgzNjg0MDYwOjIxMTkyMTI#issuecomment-1557004255 was developed using features that should mostly be in a v0.0.22 (including fancier dictionaries which are configured here, smarter exclude suggestions, and a smarter apply flow).


olblak:
Thanks for the pullrequest 🎉 hoping that it will catch all my typos

PR #1344, opened 2023-05-25
Add Updatecli integration with Udash

olblak:
As an experiment to publish Updatecli report on an Updatecli backend.
This pullrequest introduces the two following enhancements

*  Add login capability to an Updatecli backend using PKCE OAuth2 flow
* Add function to publish report to the updatecli API 

This pullrequest is designed to work with:

* https://github.com/updatecli/udash-front for the frontend application
* https://github.com/updatecli/udash for the backend application

## Test

To test this pull request, you can run the following commands (if udash is available on localhost):

```shell
go build -o bin/updatecli .
./bin/updatecli login http://localhost:8081 --experimental
./bin/updatecli diff --config updatecli/updatecli.d/updatecli.yaml --experimental 
```

### Potential improvement

- [x] As my learning goes, another popular authentication mechanism uses "the device flow" which is already used to authenticate on GitHub api...

- [x] I have an open question on https://community.auth0.com/t/pkce-flow-risk-of-exposing-publicly-authdomain-and-clientid/108278


pkg/core/pipeline/main.go
```diff
@@ -58,9 +58,9 @@ func (p *Pipeline) Init(config *config.Config, options Options) error {
 	p.Actions = make(map[string]action.Action, len(config.Spec.Actions))
 
 	// Init context resource size
-	p.Report.Sources = make(map[string]*result.Source, len(config.Spec.Sources))
-	p.Report.Conditions = make(map[string]*result.Condition, len(config.Spec.Conditions))
-	p.Report.Targets = make(map[string]*result.Target, len(config.Spec.Targets))
+	p.Report.Sources = make(map[string]result.Source, len(config.Spec.Sources))
+	p.Report.Conditions = make(map[string]result.Condition, len(config.Spec.Conditions))
+	p.Report.Targets = make(map[string]result.Target, len(config.Spec.Targets))
```
dduportal, line 63: I'm unsure about this change. As far as I remember, we used references so we could change the objects.
Now that it's copied by value, that will change the behavior right?

Was it done for a specific need ?

pkg/core/result/target.go
```diff
@@ -14,8 +14,8 @@ type Target struct {
 			* "SKIPPED"
 	*/
 	Result string
-	// OldInformation stores the old information detected by the target execution
-	OldInformation string
+	// Information stores the old information detected by the target execution
+	Information string
```
dduportal, line 18: 👍

pkg/core/result/target.go
```diff
@@ -26,10 +26,12 @@ type Target struct {
 	Changed bool
 	// Scm stores scm information
 	Scm SCM
+	// ID contains a uniq identifier for the source
```
dduportal, line 29:
```suggestion
	// ID contains a uniq identifier for the target
```

pkg/core/auth/config.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// getOAUTHInfo queries the Udash website to retrieve OAUTH configuration
+func getOAUTHInfo(endpointURL string) (issuer string, audience string, clientID string, err error) {
```
dduportal, line 11:
```suggestion
func getOauthInfo(endpointURL string) (issuer string, audience string, clientID string, err error) {
```

Does not seem to be an acronym, so better to stick with the usual CamelCase

pkg/core/auth/config.go
```diff
@@ -0,0 +1,43 @@
+package auth
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// getOAUTHInfo queries the Udash website to retrieve OAUTH configuration
```
dduportal, line 10:
```suggestion
// getOAUTHInfo queries the Udash website to retrieve Oauth configuration
```

pkg/core/pipeline/main.go
olblak: I forgot the initial intend, and it bother me on Udash side but I'll revert this change. Thanks for the reminder

pkg/core/auth/config.go
olblak: Well it's an acronym for Open Authorization, https://en.wikipedia.org/wiki/OAuth
I hesitated between `OAUTH` and `Oauth` and imho the first is more appropriate

dduportal: I'm nitpicky but it's not an acronym but a "shorthand", so the "all letters in uppercase" rule does not apply.

I haven't seen any other usage with uppercase so that's why I nitpicked on it but no porblem to keep it. I'm just wondering if there isn't a trademark such as for Apache for instance (but I'm not sure where to look for this).

dduportal (dismissed):

pkg/core/pipeline/main.go
olblak: I reverted this change to use  apointer. Thanks for this reminder

PR #1347, opened 2023-05-26
#1346: Implement the support for registry in the autodiscovery feature for helm

### LINKED ISSUE #1346: Add support for the registry value in the autodiscovery for helm

I'm experimenting the auto-discovery feature `helm`.

I see that work fine using the `tag` and `repository` entry but only if the repository include also the `registry` entry, when they are defined in different entries (that seems to be the most used approach) is not able to work.

E.g.:
```
image:
    tag: 1.0.0.
    repository: quay.io/foo/bar
```
work fine, but:
```
image:
    tag: 1.0.0.
    registry: quay.io
    repository: foo/bar
```
fail since use the default registry (`hub.docker.io`) instead the one specified in the registry.

### Solution you'd like

Add support also for the `registry` entry

Do some manipulation in the yaml before and after using updatecli to woraround the problem

### Anything else?

I already have some code working, I'll open a PR soon

mavimo:
Fix #1346

Add support for the `registry` entry in the `image` section when we are using the helm autodiscovery feature.

## Test

This pull request include test to cover the current implementation


olblak:
Thanks @mavimo for this great PR, I just tested it and it works great

PR #1368, opened 2023-06-08
Fix dockerfile target result in case of success

olblak:
While doing some test with the dockerfile autodiscovery, I noticed a regression in the dockerfile plugin resource.
Where the result is not set if nothing change 

```
TARGETS
========

registry.access.redhat.com/ubi9/ubi
-----------------------------------

**Dry Run enabled**

✔ The line #12, matched by the keyword "FROM" and the matcher "registry.access.redhat.com/ubi9/ubi", is correctly set to "FROM registry.access.redhat.com/ubi9/ubi:9.2-489".
- - 


```

## Test

```shell
cp <to_package_directory>
go test
```

### DIFF
--- a/pkg/plugins/resources/dockerfile/target.go
+++ b/pkg/plugins/resources/dockerfile/target.go
@@ -38,7 +38,9 @@ func (d *Dockerfile) target(source string, dryRun bool, resultTarget *result.Tar
 	}
 
 	if len(changedLines) == 0 {
-		logrus.Debugf("empty file detected %q, nothing to do", d.spec.File)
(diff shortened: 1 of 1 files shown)


dduportal:
Confirmed that this PR fixes the error: we have the error with https://github.com/jenkinsci/docker-ssh-agent/blob/master/updatecli/updatecli.d/git-windows.yml (the GHA checks are red for now, using latest updatecli).

Tried locally, comparing latest updatecli and a build of the upcoming 0.53.0: the problem is gone with 0.53.0.
Thanks!

PR #1382, opened 2023-06-19
Conditional targets

dcoraboeuf:
This PR proposes the addition of conditions at target level.

As of now, the `conditions` are all or nothing: they need to all pass before we go on with the pipeline and all its defined targets.

However, in some circumstances, we may want to run a target only if a given condition is met (for example: not updating all the files when only a patch). In this case, we'd need to be able to specify a condition at target level.

This would look like:

```yaml
targets:
   my-target:
      ...
      spec:
         ...
         condition:
             any-condition...
```

For now, opening as a draft since I'd like to discuss the concept, the approach and also the testability of this. Of course, documentation & tests would have to be added.

### DIFF
--- a/pkg/core/pipeline/target/main.go
+++ b/pkg/core/pipeline/target/main.go
@@ -3,6 +3,7 @@ package target
 import (
 	"errors"
 	"fmt"
+	"github.com/updatecli/updatecli/pkg/core/pipeline/condition"
 	"strings"
 
 	jschema "github.com/invopop/jsonschema"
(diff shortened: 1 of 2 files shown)


dcoraboeuf:
Hello @dduportal ,

My mistake, sorry, I had explained in the chat (in Matrix) and forgot to port the discussion here. Updating the description.

olblak:
I had a look and I didn't find an existing issue or discussion about conditional.
This means that it stayed in my head :D 

This is an important enough topic which I would like to discuss first in its own issue before any implementation.
I'll open one later today

olblak:
Closing this PR as we opened an issue to gather discussion on this topic before implementaiton

Issue #1363, opened 2023-06-07
Action ID change over time


olblak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A action with a dynamic title will generate a new PR ID each time the value change

The ID generated from https://github.com/updatecli/updatecli/blob/0289b297f73597a92c70a9990f9382897f960ed5/pkg/core/pipeline/actions.go#L120 based on an information that can change over time

### Expected Behavior

That ID should never change :) 

### Steps To Reproduce

Manifest: https://github.com/civo/kubernetes-marketplace/blob/master/.github/updatecli/updatecli.d/argocd.yaml
Generate a PR such as https://github.com/civo/kubernetes-marketplace/pull/438 where each  new version will generate a new report

### Environment

```markdown
- OS:linux
- updatecli:0.52.0
```

olblak:
The problem is we can reuse the `pipelineID` value as by design different Updatecli manifest can reuse the same pipelineID to explicitly declare the different manifest are part of the same pipeline.

olblak:
So far the best way solution that I can think of would be to allow to configure the actionID via a new setting available from the manifest

dduportal:
🤔 interesting problem. What does the ActionID maps to in the reality of a GitHub PR (or a GitLab MR)?

Is it the source branch's name or another attribute?

I'm asking because that could be really usefull to have human readable source branch names

olblak:
Another potential solution that would be useful for other situation as well such as https://github.com/updatecli/updatecli/issues/1355 or https://github.com/updatecli/updatecli/issues/1364
 
Would be to add a new function to the resource interface that would returned a "sanitized" version of the manifest that could be used to generate a uniqu resource ID 

for example 

```
kind: githubrelease
spec:
    owner: updatecli
    repository: updatecli
    token: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
    username: '{{ requiredEnv "UPDATECLI_GITHUB_ACTOR" }}'
``` 

would return 


```
kind: githubrelease
spec:
    owner: updatecli
    repository: updatecli
```

which is all the information needed to identify a githubrelease

Once we have all the manifest resource ID then we can combined them such as 

pipeline is the combinaison of all manifest resource ID (scm,source, condition, target)
action is a combinaison of all target ID linked to the action

That sanitize manifest would also be useful for the webapp, as I could have dashboard such as 

"show me all information available for a github release updatecli/updatecli"

[...]

PR #1395, opened 2023-06-22
Generate uniq resource ID

### LINKED ISSUE #1364: Feature Request: Need a uniqu ID per pipelin

While working on a webapp to visualize Updatecli reports, I struggle on a way to identify a pipeline accross different repositories or pipeline execution.

Using the manifest metadata such as path leads to collusion
The same manifest can be executed from a machine without git or different git repository/ branch

Using any information from the manifest can lead to collusion as it's quite common for the same manifest to update different repository...

### Solution you'd like

_No response_

_No response_

### Anything else?

_No response_

olblak:
Fix #1364 #1363

This pullrequest is not a "directly" user facing.

The current problem is we often rely on the Updatecli manifest to generate IDs such as pipelineID or actionID
When that manifest contains dynamic information then the generated ID changes over pipeline execution.

After many iteration, I couldn't find a better way to generate those ID.
So this PR is another attempt to solve this problem.

So for each resource we add a new function `GetAtomicSpec()` which return an minimal version of the resource specification such as 

```
githubrelease.Spec {
  owner: "updatecli",
  repository: "updatecli",
  token: "mySecretToken",
  versionFilter: version.VersionFilter {
    kind: "semver",
    pattern: "v0.3.0"
  } 
}
```

would become 

```
githubrelease.Spec {
  owner: "updatecli",
  repository: "updatecli",
}
```

`token` should be remove as it's a dynamic information and don't provide information to identify the github release
`versionfilter` should also be removed we are only interested by the information that can help identifying the github release

This means that three users executing the same manifest will have the same resourceID 

## Test

```shell
cp <to_package_directory>
go test
```

### Tradeoff

It's an new function that need to be created for the resource interface

### Potential improvement

* With the change, we would be able to simplify pullrequest body by removing the first level
* We can generate report without any credentials or free of useless information

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -127,7 +127,7 @@ require (
 	github.com/google/go-cmp v0.5.9 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 // indirect
-	github.com/google/uuid v1.3.0
+	github.com/google/uuid v1.3.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.2.3 // indirect
 	github.com/googleapis/gax-go/v2 v2.7.1 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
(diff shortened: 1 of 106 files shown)


olblak:
After additional thoughts, I think I can delay a bit this refactoring.
I am concern that the cost to add more functions to maintain for each resource will not over come the benefit.

PR #1406, opened 2023-07-02
wip: add gitpush action

dduportal:
Related to https://github.com/updatecli/updatecli/issues/465

This PR introduces a new kind of action: `git/push`.

Please note that the targets of type `git/tag` and `git/branch` have a changed (fixed!) behavior: they now won't perform the `git push` operation as it is the responsibility of the the newly introduced `gitpush` action.

## Test

To test this pull request, you can run the following commands:

```shell
cp <to_package_directory>
go test
```

PR #1407, opened 2023-07-03
fix(gittag) ensure `gittag` targets does not fail due to (unspecified) version filters

dduportal:
This PR fixes the `gittag` resource where the target fails due to the default Version Filter.

There are 2 blocking bugs fixed at the same time:

- Comparing error message between 2 packages is fragile as the errors templates are subject to changes
  - The proposed solution here is to export the errors in the `version` so their error message (template) can be reused for solid comparison.
- Validating the target configuration must be done against the `spec` (provided by user) and not against the internal configuration

## Test

- Unit tests have been checked (and will be by the github action checks as a reference)
- The End to end test has been renamed and updated to cover all cases:
  - Both Git and GitHub SCM for `gittag` sources
  - Both Git and GitHub SCM for `gittag` conditions (always passing)
  - Added both Git and GitHub SCM for `gittag` targets (to underline the fixed errors which is caught at the `updatecli diff` steps already)
  - Added a target case expected to never change (trying to create an existing tag)

## Additional Information

Caught while working on https://github.com/updatecli/updatecli/pull/1406


olblak (approved):
Very clean
Does the pullrequest introduces a breaking change?

olblak:
> Very clean Does the pullrequest introduces a breaking change?

I don't think so but better to ask

olblak (approved):

dduportal:
> Very clean Does the pullrequest introduces a breaking change?

It should not. If it does, then it would be a bug (and I would work on fixing it as soon as possible)

Issue #1028, opened 2022-12-09
YAML multiline folded block scalars broken by apply


alxgomz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using YAML target on a file which contains multi-line folded block scalars (`>-`) renders a new YAML file with single-line scalar as shown in the diff below:

```diff
       JAVA_OPTS: >-
-        -XX:MinRAMPercentage=50
-        -XX:MaxRAMPercentage=80
-
+        -XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=80
     postgres:
```

### Expected Behavior

formatting of the scalar value should be identical to the input (if valid)

### Environment

```markdown
- OS: MacOSX
- updatecli: 0.39.0
```

### Anything else?

May be connected to #794

olblak:
Thanks for reporting this issue.
This highlight that some effort is needed on the yaml resource

olblak:
After some quick tests, it appears that https://github.com/updatecli/updatecli/pull/1333 would solve this issue

Issue #903, opened 2022-10-03
Uwanted changes applied to YAML targets


jamiezieziula:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I run `updatecli apply` it works mostly as expected. However, it is performing some amount of linting functionality on my `yaml` files.  Specifically:
- it is removing top level `---` from my files
- removes double spaces between code and comments
- removes subsequent manifests if my yaml file contains more than one

### Expected Behavior

I wouldn't expect any linting behavior to be applied, or the ability to disable it.

### Steps To Reproduce

run `updatecli apply` with the following updatecli manifest.yaml specified:
```
sources:
  istio-gateway:
    kind: helmchart
    spec:
      url: "https://istio-release.storage.googleapis.com/charts"
      name: "gateway"
      version: "1.*.*"
    sourceid: istio-gateway
conditions: {}
targets:
  gateway:
    name: bump chart dependencies
    kind: yaml
    spec:
      file: "gateway.yaml"
      key: "spec.chart.spec.version"
    sourceid: istio-gateway
```
on the below istio manifest file.  The dashes & secondary manifest will be removed.
```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: gateway
  namespace: istio-ingress
spec:
  interval: 5m
  chart:
    spec:
      chart: gateway
      version: 1.15.0
      sourceRef:
        kind: HelmRepository
        name: istio
        namespace: flux-system
      interval: 5m
---
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: gateway
  namespace: istio-ingress
spec:
  selector:
    istio: gateway
  servers:
    - port:
        number: 443
        name: https
        protocol: HTTPS
      hosts:
        - "*"
      tls:
        credentialName: tls
        mode: SIMPLE
```

### Environment

```markdown
- OS: Mac Monterey & ubuntu-latest in GitHub Actions
- updatecli: 0.34.0
```

lemeurherve:
Duplicate of https://github.com/updatecli/updatecli/issues/794
This is annoying indeed.

olblak:
I agree with you.
They are a bunch of limitation from the underling library such as https://github.com/go-yaml/yaml/issues/899

olblak:
**saving for later** 
I recently discovered this library https://github.com/goccy/go-yaml which could be a good replacement for https://github.com/go-yaml/yaml.

Issue #522, opened 2022-02-10
yaml issue with file contain unicode character


sileht:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I update file with yaml plugins I got an unexpected changes to the yaml file, like:

```
-      - name: Build 🔧 & Test 🔍
+      - name: "Build \U0001F527 & Test \U0001F50D"
```

The `key` spec was not updating this yaml attribute.

### Expected Behavior

Unicode characters are not replaced by their binary representation.
The line should not be updated.

### Steps To Reproduce

Edit a yaml file containing emoji.

### Environment

```markdown
- OS: archlinux
- updatecli:

Application:	0.19.2
Golang     :	1.17.6 linux/amd64
Build Time :	2022-02-04T08:32:54Z
```

olblak:
Thanks @sileht for reporting this.
updatecli relies on go-yaml to manipulate YAML file and the issue seems to be related to https://github.com/go-yaml/yaml/issues/737

PR #1412, opened 2023-07-08
Use vmware-labs/yaml-jsonpath for yaml path

mavimo:
Fix #XXX

This PR swap the implementation used for parse yamlpath to https://github.com/vmware-labs/yaml-jsonpath/ that support

## Test

```shell
cd pkg/plugins/resources/yaml/
go test
```

### Tradeoff

Swap from https://github.com/go-yaml/yaml to https://pkg.go.dev/gopkg.in/yaml.v3

### Potential improvement

Increase test coverage
Code cleanup


olblak:
Thanks for the pullrequest, I wasn't aware of this library.
It would definitely be a great addition to Updatecli as we would be able to specify more robust yaml key by using advanced yamlpath queries.
My only concern is moving back to a library that relies on go-yaml/yaml means we would reopen some limitations. Like a document with multiple `---` or unicode in the yaml document 

Both libraries https://github.com/goccy/go-yaml and https://github.com/go-yaml/yaml have pro and cons
I need to do additional testing to better understand them.

I'll soon be fully available

mavimo:
@olblak thx for the feedback, is the same doubt that I have, in the meantime I start to investigate if I can try to create a variant of the yaml-jsonpath library that work on the AST produced by `[goccy/go-yaml](https://github.com/goccy/go-yaml)`...

olblak:
> @olblak thx for the feedback, is the same doubt that I have, in the meantime I start to investigate if I can try to create a variant of the yaml-jsonpath library that work on the AST produced by `[goccy/go-yaml](https://github.com/goccy/go-yaml)`...

I am curious how much effort would be needed.

mavimo:
@olblak not trivial but doable (I expect a couple of days of works plus the time to integrate it.

If you think is the best way I'll try to find some time to make it happen..

olblak:
What would you think to have a flag to either enable the `default` or `yamlpath` where `default` is the current situation using goccy library and `yamlpath` would rely on your PR

My reasoning is both `goccy/go-yaml` and `vmware-labs/yaml-jsonpath` have major limitation and both can still improve independently.

That being said `vmware-labs/yaml-jsonpath` is used by many large project like Tanzu, Kubernetes, or devspace which is a good sign

Let me tests how the current PR behaves by the various issues by the move to goccy

olblak:
As expected, after a quick test, this issue would reopen: 

* Fix https://github.com/updatecli/updatecli/issues/522
* Fix https://github.com/updatecli/updatecli/issues/1028
* Fix https://github.com/updatecli/updatecli/issues/903

olblak:
After some test with the new jsonpath, it's really a great, and it would make many pipeline more robust, especially when manipulating arrays. 

In the current situation, I think both implementation make sense

olblak:
I opened a PR on your fork https://github.com/mavimo/updatecli/pull/1/files with the fix I needed

mavimo:
@olblak added your suggestion + rebased

olblak:
closing in favor of https://github.com/updatecli/updatecli/pull/1741

PR #1426, opened 2023-07-28
Fix condition report by using the correct variable

olblak:
While working on Udash, I noticed the condition report data were incomplete.

image

This commit also removes declared but unused variable which were probably left over from another iteration


olblak:
After this fix 

image

dduportal (approved):
👍

Issue #1415, opened 2023-07-14
On Windows --values is broken


stevehipwell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

However I specify `--values` on a Windows machine I get the following error.

> ERROR: failed loading pipeline(s)
        * "updatecli\\updatecli.d\\test.yaml" - stat updatecli\values.yaml: invalid argument

### Expected Behavior

I'd expect the `--values` argument to work with relative or absolute paths.

### Steps To Reproduce

1. In Windows
2. Run `updatecli diff --config <config_dir> --values <values_path>`

### Environment

```markdown
- OS: Windows 11
- updatecli: v0.54.0
```

olblak:
I did some testing, and this pull request https://github.com/updatecli/updatecli/pull/1447
solves the current issue.

It will be available in the next release 0.56.0

Issue #1253, opened 2023-03-30
invalid argument error when you pass full path in --values


kuisathaverat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The pipeline fails when you pass the full path for the values file

```
>  updatecli manifest show --config .ci/update-versions/update-helm-charts.yaml --values /Users/myuser/src/a-long-repository-name/.ci/update-versions/helm-charts.yaml


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline ".ci/update-versions/update-helm-charts.yaml"
ERROR: failed loading pipeline(s)
        * ".ci/update-versions/update-helm-charts.yaml" - stat /Users/myuser/src/a-long-repository-name/.ci/update-versions/helm-charts.yaml: invalid argument
```

### Expected Behavior

The same pipeline pass in `0.44.4`, and pass in `0.45.0` if you pass a relative path

```
>  updatecli manifest show --config .ci/update-versions/update-helm-charts.yaml --values .ci/update-versions/helm-charts.yaml


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline ".ci/update-versions/update-helm-charts.yaml"

SCM repository retrieved: 1
```

### Steps To Reproduce

1 - create a `/tmp/values.yaml` file (replace the values for valid values to checkout a repo)
```yaml
Email: my-email@gmail.com
Owner: owner
Repository: repo
Token: gho_manythings
User: full name
```
2 - create a `/tmp/manifest.yaml` file
```yaml
---
title: Update
scms:
  githubConfig:
    kind: "github"
    spec:
      user: '{{ .User }}'
      email: '{{ .Email }}'
      owner: "{{ .Owner }}"
      repository: "{{ .Repository }}"
      token: '{{ requiredEnv "GITHUB_TOKEN" }}'
      username: '{{ .User }}'
      branch: "main"
```
3 - go to the `/tmp` folder
4 execute `updatecli manifest show --config manifest.yaml --values /tmp/values.yaml` it will fail with an invalid argument error
```
> updatecli manifest show --config manifest.yaml --values /tmp/values.yaml


+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "manifest.yaml"
ERROR: failed loading pipeline(s)
        * "manifest.yaml" - stat /tmp/values.yaml: invalid argument

0 pipeline(s) successfully loaded

SCM repository retrieved: 0


++++++++++++++++++
[...]

PR #1447, opened 2023-08-04
allow to merge multiple values/secret files

### LINKED ISSUE #1403: Feature Request: multiple values files can be specified

Like with the helm cli, it should be possible to specify multiple `--values` options so values file can be split in smaller chunks, which would also increase the reusability across different updatecli pipelines.

### Solution you'd like

when multiple `--values` files are provided, they got merged in a single structure, the latter has precedence.

yaml files can be pre-processed with yq before using updatecli, and may still be necessary if some custom merging behavior is desired (e.g. array merging policy)

### Anything else?

_No response_

olblak:
* fix multiple values file specified
* fix absolute values file path specified

Fix #1403  #1415 #1253

## Test

```shell
cd pkg/core/config/
go test

go build -o bin/updatecli .
./bin/updatecli diff --config manifest.yaml --values --values ../tmp/1.yaml --values ./2.yaml --values /tmp/3.yaml
```

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -22,3 +22,4 @@ Udash
 CLIENTID
 minamijoyo
 hcledit
+work-around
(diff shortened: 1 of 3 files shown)


olblak:
Tested the PR on windows and both relative and absolute path worked

PR #1451, opened 2023-08-07
disable go mod caching

olblak:
This pullrequest is an attempt to fix the now common `no space left issue`

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -18,6 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
+          cache: false
           go-version: 1.20.1
         id: go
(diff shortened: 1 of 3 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
+          cache: false
           go-version: 1.20.1
```
dduportal, line 22: Unrelated to the PR: it looks like the go version is not updated in this file

dduportal (dismissed):

.github/workflows/updatecli.yaml
olblak: Yep I noticed, but I would like first to understand how to free some space

PR #1463, opened 2023-08-10
Allow Updatecli manifest/values to use JSON file

olblak:
depends on:
 - [x] https://github.com/updatecli/updatecli/pull/1457
 
Allow to use JSON, technically speaking json a subset of YAML and therefor fully compatible.
For complex template, I find using json more convenient than yaml as we don't have to deal with `tab` and `space`

## Test

```shell
cd pkg/core/config/
go test
cd ../../
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/npm.jso
```

### Tradeoff

/

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -30,7 +30,7 @@ jobs:
           # Required: the version of golangci-lint is required
           # and must be specified without patch version:
           # we always use the latest patch version.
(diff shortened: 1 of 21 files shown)


dduportal:
As long as JSON does not replace YAML, it looks ok 👍

PR #1544, opened 2023-08-31
Update golangci-lint installation method

olblak:
At the moment golangci-lint is failing on all new PR, 
I can't reproduce the typecheck error locally, so I am switching the installation method to the binary one


olblak:
golangci-lint passed so I am not gonna wait for the full test to passed

PR #1575, opened 2023-09-08
Issue/1529

### LINKED ISSUE #1529: Using rootdir without scm with helm autodiscovery triggers a failure

### Current Behavior

feels like updatecli internally it's referencing the Chart.yaml currently selected using a path relative to the dir specified in `rootdir`, but actually the PWD could be another directory (the directory from which I am running updatecli)

Errors are like:

```
alfresco-common
---------------
ERROR: reading yaml file: ✗ The specified file "alfresco-content-services/Chart.yaml" does not exist

```

### Expected Behavior

Charts dependencies should be processed correctly in the same way they are when not using `rootdir`

### Steps To Reproduce

I have a local [repository](https://github.com/Alfresco/acs-deployment) checked out which contains an `helm` folder, so I tried initially with this config:

```
name: "Helm autodiscovery for charts dependencies"
autodiscovery:
  crawlers:
    helm:
      rootdir: helm
      ignorecontainer: true
```

but running `updatecli apply --experimental` lead to multiple errors (one for each declared dependency of the `alfresco-content-services` chart) as reported above.

### Environment

```markdown
- OS: macOS Ventura 13.5.1 
- updatecli: v0.57.0
```

### Anything else?

_No response_

olblak:
Fix #1529 

Autodiscovery relative path was wrongly defined which triggered an issue when using `rootdir` 

## Test

```shell
cd <to_package_directory>
go test
```

### DIFF
--- a/pkg/plugins/autodiscovery/cargo/dependencies.go
+++ b/pkg/plugins/autodiscovery/cargo/dependencies.go
@@ -3,6 +3,7 @@ package cargo
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
 	"sort"
 	"text/template"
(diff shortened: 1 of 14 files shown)


olblak:
I am moving this one back to draft.
After some testing, I release that this PR creates more problem than it really solves

PR #1589, opened 2023-09-12
fix: ensure docker image checks architecture

### LINKED ISSUE #1582: [Docker Image] Condition does not honour the architecture filter

### Current Behavior

Conditions of kind `dockerimage` are ignoring the architecture filter.

It leads to production deployment issues (ref. https://github.com/jenkins-infra/kubernetes-management/pull/4338) as the container images versions are published even if the target architecture is not available.

### Expected Behavior

I expect conditions of kind `dockerimage` to fail when the provided architecture is invalid and/or does not have a corresponding manifest.

### Steps To Reproduce

Any manifest with the following conditions:

```
conditions:
  checkDockerImagePublishedList:
    name: "Test jenkinsciinfra/wiki:<latest_version> docker image tag"
    kind: dockerimage
    sourceid: latestRelease
    spec:
      image: "jenkinsciinfra/wiki"
      architectures:
        - tttt
        - aaaa
  checkDockerImagePublishedString:
    name: "Test jenkinsciinfra/wiki:<latest_version> docker image tag"
    kind: dockerimage
    sourceid: latestRelease
    spec:
      image: "jenkinsciinfra/wiki"
      architecture: bbb
```

must fail as these architecture do no exist

### Environment

```markdown
- OS: Ubuntu, macOS
- updatecli: v0.59.0
```

### Anything else?

This might be a broader issue (to the whole `dockerimage `resource 😱 )

mcwarman:
Fix #1582

Moved away from setting remoteOptions for single architecture, as this was causing the multi conditions case to fail.

I can remove the tests if we don't want it to look up those images.

## Test

```shell
cd pkg/plugins/resources/dockerimage
go test
```

### Potential improvement

Mock the tests

### DIFF
--- a/pkg/plugins/resources/dockerimage/condition.go
+++ b/pkg/plugins/resources/dockerimage/condition.go
@@ -2,11 +2,7 @@ package dockerimage
 
 import (
 	"fmt"
-	"strings"
 
-	"github.com/google/go-containerregistry/pkg/name"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sirupsen/logrus"
 	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
 	"github.com/updatecli/updatecli/pkg/core/result"
@@ -16,44 +12,25 @@ import (
 // We assume that if we can't retrieve the docker image digest, then it means
 // it doesn't exist.
 func (di *DockerImage) Condition(source string, scm scm.ScmHandler, resultCondition *result.Condition) error {
-
 	if scm != nil {
 		logrus.Warningf("SCM configuration is not supported for condition of type dockerimage. Remove the `scm` directive from condition to remove this warning message")
 	}
 
-	refName := di.spec.Image
-	switch di.spec.Tag == "" {
-	case true:
-		refName += ":" + source
-	case false:
-		refName += ":" + di.spec.Tag
-	}
-
-	ref, err := name.ParseReference(refName)
+	ref, err := di.createRef(source)
 	if err != nil {
-		return fmt.Errorf("invalid image %s: %w", refName, err)
+		return err
 	}
 
-	found := false
-	switch len(di.spec.Architectures) {
-	case 0:
-		// If only one architecture is specified, then di.options are already set: let's proceed
-		found, err = checkImage(ref, di.spec.Architecture, di.options)
+	found := true
+
+	for _, arch := range di.spec.Architectures {
+		foundArchitecture, err := di.checkImage(ref, arch)
 		if err != nil {
 			return err
 		}
-
-	default:
-		for _, arch := range di.spec.Architectures {
-			remoteOptions := append(di.options, remote.WithPlatform(v1.Platform{Architecture: arch, OS: "linux"}))
-
-			found, err = checkImage(ref, arch, remoteOptions)
-			if err != nil {
-				return err
-			}
-			if found {
-				break
-			}
+		if !foundArchitecture {
+			found = false
+			break
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/plugins/resources/dockerimage/main.go
```diff
@@ -70,14 +72,64 @@ func New(spec interface{}) (*DockerImage, error) {
...
+			logrus.Infof("%s The Docker image %s doesn't exist.",
+				result.FAILURE,
+				ref.Name(),
+			)
+			return false, nil
+		}
+		return false, err
+	}
+
+	_, err = descriptor.Image()
```
mcwarman, line 115: This was the important part the `WithPlatform` were not being checked until this call.

pkg/plugins/resources/dockerimage/condition.go
```diff
@@ -16,44 +12,25 @@ import (
...
 	}
 
-	found := false
-	switch len(di.spec.Architectures) {
-	case 0:
-		// If only one architecture is specified, then di.options are already set: let's proceed
-		found, err = checkImage(ref, di.spec.Architecture, di.options)
+	found := true
+
+	for _, arch := range di.spec.Architectures {
```
mcwarman, line 26: This block has been changed how it worked previously, it will now fail if _any_ are not found.

olblak:
Thanks for looking into this issue, my understanding is they are still images out there that uses the old schema so we may need to disable the architecture support if architecture/architectures is unset. like before your fix :p

mcwarman:
I think the fix actually does deal with that scenario:

- New default to be empty list of achitectures
- on `source` if list is empty parse empty string to checkImages, else take the first in the list.
- on `condtions` if list is empty parse empty string to checkImages, else iterate over architectures 
- in `checkImage` then deals with empty arch by not checking it, if it is provided and its unsupported only return an error then.

dduportal:
> I think the fix does potentially deal with that scenario, unless I've misunderstood what you are saying:
> 
>     * New default to be empty list of achitectures
> 
>     * on `source` if list is empty parse empty string to checkImages, else take the first in the list. [source.go#L51-L60](https://github.com/mcwarman/updatecli/blob/fix/docker-images-check-arch/pkg/plugins/resources/dockerimage/source.go#L51-L60)
> 
>     * on `condtions` if list is empty parse empty string to checkImages, else iterate over architectures [condition.go#L26-L42](https://github.com/mcwarman/updatecli/blob/fix/docker-images-check-arch/pkg/plugins/resources/dockerimage/condition.go#L26-L42)
> 
>     * in `checkImage` then deals with empty arch by not checking it, if it is provided and its unsupported only return an error then.
>       
>       * [main.go#L102-L104](https://github.com/mcwarman/updatecli/blob/fix/docker-images-check-arch/pkg/plugins/resources/dockerimage/main.go#L102-L104)
>       * [main.go#L118-L136](https://github.com/mcwarman/updatecli/blob/fix/docker-images-check-arch/pkg/plugins/resources/dockerimage/main.go#L118-L136)

That looks a sane set of code paths 👍 Thanks for looking into it!

olblak:
Thanks for the explanation.
I just did some testing, and I really like how you nicely solve the problem

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -70,14 +71,75 @@ func New(spec interface{}) (*DockerImage, error) {
...
+					arch,
+				)
+				return false, nil
+			}
+			return false, err
+		}
+	}
+
+	logrus.Infof("%s The Docker image %s (%s) exists and is available.",
+		result.SUCCESS,
```
olblak, line 139:
```suggestion
```

I stopped printing "result.SUCCESS" or `result.FAILURE` in plugin, because the `failedwhen` param defined at the condition level https://github.com/updatecli/updatecli/blob/9d1e0d4a6bc9d3560eaf15bfe9e6952eca1aa336/pkg/core/pipeline/condition/main.go#L106

takes the final decision either "failing" a condition should be considered as a success or a failure.

pkg/plugins/resources/dockerimage/main.go
```diff
@@ -70,14 +71,75 @@ func New(spec interface{}) (*DockerImage, error) {
...
+
+		if descriptor.MediaType == types.DockerManifestSchema1 || descriptor.MediaType == types.DockerManifestSchema1Signed {
+			return false, fmt.Errorf("architecture check not supported with MediaType %q", descriptor.MediaType)
+		}
+
+		_, err = descriptor.Image()
+		if err != nil {
+			if strings.Contains(err.Error(), "no child with platform") {
+				logrus.Infof("%s The Docker image %s (%s) doesn't exist.",
+					result.FAILURE,
```
olblak, line 128:
```suggestion
```
Same comment than https://github.com/updatecli/updatecli/pull/1589/files#r1324074159

olblak (approved):
@mcwarman Great PR

PR #1593, opened 2023-09-13
feat: add new `auto` versionincrement field for helmchart resource

aroberts87:
The intent of this change is to allow helm charts to have a version bump that is proportional to the version bump done by a dependency, rather than just a static major, minor, or patch increment. This will allow potential major or breaking changes in subchart dependencies from being masked by a lesser increment done on the parent chart, as well as larger version increments on the parent from looking more daunting than the changes they contain of the subchart bump is a lesser type. 

As an example, with a simple one parent / one child chart relationship and the parent chart's starting version of `1.0.0`, the new behavior yields the following behavior.

|  Child Original Version | Child Version Bump | New Parent Chart Version |
|-------------------------------|----------------------------|-------------------------------------|
| 0.3.0                            | 0.3.1                         | 1.0.1                                   |
| 0.3.0                            | 0.4.0                         | 1.1.0                                   |
| 0.3.0                            | 1.0.0                         | 2.0.0                                   |

Coupled with the functionality added in https://github.com/updatecli/updatecli/pull/1541, it will now be possible to have the following behavior when dealing with parent charts that have multiple subcharts with simultaneous version bumps during the same `updatecli` invocation.

|  Child A Original Version |  Child B Original Version | Child A Version Bump | Child B Version Bump | New Parent Chart Version |
|----------------------------------|----------------------------------|-------------------------------|------------------------------|------------------------------------|
| 0.3.0                                | 5.0.3                                | 0.3.1                            | 5.0.3                           | 1.0.1                                   |
| 0.3.0                                | 5.0.3                                | 0.4.0                            | 5.0.4                           | 1.1.0                                  |
| 0.3.0                                | 5.0.3                                | 0.8.7                            | 6.0.3                           | 2.0.0                                  |

## Test

```shell
cd pkg/plugins/resources/helm
go test
```

### Potential improvement

* Based on my testing running `apply` with `--push=false` breaks the intended workflow as the `originChartMetadata` values do not appear to be updated with the cumulative updates between `helmchart` targets.

* I'm not married to the term `child-consistent` for describing the version bump type, but at the same time could not think of another concise way to describe the intent that I felt would be comfortable using as parameter.


olblak:
For information the failing check can be ignore by adding "childdependent" to `.github/actions/spelling/allow.txt`

aroberts87:
Hey @olblak! I like your proposals a lot and will make the adjustments. I apologize for the delay in response as well; I was away on vacation and am just catching back up.

Thanks for your suggestions!

olblak:
> Hey @olblak! I like your proposals a lot and will make the adjustments. I apologize for the delay in response as well; I was away on vacation and am just catching back up.
> 
> Thanks for your suggestions!

Thanks for the contribution and don't worry for the delay, I was a few day off as well.
I'll try to review in the coming days and target the next minor version update

pkg/plugins/resources/helm/utils.go
```diff
@@ -263,6 +263,22 @@ forLoop:
...
+				computedVersion = v.IncMinor().String()
+				continue
+			}
+
+			if newVer.Major() > origVer.Major() {
+				computedVersion = v.IncMajor().String()
+			} else if newVer.Minor() > origVer.Minor() {
+				computedVersion = v.IncMinor().String()
+			} else if newVer.Patch() > origVer.Patch() {
+				computedVersion = v.IncPatch().String()
```
olblak, line 280: What would you think to also catch downgrade scenarios and prerelease/metadata change? 

```
  if newVer.Minor() > origVer.Minor() || newVer.Minor() < origVer.Minor()
``` 

And at the end of the condition, a catchall for situation like prelease or metadata change

else {
  computedVersion = v.IncMinor().String()
}

olblak:
I like the idea behind the pullrequest very much, just a minor comment about downgrade scenarios

pkg/plugins/resources/helm/utils.go
aroberts87: Hey @olblak! I have attempted to address both of these comments in my most recent commits. I did take the liberty of shortening the `if a > b || a < b` to simply `if a != b`, but other than that it's the same.

I added a new field in the `values.yaml` file in the test chart to allow version downgrades without changing any existing tests. I hope that's okay.

olblak: Indeed way easier to read

olblak (approved):

PR #1605, opened 2023-09-18
Allow to share/reuse Updatecli manifest using any OCI registry

### LINKED ISSUE #1600: Allow sharing Updatecli configuration using OCI registry such as ghcr.io

When writing and maintaining Updatecli manifest(s), we end up have to maintain a lot of manifest duplication.

### Solution you'd like

I would like to write once a policy and then share it across many different project/repositories
An OCI registry could be a great solution for that.
The idea would be to reuse the same workflow than for any containers.

**Publish to registry**

A policy is a combination of manifests, values, and secrets files.
The first step is to publish that policy under a specific name.

```
updatecli manifest push --config updatecli.d --values values.yaml --secrets secrets.yaml ghcr.io/updatecli/policies/hugo:v0.1.0
```

We could publish the following Updatecli files:
* manifest using a new `media type` named `application/io.updatecli.manifest`
* values using a new `media type` named `application/io.updatecli.values`
* secrets using a new `media type` named `application/io.updatecli.secrets`

**Pull from registry**

The second step would be to the download policy to the Updatecli temporary direcotry such as /tmp/updatecli/store/ (on Linux)

```
updatecli manifest pull ghcr.io/updatecli/policies/hugo:v0.1.0
```

**Execute policy**

The final stage would be to apply/diff the policy such as

```
updatecli apply ghcr.io/updatecli/policies/hugo:v0.1.0
```

or by overriding the default values.yaml

```
updatecli apply --values updatecli/values.yaml ghcr.io/updatecli/policies/hugo:v0.1.0
```

Any values/secrets file specified in the command would override the default policy content

Or multiple policies

```
updatecli apply ghcr.io/updatecli/policies/hugo:v0.1.0 ghcr.io/updatecli/policies/node:v0.3.0
```

Nothing comes to my mind yet

### Anything else?

Similar to docker-compose.yaml, I am considering having a file named `update-compose.yaml` that would contain a list of "policies" such as

.update-compose.yaml
```
policies:
  policy-1:
    name: ghcr.io/updatecli/policies/autodiscovery:latest
  policy-2:
    manifests: updatecli/updatecli.d
    values: updatecli/values.yaml
    secrets: updatecli/secrets.yaml
```

then we could run something like `updatecli compose apply` to apply each policiy one by one

Having a single files containing a list of policies to install would simplify the CI configuration as it would always be the same command to execute

olblak:
Fix #1600

TODO
- [x] Gate this feature behind the experimental flag
- [x] Show proper title when push/pull registry
- [x] Move the oras module update to a proper pullrequest

This pullrequest allows to: 

* Push an update policy to an OCI registry
* Pull an update policy from an OCI registry
* Apply/Show/Diff update policy from an OCI registry

A policy can be composed of manifest, values, and secrets files
values and secret files can be overridden at runtime similarly the current behavior 
when we specify multiple `--values` or `--secrets`

I also did a few code cleanup:
* I took the opportunity to split the engine package to split the main file which became huge to multiple one as it became very hard to read.
* I updated `oras-project/oras-go` to `1.2.4`. This dependency was needed for go-container-test and several other dependency in the container space like moby or helm. It potentially introduces a breaking change for older version  of those dependencies. I had a quick look and it shouldn't affect the code we use. This version bump  will also unblock PR like https://github.com/updatecli/updatecli/pull/1578

## Test

```shell
go build -o bin/updatecli
./bin/updatecli manifest push --config updatecli/updatecli.d/updatecli.yaml -t ghcr.io/olblak/policies/updatecli:latest .
./bin/updatecli manifest pull --disable-tls localhost:5000/myrepo:latest
./bin/updatecli manifest show --disable-tls ghcr.io/olblak/policies/updatecli:latest
./bin/updatecli diff --disable-tls ghcr.io/olblak/policies/updatecli:latest
```

## Additional Information

This feature relies on the `docker login` to create the credential store, updatecli doesn't provide login mechanism

Secret or Values files specified at command execution will override default values publish on the registry.
Similar to doing `updatecli diff --config manifest.yaml --values default.yaml --values override.yaml` 
File are not merged using deep merged 

### Potential improvement

while looking at the cmd package, I realize how none consistent the command and I wonder if 

`updatecli diff` should become `updatecli pipeline diff`
`updatecli apply` should become `updatecli pipeline apply`

or rename 

`updatecli diff` should become `updatecli policy diff`
`updatecli apply` should become `updatecli policy apply`


olblak:
I am starting to wonder if I really want to gate this feature behind the experimental flag

olblak:
Well maybe the oci media type...

olblak:
Well the oras library is bumped with the Helm dependency update in https://github.com/updatecli/updatecli/pull/1626

PR #1616, opened 2023-09-25
Run helm chart metadata update only on a changed target

### LINKED ISSUE #1614: Helm chart using versionincrement always bump the chart version

### Current Behavior

https://github.com/updatecli/charts/pull/3

The chart is always bump even if it shouldn't :( 

### Expected Behavior

Only bump a helm chart if the target was effectively changed

### Steps To Reproduce

Using the helm chart auto-discovery without any change to apply

### Environment

```markdown
- OS: linux
- updatecli:v0.60.1, v0.60.0
```

### Anything else?

_No response_

olblak:
Fix #1614 

We only want to update the Chart metadata if the chart has been modified during the current target execution.
To make this process more idempotent in the context of a SCM, we could also check if the helm chart
has been modified during one of the previous target execution by comparing the current chart versus the one defined
on the source branch. But the code complexity induced by this check is probably not worth the effort today.

## Test

```shell
cd pkg/plugins/resources/helm/ 
go test
```

### DIFF
--- a/pkg/plugins/resources/helm/target.go
+++ b/pkg/plugins/resources/helm/target.go
@@ -45,9 +45,17 @@ func (c *Chart) Target(source string, scm scm.ScmHandler, dryRun bool, resultTar
 		chartPath = filepath.Join(scm.GetDirectory(), c.spec.Name)
 	}
 
-	err = c.MetadataUpdate(resultTarget.NewInformation, scm, dryRun, resultTarget)
-	if err != nil {
-		return fmt.Errorf("unable to update chart metadata: %s", err)
+	/*
+	  We only want to update the Chart metadata if the chart has been modified during the current target execution.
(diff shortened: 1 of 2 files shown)


dduportal (approved):
👍

dduportal (approved):
Early returns FTW! 👏

olblak:
Thanks I slightly changed the PR to also update the appVersion field update even if the target didn't change.

dduportal:
> Thanks I slightly changed the PR to also update the appVersion field update even if the target didn't change.

I'm not sure to understand the use case (but I don't use updatecli to track appversion so I guess it should be fine?).

Currently trying a local build of https://github.com/updatecli/updatecli/pull/1616/commits/3c9e749f1067ecf478d5cd5da96a8599a6ff7089

PR #1621, opened 2023-09-26
feat: terraform autodiscovery

### LINKED ISSUE #534: Feature Request: terraform support

Terraform modules can get far behind.

### Solution you'd like

I'd like to use updatecli to update two different things for a Terraform workspace.

### Terraform Providers
The first is Terraform providers, which provide backends for Terraform code. These are usually APIs such as AWS, Azure, or others such as Kubernetes, and many more. A full list is on the [Terraform registry](https://registry.terraform.io/browse/providers). As far as the source code that needs to be updated, I think we can simply target terraform 0.13 and up which allows you to specify providers as below:

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 3.0"
    }
  }
}
```

With this, we can target the version and see what the latest "source" provider is using the registry API. E.g. for `hashicorp/aws` provider, we can query the search:
```
$ curl 'https://registry.terraform.io/v2/providers?filter%5Bnamespace%5D=hashicorp&filter%5Bname%5D=aws&filter%5Bunlisted%5D=true&filter%5Bwithout-versions%5D=true' | jq
{
  "data": [
    {
      "attributes": {
...
      },
      "id": "323",
      "links": {
        "self": "/v2/providers/323"
      },
      "type": "providers"
    }
  ],
  "links": {
...
  },
  "meta": {
...
  }
}

```
This will give us an ID, which we can then use to query all the versions:
```
curl 'https://registry.terraform.io/v2/providers/323?include=provider-versions' | jq
```

We can use [Hashicorp's HCL](https://github.com/hashicorp/hcl) package to parse and write the new version. This is the package that is used by Terraform itself.

### Terraform Modules
On the other side, there is also Terraform modules. Modules are essentially bundles of Terraform code (resources, data sources etc.) that are combined into a single module code. An example of a module call is:

```hcl
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "3.11.5"
  # other attributes for the aws vpc module.
}
```

They here is retrieving the `version` attribute (or the `source` attribute in certain cases), and subsequently retrieving the latest version accordingly. Getting the latest version of a module is a bit more nuanced than the provider versions, because modules can be retrieved from different places. Git repos, Terraform registry, Terraform Enterprise private registry, Terraform Cloud private registry, [Terraform's public registry](https://registry.terraform.io/browse/modules) and likely other places I'm not aware of.

Terraform internally uses Hashicorp's [go-getter](https://github.com/hashicorp/go-getter) package, which allows for parsing the the `source` attribute as above. We can utilise it to figure out where to get the latest version from, and subsequently retrieve the relevant latest version, and compare. The similar to the providers, we can the HCL package to the write it back to file.

_No response_

### Anything else?

_No response_

mcwarman:
Fix #534

Add terraform autodiscovery support for lock files.

## Test

```shell
cd pkg/plugins/autodiscovery/terraform/
go test ./...
```

```yaml
name: Terraform Auto Discovery

autodiscovery:
  crawlers:
    terraform:
      platforms:
        - linux_amd64
        - linux_arm64
        - darwin_amd64
        - darwin_arm64
```

### Potential improvement

* Support `terraform/providers`


olblak:
While this a great PR, I have  one important change and a few questions

**Jsonschema** 

The variable `AutodiscoverySpecsMapping`: https://github.com/updatecli/updatecli/blob/91b2be178f4f1a42e9c7c2e15c5bbfc07f11a4f2/pkg/core/pipeline/autodiscovery/main.go#L36

needs the mapping for Terraform similarly to the other entries, that variable is used to generate the Jsonschema spec which is ultimately used to generate the website documentation and IDE.

**Default Autodiscovery**

The variable `DefaultCrawlerSpecs` could have terraform listed there, that variable contains the default autodiscovery plugins we want to execute when we run `updatecli diff --experimental`. Note that every autodiscovery plugins shouldn't necessary be there as we may have collusion between plugin but I think it could be interesting in the current situation.

**Platform** 

I did a quick test with the following manifest on https://github.com/jenkins-infra/azure

```
autodiscovery:
  crawlers:
    terraform:
```

and it complained about missing platform, I had to use 

```
autodiscovery:
  crawlers:
    terraform:
      platforms:
        - linux_amd64
        - linux_arm64
        - darwin_amd64
        - darwin_arm64
```

It's been a while since I had to maintain Terraform and I don't remember the purpose of that platform setting, could it be detected? If not that's ok but then we shouldn't add the terraform plugin to the default list of plugin execute as it will always fail

mcwarman:
**AutodiscoverySpecsMapping**

I'll add that.

**Default Autodiscovery**

It can be added, but only if we solve Platform

**Platform**

When running `terraform init` it will detect the platform you are running on you can then add hashes for that. You can add additional platforms using `terraform providers lock` command.

An option is provide a sensible default maybe just `linux_amd64` if nothing is specified.

olblak:
Thanks for the information, 

> An option is provide a sensible default maybe just linux_amd64 if nothing is specified.

That's what I initially tried but it was confusing because Updatecli reported some changes such as bump version from "1.0.0" to "1.0.0" basically I had to specify the four platforms to show me the correct result.

If it's not possible to detect what platform is in used within the lock then I guess the best approach is to not run the terraform autodiscovery by default

mcwarman:
> That's what I initially tried but it was confusing because Updatecli reported some changes such as bump version from "1.0.0" to "1.0.0" basically I had to specify the four platforms to show me the correct result.

Yeah that would be because the hashes would have been removed for the other platforms, there's probably some additional logging required for when that happens. 

> If it's not possible to detect what platform is in used within the lock then I guess the best approach is to not run the terraform autodiscovery by default

Detecting platforms would be expensive as I don't think there's anyway to look up hashes origin without downloading all of the artefacts and cross checking them.

Should there be a default added so the empty config works?

```yaml
autodiscovery:
  crawlers:
    terraform:
```

olblak:
I think I would do something like 

```
if len(platforms) == 0 { 
  logrus.Warningf("No platforms specified, we fallback to linux_amd64, linux_arm64, darwin_amd64, darwin_arm64")
  set default value
}
```

mcwarman:
Also added log message:

```
⚠ - "hashicorp/random" already set to "3.5.1" hashes for the provider have been updated in file ".terraform.lock.hcl"
```

pkg/core/pipeline/autodiscovery/main.go
```diff
@@ -40,6 +42,7 @@ var (
 		"golang/gomod":  golang.Spec{},
 		"helm":          &helm.Spec{},
 		"helmfile":      &helmfile.Spec{},
+		"terraform":     &terraform.Spec{},
```
mcwarman, line 45: I added this is while we were discussing i'm guessing it should now be removed due to platforms.

olblak: This is one is needed to generate the jsonschema, and the one line 31 is useful as I can just run `updatecli diff --experimental` to see what would be changed.

I now often run `updatecli diff --experimental` when I look at a repository as it tells me how "outdated" the project is

olblak:
Thanks for the quick feedback, I just did some testing and it works great

olblak (approved):

olblak:
Thanks @mcwarman for the PR

olblak:
@mcwarman  I almost forgot but that would be great to have some documentation on https://github.com/updatecli/website/tree/master/content/en/docs/plugins/autodiscovery

mcwarman:
@olblak Added the following:

* #1627 - The curse of copy and paste
* https://github.com/updatecli/website/pull/1162

PR #1695, opened 2023-10-13
fix: fallback to updatecli.yaml or updatecli.d if no manifest specified

### LINKED ISSUE #1694: Config not being picked up in updatecli.d or updatecli.yaml

### Current Behavior

From version `v0.63.0` configuration are not loaded from `updatecli.d` directory or file `updatelci.yaml`.

### Expected Behavior

Configuration files in `updatecli.d` directory  or file `updatelci.yaml` should be loaded

### Steps To Reproduce

From the updatecli repository

1. `mkdir updatecli.d`
2. Create a pipeline `updatecli.d/test.yaml`
3. `updatecli diff`

```yaml
# updatecli.d/test.yaml
sources:
  JenkinsOldestSupportedWeekly:
    name: Get the oldest supported Jenkins weekly version
    kind: json
    spec:
      file: https://updates.jenkins.io/tiers.json
      key: "weeklyCores.[0]"
```

NOTE: This still works: `updatecli diff --config ./updatecli.d/test.yaml`

### Environment

```markdown
- OS: Ubuntu 20.04
- updatecli: v0.63.0
```

### Anything else?

Tested in v0.62.0 and works as expected

olblak:
Fix #1694 

## Test

I  build the binary running `go build -o bin/updatecli . ` then I executed from a directory:
* with a file named updatecli.yaml then it tried to load it
* with a directory named updatecli.d then it tried to load manifests from it
* with both a file named updatecli.yaml and updatecli.d then it tried to load both manifests
* without a file named `updatecli.yaml` or `updatecli.d` then it tried to load autodiscovery


pkg/core/engine/configuration.go
```diff
@@ -17,8 +18,20 @@ func (e *Engine) LoadConfigurations() error {
 
 	for i := range e.Options.Manifests {
 		if e.Options.Manifests[i].IsZero() {
-			ErrNoManifestDetectedCounter++
-			continue
+
+			// If no manifest file is specified, we try to detect one
+			if _, err := os.Stat("updatecli.yaml"); err == nil {
+				e.Options.Manifests[i].Manifests = append(e.Options.Manifests[i].Manifests, "updatecli.yaml")
```
mcwarman, line 24: Should this be used? 

https://github.com/updatecli/updatecli/blob/9344acd1230cfa7d58c61f2fa2312f6b56a3425a/pkg/core/config/main.go#L35

pkg/core/engine/configuration.go
```diff
@@ -17,8 +18,20 @@ func (e *Engine) LoadConfigurations() error {
...
+			}
+
+			if _, err := os.Stat("updatecli.d"); err == nil {
+				e.Options.Manifests[i].Manifests = append(e.Options.Manifests[i].Manifests, "updatecli.d")
```
mcwarman, line 28: And this?

https://github.com/updatecli/updatecli/blob/9344acd1230cfa7d58c61f2fa2312f6b56a3425a/pkg/core/config/main.go#L37

mcwarman (dismissed):
Tested the 4 scenarios working as expected. 

Also retested project where found #1694 and working as expected.

pkg/core/engine/configuration.go
olblak: Thanks a lot for your review, you made a good catch.
The code was there but not reachable anymore.

I have additional code cleanup to do.

pkg/core/engine/configuration.go
olblak: Same comment than above.

olblak:
I reorganize a bit the code as I find it clearer now.
After additional thought, the constant should never have been in a different package, I lost track of that.

mcwarman:
The fact it was there and not reachable makes more sense than it getting removed.

I found the constants when I was first looking into the bug.

PR #1713, opened 2023-10-21
fix: revert pipelineid requirement

olblak:
As discussed on Matrix, it's annoying having to specify the pipelineid.
I propose to remove this requiremement  to be optionnal.

Now instead of using the manifest file checksum, we'll use the manifest `name` to generate the pipelineid if it's undefined
More info on https://github.com/orgs/updatecli/discussions/1692

## Test

```shell
cd <to_package_directory>
go test
```

### DIFF
--- a/pkg/core/config/main.go
+++ b/pkg/core/config/main.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/core/config/main.go
```diff
@@ -288,7 +289,20 @@ func New(option Option) (configs []Config, err error) {
...
 		// config.PipelineID is required for config.Validate()
 		if len(configs[id].Spec.PipelineID) == 0 {
-			configs[id].Spec.PipelineID = pipelineID
+			logrus.Debugln("pipelineid undefined, we'll try to generate one")
+
+			// If pipeline name is defined then we use it to generate a pipeline id
+			// there is less change for the pipeline name to change.
+			switch configs[id].Spec.Name {
+			case "":
+				logrus.Debugln("pipeline name undefined, we'll use the manifest filechecksum")
```
olblak, line 298:
```suggestion
				logrus.Debugln("pipeline name undefined, we'll use the manifest file checksum")
```

dduportal (approved):
Thanks!

Issue #1026, opened 2022-12-09
Feature Request: Support multiple files with `spec: files` in `conditions` of `kind:file`


macedogm:
### Is your feature request related to a problem?

As discussed in https://github.com/updatecli/updatecli/discussions/1013, Updatecli already has the logic to support multiple files with `spec: files` in `conditions` of `kind:file`. However, this functionality is not allowed due to a check of the number of files:

https://github.com/updatecli/updatecli/blob/1c37a46e5592eb0397bd5cfd96d95a7459da7d2f/pkg/plugins/resources/file/condition.go#L53-L58

### Solution you'd like

Would be nice to have this feature so we can write more concise pipelines. Example:

```yaml
conditions:
  dockerfiles:
    name: "Bump golang:alpine image version in Dockerfiles"
    kind: "file"
    scmid: "k3s"
    disablesourceinput: true
    spec:
      files:
        - "Dockerfile.dapper"
        - "Dockerfile.test"
        - "Dockerfile.manifest"
      matchpattern: 'ARG GOLANG=golang:v?\d+\.\d+\.\d+(-alpine\S+)?'
```

### Alternatives you've considered

To work around such limitation, the user needs to write a verbose and repetitive pipelines. Example:

```yaml
conditions:
  dockerfile-dapper:
    name: "Check if 'ARG GOLANG' is set in Dockerfile.dapper"
    kind: "dockerfile"
    scmid: "k3s"
    sourceid: "golang-alpine-docker-image"
    spec:
      file: "Dockerfile.dapper"
      instruction:
        keyword: "ARG"
        matcher: "GOLANG"
  dockerfile-test:
    name: "Check if 'ARG GOLANG' is set in Dockerfile.test"
    kind: "dockerfile"
    scmid: "k3s"
    sourceid: "golang-alpine-docker-image"
    spec:
      file: "Dockerfile.test"
      instruction:
        keyword: "ARG"
        matcher: "GOLANG"
```

Another option would be to add support to `spec: files` to `kind: "dockerfile"`, but this would require extra code.

PR #1739, opened 2023-11-04
feat(dockerfile) support multiple files

dduportal:
This PR adds the support of multiple files for the resources of kind `dockerfile` (conditions and targets).

Use cases:
- It is an additional feature to help for #1026.
- Also, on the Jenkins project, we have cases were we want to update the same instruction to multiple files: https://github.com/jenkinsci/docker-agent/blob/1182cad9d14ece000ab38433eee05daf1872a993/updatecli/updatecli.d/remoting.yaml#L37-L81

## Test

```shell
cd pkg/plugins/resources/dockerfile
go test
```

Please note that a new e2e test as added for `dockerfile` which tests the current `file` (for non regression) and the new `files`.

I've tested this E2E test locally with a `diff` and an `apply`. The only untested part is when using with an SCM.

### Tradeoff

Using autodiscovery would help, of course but we want to control the list of files for now.

### Potential improvement

N.A.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -57,3 +57,7 @@ joho
 godotenv
 filechecksum
 errdef
+mcr
+nanoserver
+ltsc
+windowscore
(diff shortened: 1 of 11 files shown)


olblak (approved):
Great pulllrequest, I just did some testing locally and it works great 🎆

dduportal:
Associated documentation PR: https://github.com/updatecli/website/pull/1207

PR #1740, opened 2023-11-05
chore(actions) avoid code duplication for report's PipelineURL generation

dduportal:
Follow up of #1737 (see https://github.com/updatecli/updatecli/pull/1737#discussion_r1382388606)

This PR only refactor the code to avoid duplication, by introducing a new `ci` package.
Allows extensibility with custom  behaviors in the future.

Please note the validation of each CI has been decreased: we can add more validation but it looks like a lot of code for something not really probable. 

## Test

```shell
cd pkg/plugins/utils/ci
go test
```

### DIFF
--- a/pkg/plugins/utils/ci/main.go
+++ b/pkg/plugins/utils/ci/main.go
@@ -1,0 +1,29 @@
+package ci
+
+import (
+	"fmt"
+	"os"
+)
+
+// CIEngine is an interface which allows to detects based on environment variable if Updatecli is executed from a CI environment like Jenkins or GitLab CI
+type CIEngine interface {
+	URL() string
+	Name() string
+}
+
+// New returns a newly initialized CIEngine or an error
+func New() (ci CIEngine, err error) {
+	if os.Getenv("JENKINS_URL") != "" {
+		return Jenkins{}, nil
+	}
+
+	if os.Getenv("GITLAB_CI") != "" {
+		return GitLabCi{}, nil
+	}
+
+	if os.Getenv("GITHUB_ACTION") != "" {
+		return GitHubActions{}, nil
+	}
+
+	return nil, fmt.Errorf("Unknown CI Engine.")
+}
--- a/pkg/plugins/utils/ci/main_test.go
+++ b/pkg/plugins/utils/ci/main_test.go
@@ -1,0 +1,60 @@
+package ci
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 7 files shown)


pkg/plugins/utils/ci/main.go
```diff
@@ -0,0 +1,29 @@
+package ci
+
+import (
+	"fmt"
+	"os"
+)
+
+// CIEngine <TODO>
```
olblak, line 8:
```suggestion
// CIEngine is an interface which allows to detects based on environment variable if Updatecli is executed from a CI environment like Jenkins or GitLab CI
```

olblak (approved):

olblak:
Thanks @dduportal for the improvement, I just added an interface comment

olblak:
@dduportal Just in case you missed it, test are not passing :)

dduportal:
> @dduportal Just in case you missed it, test are not passing :)

Weird, this test passes locally. I wonder if it could be related to the test `SetEnv` variable not having the same behavio. I'm putting the PR as draft and I'll send a debug commit.

If I don't get any result, I'll remove the test itself.

pkg/plugins/utils/ci/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	tests := []struct {
+		name    string
+		env     map[string]string
+		expect  CIEngine
+		wantErr error
+	}{
+		{
+			name:    "Unknown",
+			wantErr: fmt.Errorf("Unknown CI Engine."),
```
olblak, line 21: This test can't work from GitHub Action because `GITHUB_ACTION` is already set so we'll never reach  the final return

olblak: And the Jenkins and Gitlab test will only work as long as we don't reorder the if conditions

dduportal: ooooh good catch! I need to unset variables prior to the tests \o/ Thanks!

olblak: I wouldn't "unset" environment varialbe used by CI as it may be hard in the futur to debug CI issues
 I think in this case it's just fine to no test it

dduportal:
Fixed: checks are passing except the `updatecli` one which fails for the same reason as the main branch check @olblak

olblak (approved):
Now the test should passed, I am just waiting for the api rate limit to be reset

Issue #1732, opened 2023-10-29
Error Cloning repository using GitHub SCM Plugin


treezio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This scenario is kind of weird.
I'm currently setting up the updatecli pipeline for a "big" repository based on GitHub Actions, using the `GITHUB_TOKEN` as authentication.
Even though I have been able to clone many private repositories, I'm facing errors when attempting to clone this GitHub private repository using the `GitHub` SCM Plugin. 
I have recreated the errors in my local using an Ubuntu docker Image.

The main issue is that although I'm setting the `GITHUB_TOKEN` environment variable, it seems like it is trying to clone the repository using the SSH protocol, and it is complaining about not being able to create an SSH Agent.

```
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "./path/to/app/updatecli.yaml"

SCM repository retrieved: 1
ERROR: failed cloning GitHub repository "https://github.com/organization/repository.git"
ERROR: err - error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"
```

I've tried to dive a bit in the `go-git` library but I have no clue why this is actually happening.

Maybe I'm missing something but I think this is not related to authentication errors, as I have been able to clone the repository using the `git` cli and the `GITHUB_TOKEN` via `https://` URL 

Pipeline Config:

```
---
name: Bump aplication chart version

scms:
  default:
    kind: github
    spec:
      user: "{{ .github.user }}"
      email: "{{ .github.email }}"
      owner: "{{ .github.owner }}"
      repository: "{{ .github.repository }}"
      token: "{{ requiredEnv .github.token }}"
      username: "{{ .github.username }}"
      branch: "{{ .github.branch }}"
```

Config values:
```
github:
  user: "SomeUser"
  email: "mail@organization.es"
  username: "treezio"
  token: "GITHUB_TOKEN"
  branch: "master"
  owner: "organization"
  repository: "repository"
```

I've tried this very same configuration just changing the `repository` value and it works.

### Expected Behavior

The repository is cloned via API using the authentication token `GITHUB_TOKEN`  

### Steps To Reproduce

- Run ubuntu 22.04 docker image in repository directory
- `docker run -ti -v $(pwd):/opt ubuntu:22.04`
- Install updatecli
- export `GITHUB_TOKEN` 
- Run `updatecli diff`

[...]

PR #1758, opened 2023-11-11
feat(git): allow to decide if we want to fetch submodules or not

mavimo:
Fix #[1732](https://github.com/updatecli/updatecli/issues/1732)

If will allow to define if the submodule need to be fetched or not when a repository is initialised.

## Test

```shell
pushd pkg/plugins/scms
go test
popd

pushd pkg/plugins/utils/gitgeneric
go test
popd
```

### DIFF
--- a/pkg/plugins/scms/gitea/main.go
+++ b/pkg/plugins/scms/gitea/main.go
@@ -114,8 +114,8 @@ type Spec struct {
 			If you need to push changes to a specific branch, you must use the plugin "git" instead of this
 	*/
 	Branch string `yaml:",omitempty"`
-	// Whether to checkout submodules: `true` to checkout submodules or `recursive` to recursively checkout submodules.
-	Submodules bool `yaml:",omitempty"`
+	// Whether to checkout submodules: `true` to checkout submodules or `false` to skip.
+	Submodules *bool `yaml:",omitempty"`
 }
 
 // Gitea contains information to interact with Gitea api
--- a/pkg/plugins/scms/gitea/main.go
+++ b/pkg/plugins/scms/gitea/main.go
@@ -114,6 +114,8 @@ type Spec struct {
 			If you need to push changes to a specific branch, you must use the plugin "git" instead of this
 	*/
 	Branch string `yaml:",omitempty"`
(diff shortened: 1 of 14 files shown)


olblak:
Thanks @mavimo for opening this pullrequest.

I recently stumbled on this problem and I asked my self if it was needed and you confirm that it is.

olblak:
Something worth mentioning, we always try to avoid breaking changes between Updatecli releases so something we use is the feature flag `experimental` like [here](https://github.com/updatecli/updatecli/blob/79d806239960a845dd97b392bf3dee9439c3f627/pkg/core/engine/autodiscovery.go#L52)

So if we need to introduce a breaking change, we gate it before the experimental flag for a few release while we communicate about it.

mavimo:
@olblak I was thinking more to keep the current behaviour as default (aka by default is "true"), more than including a flag for that, whyt? 
At the moment the only way to use this approach is to use a nullable pointer in the yaml unmarshaling (unless we include something like https://github.com/creasty/defaults ), not sure if we want to move on this direction (I push a change in the code here)

olblak:
> @olblak I was thinking more to keep the current behaviour as default (aka by default is "true"), more than including a flag for that, whyt? At the moment the only way to use this approach is to use a nullable pointer in the yaml unmarshaling (unless we include something like https://github.com/creasty/defaults ), not sure if we want to move on this direction (I push a change in the code here)

Great then, I didn't think about using a pointer to keep the current behavior :)

pkg/plugins/scms/gitea/main.go
```diff
@@ -114,6 +114,8 @@ type Spec struct {
 			If you need to push changes to a specific branch, you must use the plugin "git" instead of this
 	*/
 	Branch string `yaml:",omitempty"`
+	// Whether to checkout submodules: `true` to checkout submodules or `false` to skip.
+	Submodules *bool `yaml:",omitempty"`
```
olblak, line 118: I really like this approach of using a pointer to control if the default value must be false or true

olblak:
Nice pullrequest. It's really awesome the effort put on the testing.
I'll wait for the test to pass and merge unless something goes wrong

olblak (approved):
All things appears good, thanks for the pullrequest

PR #1761, opened 2023-11-12
chore: remove usage of io/ioutil package as deprecated since Go 1.16

mavimo:
Remove `io/ioutil`, the package is deprecated since Go 1.16, see https://pkg.go.dev/io/ioutil 

## Test

No specific test required


olblak:
Thanks for the cleanun

olblak (approved):

PR #1762, opened 2023-11-12
chore: remove helper package that is not used anymore

mavimo:
Remove package that is not used anymore.

## Test

No specific test needed.


olblak (approved):
Thanks, I love old and unused code being removed

PR #1773, opened 2023-11-18
feat: introduces a new resource "http" (source only)

### LINKED ISSUE #268: New resource of type "http"

I'm trying to build an updatecli workflow that retrieve the maven latest version, and I would like to add a condition that checks if the tarball of the latest version exists (same idea as checking if a Docker image is published but it's a file at a given URL).

There are no way to do it, except using a condition of type `file` and a transformer that empties the buffer but it's hacky AND slow.

I would like a new resource of type `http` that would allow the following:

- Source: Retrieve the body or the header of an URL
- Condition: Check if an URL exists (e.g. answer a certain type of HTTP code that could be defined) or if it has a given header defined (key or value)
- Target: POST/PUT the source to change an URL

I could use the "shell" (#264)  resource with  `curl` or`wget`

**Additional context**

dduportal:
Implements #268

This PR introduces a new resource of type `http` with source and condition capability.

The initial UX proposal has been updated to ensure separation between source and condition are easier (see the E2E test).

Here is a "complete reference example":

```
sources:
  getWithCustomRequest:
    kind: http
    spec:
      url: https://repo.jenkins-ci.org/releases/org/jenkins-ci/main/jenkins-war/maven-metadata.xml
      request:
        headers:
          Authorization: 'Bearer Token'
          Accept: 'application/xml'
        verb: GET

conditions:
  checkWithCustomRequest:
    kind: http
    sourceid: getJenkinsWarArtifactMetadatas
    spec:
      url: https://repo.jenkins-ci.org/releases/org/jenkins-ci/main/jenkins-war/maven-metadata.xml
      request:
        headers:
          Authorization: 'Bearer Token'
          Accept: 'application/xml'
        verb: HEAD
      responseasserts:
        statuscode: 302
        headers:
          Content-Type: "text/html; charset=utf-8"
```

## Test

```shell
cd ./pkg/plugins/resources/updateclihttp/
go test
```

## Additional Information

- For a source, both HTTP/4xx and HTTP/5xx response status code are considered a failure (as both client and server side errors are not expected to provide a body). This behavior can be adapted on a subsequent PR if we have a use case to retrieve a body on a HTTP/404 error.
- For a condition there are 2 cases:
  - If no assertion is provided, HTTP/5xx is considered an error, while HTTP/4xx is considered a non passing condition.
  - If assertion is provided, then the status code is compared to the provided status code assertion

### DIFF
--- a/pkg/plugins/resources/cargopackage/spec.go
+++ b/pkg/plugins/resources/cargopackage/spec.go
@@ -5,7 +5,7 @@ import (
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
-// Spec defines a specification for a "dockerimage" resource
+// Spec defines a specification for a "cargopackage" resource
 // parsed from an updatecli manifest file
 type Spec struct {
 	// !deprecated, please use Registry.URL
--- a/pkg/plugins/resources/updateclihttp/main.go
+++ b/pkg/plugins/resources/updateclihttp/main.go
@@ -17,8 +17,12 @@ type Http struct {
 	httpReq    *http.Request
 }
 
-// New returns a reference to a newly initialized (TODO)
-// or an error if the provided Spec triggers a validation error.
+/*
+*
+New returns a reference to a newly initialized Http resource
+or an error if the provided Spec triggers a validation error.
+*
+*/
 func New(spec interface{}) (*Http, error) {
 
 	newSpec := Spec{}
--- a/pkg/plugins/resources/updateclihttp/main.go
+++ b/pkg/plugins/resources/updateclihttp/main.go
@@ -1,0 +1,89 @@
+package updateclihttp
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/core/httpclient"
+)
+
+// Http defines a resource of type "http"
+type Http struct {
+	spec       Spec
+	httpClient httpclient.HTTPClient
+	httpReq    *http.Request
+}
+
+// New returns a reference to a newly initialized (TODO)
+// or an error if the provided Spec triggers a validation error.
+func New(spec interface{}) (*Http, error) {
+
+	newSpec := Spec{}
+
+	err := mapstructure.Decode(spec, &newSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	if newSpec.Url == "" {
+		return nil, fmt.Errorf("spec.url is not set but required.")
+	}
+
+	if newSpec.ResponseAsserts.StatusCode != 0 || len(newSpec.ResponseAsserts.Headers) > 0 {
+		// This resource is a condition as the asserts are specified
+		if newSpec.ReturnResponseHeader != "" {
+			// Cannot be both source and condition
+			return nil, fmt.Errorf("Cannot define both spec.responseasserts (source only) and spec.responseasserts (condition only).")
+		}
+	}
+
+	httpClient := http.DefaultClient
+	httpClient.Transport = httpclient.NewThrottledTransport(1*time.Second, 1, http.DefaultTransport)
+
+	// Do not follow redirect as per https://pkg.go.dev/net/http when we want to get a header from original request
+	httpClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		if newSpec.ReturnResponseHeader != "" || newSpec.Request.NoFollowRedirects || newSpec.ResponseAsserts.StatusCode > 0 || len(newSpec.ResponseAsserts.Headers) > 0 {
+			logrus.Debugf("spec.returnresponseheader defined: HTTP client won't follow redirects")
+			return http.ErrUseLastResponse
+		}
+
+		return nil
+	}
+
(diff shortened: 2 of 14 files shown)


pkg/plugins/resources/cargopackage/spec.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
-// Spec defines a specification for a "dockerimage" resource
+// Spec defines a specification for a "cargopackage" resource
```
dduportal, line 8: Caught while reading code. Added to this (unrelated) PR as it is only a comment (e.g. no impact on code at all)

pkg/plugins/resources/updateclihttp/main.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Http defines a resource of type "http"
+type Http struct {
+	spec       Spec
+	httpClient httpclient.HTTPClient
+	httpReq    *http.Request
+}
+
+// New returns a reference to a newly initialized (TODO)
```
olblak, line 20: Is the "TODO" a left over?

dduportal: oh I forgot to fill it. Let me update, thanks for the catch!

pkg/plugins/resources/updateclihttp/spec.go
```diff
@@ -0,0 +1,32 @@
+package updateclihttp
+
+// Spec defines a specification for a "http" resource
+// parsed from an updatecli manifest file
+type Spec struct {
+	// [S][C] Specifies the URL to use for the HTTP request of this resource
```
olblak, line 6: You may have notice but I started using multiline comment `/* */` so I can provide more information to the generated jsonschema which ultimately benefit to website and IDE integration

You have an example here 
-> https://github.com/updatecli/updatecli/blob/8490900bab55b2a7c153a5f94c53faec6aded5f3/pkg/plugins/resources/yaml/main.go#L18

Which render like 
-> https://www.updatecli.io/docs/plugins/resource/yaml/

It's a personnal taste but I like multiline comments as I find them easier read 

Remark:

The only thing to keep in mind is that I noticed that sometimes go fmt reformat comment using tab instead of space which doesn't render the same in IDE

pkg/plugins/resources/updateclihttp/target.go
```diff
@@ -0,0 +1,11 @@
+package updateclihttp
+
+import (
+	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
+	"github.com/updatecli/updatecli/pkg/core/result"
+)
+
+// Target
```
olblak, line 8:
```suggestion
// Target is not implemented. If you ever feel the need, you can still open a GitHub issue with a valid usecase.
```

pkg/plugins/resources/updateclihttp/spec.go
olblak: By the way, not all plugins use the multi line approach, I usually migrate them one by one when I work on them

pkg/plugins/resources/updateclihttp/target.go
olblak: I don't see a target usecase so 🤷🏾

olblak (approved):
This plugin is awesome. I tested locally and it works great.
I also like the unit tests

I made a few suggestion to the plugin comment as I think they can be improve a bit but it's not a blocker

I am just wondering how we should communicate if it's better to fetch content using the `http` plugin or the `file` with the http protocol in the path

Feel free to resolve the conversations and the merge whenever you want

pkg/plugins/resources/updateclihttp/spec.go
dduportal: Treated in https://github.com/updatecli/updatecli/pull/1773/commits/494ac44c90ba03bc98496c37b724efa2a894ff8c

pkg/plugins/resources/updateclihttp/main.go
dduportal: Treated in https://github.com/updatecli/updatecli/pull/1773/commits/494ac44c90ba03bc98496c37b724efa2a894ff8c

dduportal:
> This plugin is awesome. I tested locally and it works great. I also like the unit tests
> 
> I made a few suggestion to the plugin comment as I think they can be improve a bit but it's not a blocker

Thanks! I've taken your suggestions. If you see further improvement (on the comment particularly) don't hesitate to comment the PR, even if merged: i'll send a subsequent PR if need be.


> I am just wondering how we should communicate if it's better to fetch content using the `http` plugin or the `file` with the http protocol in the path

Proposal, WDYT?

- Recommended to use `http` for cases where the whole body is required and/or when the HTTP request need to be customized (e.g. adding headers for authentication/cache/etc.)
- Recommended to use `file` if you want to apply a pattern matching (for instance https://github.com/jenkins-infra/jenkins-infra/blob/297dd8ab1d9b897710f73bb10387c50e7b440df8/updatecli/weekly.d/jenkinscontroller-tools-maven.yaml#L25-L32)

dduportal:
Self merging because:

> Feel free to resolve the conversations and the merge whenever you want 

from @olblak 

and I only changed comments since his last review

PR #1775, opened 2023-11-23
chore: factorize condition result code out of resource plugins

dduportal:
This PR factorizes condition result code out of resource plugins:

- Simplify code for plugins (easier for contributors to write/update resources as they have less things to handle and can focus on the resource code)
- Centralize result object configuration (we can now change resource condition behavior in the core without impacting resources)
- Separate condition result (true or false) from error (unexpected behavior such as network error) 
- Introduces the following behavioral changes (side effects of the PR):
  -  `githubrelease` condition returns `false` but no error anymore when no upstream errors are triggered, with the same textual output as before.
  -  `gittag` condition returns `false` but no error anymore when no upstream errors are triggered, with the same textual output as before.
  -  `golang` condition returns `false` but no error anymore when no upstream errors are triggered, with the same textual output as before
  -  `gomod` condition returns `false` but no error anymore when no upstream errors are triggered, with the same textual output as before
  -  `gomodule` condition returns `false` but no error anymore when no upstream errors are triggered, with the same textual output as before
  - (nit) Homogenize methods to return an error (`error.new()` vs. `fmt.Errorf()`)
  - (cleanup) removing obsolete "ConditionFromScm" tests

## Test

```shell
make test
```

## Additional Information

- Unit test are not checking for the message returned by the `Condition()` function. It wasn't before this PR, so I chose to focus on the refactor to avoid too much changes in this PR

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -110,7 +110,7 @@ A short one-line description of the parameter
 
 ===== 3. Respect the contract
 
-Your 'packageName' must provides a `struct` which implements the `Resource` interface by defining the following functions:
+Your 'packageName' must provide a `struct` which implements the `Resource` interface by defining the following functions:
 
 [cols="1,2a,2", options="header"]
 .Rules
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -110,7 +110,7 @@ A short one-line description of the parameter
 
 ===== 3. Respect the contract
 
-Your 'packageName' must respect at least one of the following interface contract by defining appropriated functions.
+Your 'packageName' must provides a `struct` which implements the `Resource` interface by defining the following functions:
 
 [cols="1,2a,2", options="header"]
 .Rules
@@ -121,34 +121,25 @@ Your 'packageName' must respect at least one of the following interface contract
 
 | Source
 | ```
-type Spec interface {
-    Source() (string, error)
-}
+Source(workingDir string, resultSource *result.Source) error
 ```
 | Defines how a version will be retrieved then passed the following stages
 
 | Changelog
 | ```
-type Changelog interface {
-    Changelog(release string) (string, error)
-}
+Changelog() string
 ```
 | Retrieve the changelog for a specific source.
 
 | Condition
 | ```
-type Spec interface {
-    Condition(version string) (bool, error)
-    ConditionFromSCM(version string, scm scm.Scm) (bool, error)
-}
+Condition(source string, scm scm.ScmHandler) (pass bool, message string, err error)
 ```
 | Define a condition which has to pass in order to proceed
 
 | Target
 | ```
-type Spec interface {
-    Target(source string, scm scm.Scm, dryRun bool) (changed bool, files []string, message string, err error)
-
+Target(source string, scm scm.ScmHandler, dryRun bool, resultTarget *result.Target) error
 ```
 | Define how a target file is updated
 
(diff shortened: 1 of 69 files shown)


mcwarman:
Going to continue to look more at the code as this is the sort of PR that takes multiple passes.

But I remember when I first looked to create a contribution I read this: https://github.com/updatecli/updatecli/blob/main/CONTRIBUTING.adoc#3-respect-the-contract

I realised it was out of date and moved on, in hind sight I should have looked to to update it. The whole page possibly needs a overhaul so might be out of scope for this change but contacts could be updated.

pkg/plugins/resources/awsami/condition.go
```diff
@@ -7,21 +7,21 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
-	"github.com/updatecli/updatecli/pkg/core/result"
 )
 
 // Condition tests if an image matching the specific filters exists.
-func (a *AMI) Condition(source string, scm scm.ScmHandler, resultCondition *result.Condition) error {
+func (a *AMI) Condition(source string, scm scm.ScmHandler) (pass bool, message string, err error) {
+	// func (a *AMI) Condition(source string, scm scm.ScmHandler, resultCondition *result.Condition) error {
```
mcwarman, line 14: Comment left over from refactor?

dduportal: Good catch, thanks!

dduportal:
> Going to continue to look more at the code as this is the sort of PR that takes multiple passes.
> 
> But I remember when I first looked to create a contribution I read this: https://github.com/updatecli/updatecli/blob/main/CONTRIBUTING.adoc#3-respect-the-contract
> 
> I realised it was out of date and moved on, in hind sight I should have looked to to update it. The whole page possibly needs a overhaul so might be out of scope for this change but contacts could be updated.

Good point! I've tried to update a bit in https://github.com/updatecli/updatecli/pull/1775/commits/0d845e01c6046bfb29d86dc64dd43b74679ced6b, WDYT?

pkg/plugins/resources/awsami/condition.go
dduportal: Fixed by https://github.com/updatecli/updatecli/pull/1775/commits/f91437925c08f3ae0065eec232a5ada64b0f70d4

CONTRIBUTING.adoc
```diff
@@ -110,7 +110,7 @@ A short one-line description of the parameter
 
 ===== 3. Respect the contract
 
-Your 'packageName' must respect at least one of the following interface contract by defining appropriated functions.
+Your 'packageName' must provides a `struct` which implements the `Resource` interface by defining the following functions:
```
olblak, line 113:
```suggestion
Your 'packageName' must provide a `struct` which implements the `Resource` interface by defining the following functions:
```

pkg/plugins/resources/awsami/condition_test.go
```diff
@@ -21,18 +20,17 @@ func TestCondition(t *testing.T) {
 		d.ami.apiClient = mockDescribeImagesOutput{
 			Resp: d.mockedResponse,
 		}
-		gotResult := result.Condition{}
-		err := d.ami.Condition("", nil, &gotResult)
+		// gotResult := result.Condition{}
```
olblak, line 23:
```suggestion
```

Shouldn't this comment be removed?

olblak:
I made another pass and I catch a few additional things. It's nice to see a lot of go code being cleanup

pkg/plugins/resources/awsami/condition_test.go
dduportal: absolutely, looks like I forgot to remove it. Thanks!

PR #1776, opened 2023-11-24
chore(tests) test and lint fixes

dduportal:
While working on #1775 , caught a few minor things which I've put in this PR.

I tried to split each one on distinct commit in case we need to NOT add all (or if you want me to split in distinct PRs as well : I don't mind).

Tl;DR;

- Unit ("short") tests should not be requiring Docker Engine
- Some integration tests require Docker Engine: we should fail fast when Docker is not present and running
- Tests doing assertions with `/tmp` must us the `os.TmpDir()` to ensure all contributors can have these tests executed with success on their operating system
- Old validation error message referencing `scmID` instead of `scmid`
- A fix on target commits which are trying to access an undefined field (fixup of #412 )

## Test

```shell
make tests
```

### DIFF
--- a/pkg/plugins/resources/gitbranch/main.go
+++ b/pkg/plugins/resources/gitbranch/main.go
@@ -61,7 +61,7 @@ func New(spec interface{}) (*GitBranch, error) {
 func (gt *GitBranch) Validate() error {
 	validationErrors := []string{}
 	if gt.spec.Path == "" {
-		validationErrors = append(validationErrors, "Git working directory path is empty while it must be specified. Did you specify an `scmID` or a `spec.path`?")
+		validationErrors = append(validationErrors, "Git working directory path is empty while it must be specified. Did you specify a `scmid` or a `spec.path`?")
 	}
 
 	// Return all the validation errors if found any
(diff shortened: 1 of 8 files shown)


pkg/plugins/resources/gitbranch/main.go
```diff
@@ -61,7 +61,7 @@ func New(spec interface{}) (*GitBranch, error) {
 func (gt *GitBranch) Validate() error {
 	validationErrors := []string{}
 	if gt.spec.Path == "" {
-		validationErrors = append(validationErrors, "Git working directory path is empty while it must be specified. Did you specify an `scmID` or a `spec.path`?")
+		validationErrors = append(validationErrors, "Git working directory path is empty while it must be specified. Did you specify a `smcid` or a `spec.path`?")
```
olblak, line 64:
```suggestion
		validationErrors = append(validationErrors, "Git working directory path is empty while it must be specified. Did you specify a `scmid` or a `spec.path`?")
```

olblak (approved):
I know it's still a draft but already great as is

dduportal:
I've set it in ready to review, as I don't think I'll have more things to add for now. I let you drive the merging.

=> Thanks for the typo catch , it was driving me crazy :D
=> I've set the label to "skip-changelog" as I don't think it is worth a changelog entry. Feel free to change it!

PR #1786, opened 2023-11-29
feat(yaml): Add search pattern && fix multiple file condition

olblak:
This pullrequest add the search pattern support for the yaml plugin and fix the yaml multi file condition.

## Test

```shell
cd pkg/plugins/resources/yaml/
go test .
```

### DIFF
--- a/pkg/plugins/resources/yaml/condition.go
+++ b/pkg/plugins/resources/yaml/condition.go
@@ -3,24 +3,24 @@ package yaml
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	goyaml "github.com/goccy/go-yaml"
 	"github.com/goccy/go-yaml/parser"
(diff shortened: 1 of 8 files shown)


pkg/plugins/resources/yaml/condition.go
```diff
@@ -3,99 +3,154 @@ package yaml
...
 	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
-	"github.com/updatecli/updatecli/pkg/core/result"
 
 	"github.com/vmware-labs/yaml-jsonpath/pkg/yamlpath"
 	"gopkg.in/yaml.v3"
 )
 
+var (
+	ErrKeyNotFound = errors.New("key not found")
+)
```
dduportal, line 18: WDYT about defining an `errors.go` file in this package and define this as an explicit errors to export it while following usual go conventions?

PR #1831, opened 2023-12-10
Fix comment that break the generated doc

mavimo:
Fix a comment that break the generated doc

## Test

See [doc](https://www.updatecli.io/docs/plugins/resource/file/#_parameters) for `forcecreate`


olblak:
Good catch, thanks for the pullrequest

olblak (approved):

PR #1854, opened 2024-01-03
fix: shell exitcode

olblak:
While doing some tests, I noticed that the shell exit code return the wrong value

success should be "change return set to false and warning should be change set to true

## Test

```shell
cd pkg/plugins/resources/shell/success/exitcode/
go test
```


dduportal (approved):
Good catch 👍 👏

PR #1899, opened 2024-01-30
feat: add base branch to temporary working branch

### LINKED ISSUE #1897: Feature Request: Different temporary branch per base branch

When using the same Updatecli manifest targeting different branches, Updatecli generates the same temporary branch which results in conflicts.
Today the temporary branch is something like `updatecli_<a hash or pipelineID>`

### Solution you'd like

I would like to include the base branch in the  branch such as 

master: `updatecli_master_<a hash or pipelineID>`
main: `updatecli_main_<a hash or pipelineID>`

Today's approach is to have different `pipelineid` per branch but I find it cumbersome to update each manifests, each time I create a new release branch while we have access to the branch information

### Anything else?

The fix seems to be simple to do but once apply, Updatecli will generates new pullrequest which IMO is not a big deal but worth mentioning

olblak:
Fix #1897 

## Test

I test this pullrequest by running the following commands:

```shell
go build -o bin/updatecli . 
./bin/updatecli apply --config updatecli/updatecli.d/venom.yaml --push=false 
cd /tmp/updatecli/github/updatecli/updatecli/ 
git branch 
```

I should be in the branch named `updatecli_main_venom`

### DIFF
--- a/pkg/plugins/scms/gitea/scm.go
+++ b/pkg/plugins/scms/gitea/scm.go
@@ -14,7 +14,7 @@ func (g *Gitea) GetBranches() (sourceBranch, workingBranch, targetBranch string)
 	targetBranch = g.Spec.Branch
 
 	if len(g.pipelineID) > 0 {
-		workingBranch = g.nativeGitHandler.SanitizeBranchName(fmt.Sprintf("updatecli_%v", g.pipelineID))
+		workingBranch = g.nativeGitHandler.SanitizeBranchName(fmt.Sprintf("updatecli_%s_%s", targetBranch, g.pipelineID))
 	}
 
 	return sourceBranch, workingBranch, targetBranch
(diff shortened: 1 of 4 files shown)


lemeurherve (approved):

olblak:
> ~Maybe add a test with a really long branch name to see if it still works?~
> 
> ~I'm thinking about https://stackoverflow.com/questions/60045157/what-is-the-maximum-length-of-a-github-branch-name but other SCMs might have different limits.~
> 
> Nevermind, already taken care of:
> 
> https://github.com/updatecli/updatecli/blob/86565028ee43bea837034e5bcfec5b92893c23c3/pkg/plugins/utils/gitgeneric/main.go#L687-L689

Thanks for the feedback.
That's indeed something I had a look

PR #1950, opened 2024-02-21
feat: add flux autodiscovery plugin

olblak:
add flux autodiscovery plugin.
Support updating helm chart specified in `helmrelease` configuration and oci tag in `ocirepository` configuration.

- [x] Documentation pullrequest https://github.com/updatecli/website/pull/1338
- [x] Improve matching rule

## Test

```shell
cd pkg/plugins/autodiscovery/flux/
go test .
```

### Potential improvement

* Adding support for `gitrepository`


pkg/plugins/autodiscovery/flux/utils_test.go
```diff
@@ -0,0 +1,35 @@
...
+	f, err := New(
+		Spec{
+			RootDir: "testdata",
+		}, "", "")
+
+	require.NoError(t, err)
+
+	err = f.searchFluxFiles("testdata", defaultFluxFiles[:])
+	if err != nil {
+		t.Errorf("%s\n", err)
```
bigkevmcd, line 20: You can use `t.Error()` here and it'll output a stringified version of the error with a newline.

pkg/plugins/autodiscovery/flux/utils.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// searchFluxFiles will look, recursively, for every files containing a flux helmrelease or helmrepository from a root directory.
+func (f *Flux) searchFluxFiles(rootDir string, files []string) error {
+
+	err := filepath.Walk(rootDir, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			fmt.Printf("prevent panic by handling failure accessing a path %q: %v\n", path, err)
```
bigkevmcd, line 16: Should this be a log rather than `fmt.Printf`?

pkg/plugins/autodiscovery/flux/ocirepository.go
```diff
@@ -0,0 +1,48 @@
...
+
+type ociRepositorySpecRef struct {
+	Tag string `yaml:"tag,omitempty"`
+}
+
+func loadOCIRepository(filename string) (*ociRepository, error) {
+	var ociRepository ociRepository
+	var data []byte
+
+	data, err := os.ReadFile(filename)
```
bigkevmcd, line 33: You declare `data` above and then redeclare it here?

You could move the declaration of the `ociRepository` to just above the place where you use it.

pkg/plugins/autodiscovery/flux/ocirepository.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("opening file %s: %s", filename, err.Error())
+	}
+
+	err = yaml.Unmarshal(data, &ociRepository)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling OCIRepository file %s: %s", filename, err.Error())
```
bigkevmcd, line 40: You can pass in `err` here without calling `.Error()` on it https://pkg.go.dev/fmt

> If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %x %X), or is %v but not %#v, the following two rules apply:
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

pkg/plugins/autodiscovery/flux/helmrepository.go
```diff
@@ -0,0 +1,66 @@
+package flux
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"gopkg.in/yaml.v3"
```
bigkevmcd, line 8: You could use `metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"` here and `"sigs.k8s.io/yaml"`

Which would let you use `metav1.TypeMeta` instead of extracting the `ApiVersion` and `Kind` here?

This would then let you do something like...

```go
gvk := helmRepository.GroupVersionKind()
if gvk.GroupKind().String() == "HelmRepository.source.toolkit.fluxcd.io" {
        return &helmRepository, nil
}
```
Which would tighten down the criteria for `HelmRepository` resources in the repo?

Bearing in mind that these are not coming from the cluster, so...there could be invalid resources in-repo that would not be applied to the cluster.

pkg/plugins/autodiscovery/flux/main.go
```diff
@@ -0,0 +1,168 @@
...
+var (
+	// defaultFluxFiles specifies accepted Helm chart metadata file name
+	defaultFluxFiles            []string = []string{"*.yaml", "*.yml"}
+	defaultVersionFilterPattern string   = "*"
+	defaultVersionFilterKind    string   = "semver"
+)
+
+// Spec defines the parameters which can be provided to the Flux crawler.
+type Spec struct {
+	/*
```
bigkevmcd, line 21: I'm curious about why use multi-line comments like this here?

This is normally written as

```go
// Auths provides a map of registry credentials where the key is the registry URL without scheme
```
If it's split over multiple lines, just add `//` at the start of each one?

olblak: Multi-line comments was needed for https://github.com/invopop/jsonschema to correctly generate the jsonschema description otherwise it would just pick the last `//` before the struct field

pkg/plugins/autodiscovery/flux/utils.go
olblak: Thanks that's a good catch

pkg/plugins/autodiscovery/flux/helmrepository.go
olblak: I admit I hesitated doing that for other plugins as well, my main motivation for not doing it was to reduce the number of Golang dependencies needed for Updatecli considering I could just create the struct I need.

olblak: Regarding `sigs.k8s.io/yaml` vs `gopkg.in/yaml.v3` if I recall correctly the main motivation of `sigs.k8s.io/yaml` is being able to use the same struct tag for both json and yaml files because otherwise sigs.k8s.io is based on `gopkg.in/yaml.v3` and so share the same limitation

bigkevmcd: One benefit of the sigs.k8s.io approach is that it's parsing it as it would be parsed by K8s.

olblak: Thanks for the suggestion, I switched to `sigs.k8s.io/yaml` in futur, I may applied the same improvement to other Kubernetes related plugins

pkg/plugins/autodiscovery/flux/main.go
olblak: Thanks @bigkevmcd for double checking that the syntax `/* */` wasn't required anymore.
I switched to `//` which I also find easier to read

pkg/plugins/autodiscovery/flux/ocirepository.go
olblak: fixed

olblak:
Thanks @bigkevmcd for the reviews. It think it's should be fine to start with the current implementation

PR #1965, opened 2024-02-29
fix: remove useless printf

olblak:
fix: remove useless printf

## Test

```shell
cd <to_package_directory>
go test
```


olblak:
Merging as only codeql is failing due to https://github.com/github/codeql/issues/15647

PR #2046, opened 2024-04-24
feat: Add support for GitHub API commit

### LINKED ISSUE #1914: Feature Request: Automated signed commits for GitHub Actions if a github-bot

Signed commits can be done within the existing environment or by the CI if a bot exists.

### Solution you'd like

If running for GitHub actions, then I'd like to enable new settings to do the magic 

https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c is an example of using the existing GITHUB_TOKEN 

```yaml
actions:
  pull-request:
    kind: "github"
    scmid: default
    spec:
      autogithubbotsign: true
```

The if `autogithubbotsign` and `CI=true` and GitHub actions, then use signing with the Rest API

Configure the local environment with the GPG as explained in https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

### Anything else?

_No response_

MattiasAng:
Fix #1914 

This pull request adds support for GitHub API commit. It allows Github SCM to commit a file in a GitHub repository using the GitHub API instead of using native Git.

The GitHub commit API is useful when a user is using the [automated token from Github Actions](https://docs.github.com/en/enterprise-cloud@latest/actions/security-guides/automatic-token-authentication) as it creates a signed commit automatically, removing the requirement of using a bot account with a GPG key configured.

For now I've hidden this feature behind the `commitUsingApi` (not loving the parameter name, happy for suggestions).

In addition I changed the Github pull request action to check if working branch is ahead using the API instead of relying on the native Git handler as I kept running in to issues where Github Actions wasn't recognizing changes done (not exactly sure why, but I think this will 100% ensure that the commit is available in Github).

As can be seen in the below screenshot, the commit is shown as verified with the Github Actions bot user committing the changes.
image

## Test

Build my fork and use the Github SCM with `commitUsingApi: true`.

### DIFF
--- a/pkg/plugins/scms/github/main.go
+++ b/pkg/plugins/scms/github/main.go
@@ -150,6 +150,13 @@ type Spec struct {
 	//
 	//  default: true
 	WorkingBranch *bool `yaml:",omitempty"`
+	//  "commitUsingApi" defines if Updatecli should use Github GraphQL API to create the commit.
+	//
+	//  compatible:
+	//	  * scm
+	//
+	//  default: false
+	CommitUsingAPI *bool `yaml:",omitempty"`
 }
 
 // GitHub contains settings to interact with GitHub
@@ -162,6 +169,7 @@ type Github struct {
 	nativeGitHandler gitgeneric.GitHandler
 	mu               sync.RWMutex
 	workingBranch    bool
+	commitUsingApi   bool
 }
 
 // Repository contains GitHub repository data
@@ -172,6 +180,7 @@ type Repository struct {
 	ParentID    string
 	ParentName  string
 	ParentOwner string
+	Status      string
 }
 
 // New returns a new valid GitHub object.
@@ -217,6 +226,11 @@ func New(s Spec, pipelineID string) (*Github, error) {
 		force = *s.Force
 	}
 
+	commitUsingApi := false
+	if s.CommitUsingAPI != nil {
+		commitUsingApi = *s.CommitUsingAPI
+	}
+
 	if force {
 		if !workingBranch && s.Force == nil {
 			errorMsg := fmt.Sprintf(`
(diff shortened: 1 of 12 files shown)


MattiasAng:
@olblak Leaving this in draft for now as I am still doing some testing in our organization, but happy to get any feedback. I'll see if I can get some tests in as well when I get some more time.

pkg/plugins/scms/github/main.go
```diff
@@ -150,6 +150,13 @@ type Spec struct {
 	//
 	//  default: true
 	WorkingBranch *bool `yaml:",omitempty"`
+	//  "commitUsingApi" defines if Updatecli should use Github GraphQL API to create the commit.
```
olblak, line 153:
```suggestion
	//  "commitUsingApi" defines if Updatecli should use GitHub GraphQL API to create the commit.
```

pkg/plugins/scms/github/main.go
```diff
@@ -217,6 +226,11 @@ func New(s Spec, pipelineID string) (*Github, error) {
 		force = *s.Force
 	}
 
+	commitUsingApi := false
```
olblak, line 229: I like that you set false by default, once the feature has been proved to be working well in production, I would imagine switching to true by default in a future release

olblak:
Spelling issues can be ignored by adding them to https://github.com/updatecli/updatecli/blob/main/.github/actions/spelling/allow.txt

MattiasAng:
@olblak Added a pattern to remove spellcheck on lines where it didn't make sense which was a bit easier than having to add exclusions for base64 strings. 😄

olblak (approved):
Thanks for the pullrequest, it looks great.
I didn't test it yet, but I am planning to release a new release candidate 0.76.0-rc.2 so we can see how it behaves during a few days

Issue #2068, opened 2024-04-29
GitHub pullrequest description updated to a wrong content


olblak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

From time to time, Updatecli updates a pullrequest description to something totally unrelated and then can restore it back. 

![image](https://github.com/updatecli/updatecli/assets/2360224/7ed00e0e-f457-4533-a900-62eeb4b449c5)

### Expected Behavior

The pullrequest should always represents the content of the pullrequest.

### Environment

```markdown
- OS:linux
- updatecli: v0.76.0
```

PR #2070, opened 2024-04-29
fix: use API for creating branch and fetching hashes when committing using Github API

MattiasAng:
Related to #1914 and issues described in [this comment](https://github.com/updatecli/updatecli/issues/1914#issuecomment-2080851598).

Note: I will probably not have time to add any tests this week, so feel free to take over @olblak or if approve if you feel it's good enough since most are API actions that can't be properly tested without integration tests.

The previous implementation used native Git to both create the working branch and fetching latest commit hash used when committing using API. This caused issues when using multiple targets with the same SCM due to the local cloned repository not having pulled the latest changes from the remote branch, which meant that `PushBranch` reverted changes done in the first target that ran.

To fix this, I've added the following: 

- Latest commit hashes are fetched from Github API instead of the native Git handler.
- Branch is now created using the Github API if an existing branch is not found.

## Test

- Build forked repository and branch and run with a [manifest that contains one SCM and multiple targets that uses the SCM](https://github.com/olblak/kubernetes-marketplace/blob/master/.github/updatecli/updatecli.d/kubewarden.yaml).

### Potential improvement

The `queryHeadOid` function use much of the same code as `queryRepository` but this function fails if the working branch is not available due to it being used as an input. We could probably rework this to be a more dynamic struct, but I think having an explicit function that only fetches the required information (i.e. hashes) makes sense even if it's not following DRY.

### DIFF
--- a/pkg/plugins/scms/github/main.go
+++ b/pkg/plugins/scms/github/main.go
@@ -550,3 +550,18 @@ type refQuery struct {
 		}
 	} `graphql:"createRef(input:$input)"`
 }
+
+func (g *Github) createBranch(branchName string, repositoryId string, headOid string) error {
+	var query refQuery
+
+	input := githubv4.CreateRefInput{
+		RepositoryID: repositoryId,
+		Name:         githubv4.String(fmt.Sprintf("refs/heads/%s", branchName)),
+		Oid:          githubv4.GitObjectID(headOid),
+	}
+
+	if err := g.client.Mutate(context.Background(), &query, input, nil); err != nil {
+		return err
+	}
+	return nil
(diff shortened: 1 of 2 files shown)


olblak:
Thanks for the quick pullrequest, as you said, integration testing is not easy in that area of the code.
I'll run some manual test but since this feature is still in "experimental" mode, I'll probably rush the test

olblak:
I am going to merge as I need to do some testing in that area anyway because of this issue
https://github.com/updatecli/updatecli/issues/2068

olblak:
Thanks for the quick pullrequest

Issue #1879, opened 2024-01-18
Feature Request: Renaming Updatecli commands


olblak:
### Is your feature request related to a problem?

while looking at the cmd package, I realize how none consistent the command and I wonder if

### Solution you'd like

`updatecli diff` should become `updatecli pipeline diff`
`updatecli apply` should become `updatecli pipeline apply`

### Alternatives you've considered

`updatecli diff` should become `updatecli policy diff`
`updatecli apply` should become `updatecli policy apply`

### Anything else?

We can still keep `updatecli diff` and `updatecli apply` for backward compatibility

dduportal:
This is a really good idea, ala. "docker"

olblak:
For some reason my previous message was lost 🤷🏾 

I realize there is another perspective to consider. 
A while ago I deprecated `updatecli show` in favor of `updatecli manifest show` with the evolution of the project, I now find this decision weird... 
I think I should have used `pipeline` or `policy` instead of `manifest` with a preference over `policy`. 

Today `updatecli manifest` accepts the following sub commands: 

```
 ~: updatecli manifest --help                                                                                                                                                                                                            
manifest executes specific manifest task such as upgrade

Usage:
  updatecli manifest [command]

Available Commands:
  pull        pull manifest(s) from an OCI registry
  push        push manifest(s) to an OCI registry
  show        show manifest(s) which will be executed
  upgrade     upgrade executes manifest upgrade task

Flags:
  -h, --help   help for manifest

Global Flags:
      --debug          Debug Output
      --experimental   Enable Experimental mode
```

I find it weird to add `apply` and `diff` under `manifest`

In mid/long term, I would prefer using `updatecli pipeline` when it comes to executing Updatecli pipeline and renaming `updatecli manifest` to `updatecli policy` when it comes to policy manipulation

Any thoughts?

olblak:
I think this pullrequest is kind of related 
https://github.com/updatecli/updatecli/pull/1884

I am creating a new subcommand under `manifest` such as ` updatecli manifest init` but I prefer something like `updatecli policy init`

PR #2288, opened 2024-07-08
feat(compose lint): support lint compose policies

v1v:
Fix #XXX

Replace shell scripting in https://github.com/updatecli/policies/blob/main/.ci/scripts/release.bash#L41-L122 with a CLI command, for such I decided to use `updatecli compose lint <path>`.

I have yet to implement everything, just a tiny validation. Please let me know what your thoughts are.

## Test

```shell
cd <to_package_directory>
go test
```

### DIFF
--- a/cmd/compose.go
+++ b/cmd/compose.go
@@ -18,4 +18,6 @@ var (
 		Use:   "compose",
 		Short: "compose executes specific Updatecli compose tasks such as diff or apply",
 	}
+	// policyFolder represents the policy folder
+	policyFolder string
 )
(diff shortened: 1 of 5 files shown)


olblak:
Thanks for the pullrequest, having this checks within the Updatecli command make sense.
I am just wondering the name `updatecli compose lint` because it's not related to the compose feature but more policy. 

I don't have a better solution at the moment, for info I am considering renaming a few command cfr https://github.com/updatecli/updatecli/issues/1879

PR #2316, opened 2024-07-16
feat: Allow the dockerimage plugin to retrieve changelog information

### LINKED ISSUE #2286: Feature Request: autodiscovery for updatecli policies to support changelog

It's not possible to know what's the changelog for the updatecli policies when using the autodiscovery

### Solution you'd like

I want to see https://github.com/updatecli/udash/pull/93, which contains the changelog 

_No response_

### Anything else?

_No response_

olblak:
Fix #2286 

While doing some testing for this pullrequest, I spotted https://github.com/updatecli/updatecli/pull/2316

If an OCI manifest like a docker image or an updatecli policy has an annotation `org.opencontainers.image.changelog` , is configured with a URL pointing to a markdown file such as `https://github.com/updatecli/policies/blob/main/updatecli/policies/golang/version/CHANGELOG.md`, then Updatecli tries to parse that markdown and retrieve an information.

For this work, it's important for the markdown file to have "release version" at the second level title such as 

CHANGELOG.md
```
# Changelog

## 0.2.0

Refactor the authentication mechanism.

## 0.1.0

Init the project
```

## Test

```shell
cd pkg/plugins/resources/dockerimage
go test
cd pkg/plugins/changelog/markdown 
go test
```

### Tradeoff

It appears that to retrieve the oci descriptor we need to download the oci artifact locally

I hesitated between adding the changelog URL to the manifest or the changelog content but after some discussion, it appeared that the URL would be more lightweight from an OCI registry point of view

I decided to use the annotation `org.opencontainers.image.changelog` which is not a "standard" OCI label (but I wish it was)

### Potential improvement

* We could add asciidoctor parser
* We could add a parser for [KeepAChangelog](https://keepachangelog.com/) syntax
* It could be better to provide more settings to how retrieving information from markdown like all versions matching semver pattern like `>3, <4`


olblak:
I am merging this pull request, for now it will applies to any OCI artifacts including docker image. This pullrequest can have a significant impact on performance as in order to read OCI annotations, we need to download the OCI artifacts first.

PR #2567, opened 2024-08-26
feat: implement prow autodiscovery

loispostula:
Implement Prow autodiscovery inside of kubernetes autodiscovery

## Test

```shell
cd pkg/plugins/autodiscovery/kubernetes
go test
```


olblak:
@loispostula I only have limited knowledge of Prow, as far as I know it was built to be used by the Kubernetes project.
I am puzzled to decide if Prow should be in the kubernetes plugin or next to it.
My understanding is that Prow is a different project and therefor should have its own plugin.

```
type kubernetesManifestSpec struct {
	ApiVersion string          `yaml:"apiVersion,omitempty"`
	Kind       string          `yaml:"kind,omitempty"`
	Spec       podTemplateSpec `yaml:"spec,omitempty"`
	// [ProwJobs](https://docs.prow.k8s.io/docs/jobs/) follows the Kubernetes spec
	ProwPreSubmitJobs  map[string][]podTemplate2Spec `yaml:"presubmits,omitempty"`
	ProwPostSubmitJobs map[string][]podTemplate2Spec `yaml:"postsubmits,omitempty"`
	ProwPeriodicJobs   []podTemplate2Spec            `yaml:"periodics,omitempty"`
}
```

Can we have "Prow" manifest with podtemplatespec?


What would be the most intuitive way to use this plugin

```
autodiscovery:
  crawlers:
    kubernetes:
```


```
autodiscovery:
  crawlers:
    kubernetes/prow:
```

```
autodiscovery:
  crawlers:
    prow:
```

Looking at https://docs.prow.k8s.io/
The latest one seems more intuitive to me

loispostula:
I agree with you, as stated over on element, the reason, I've gone this route was to prevent duplicating the kubernetes code.

I guess i could move the common code in `pkg/utils` and reuse it

olblak:
@loispostula I split the big file into smaller to improve visibility
What would you think to rename the package `kubernetes` in favor of `container` ? 
I think it would better represent its content

loispostula:
@olblak You mean the `utils` one?

olblak (approved):
> @olblak You mean the utils one?

I think I mainly annoyed that the prow plugin is reusing the Kubernetes plugin even thought it makes sense codewise. 

I was considering renaming the kubernetes package to "container" but I think it's neat picking without adding real value because containers are not specific to Docker or OCI container...

I am going to merge this pull request, and we can always rename or refactring this pluigin later if needed

PR #2591, opened 2024-08-30
feat(dockerfile): source implementation

### LINKED ISSUE #2022: Feature Request: Support Dockerfile FROM with argument

Today the dockerfile plugin doesn't support updating dockerfile target such as: 

```
FROM --platform=${BUILDPLATFORM} debian:stable-slim
```

### Solution you'd like

I would like the mobyparser to work with argument
https://github.com/updatecli/updatecli/blob/main/pkg/plugins/resources/dockerfile/mobyparser/main.go

The workaround is to rely on the instruction ENV or ARG such as 

```
ARG DEBIAN_VERSION=stable-slim
FROM --platform=${BUILDPLATFORM} debian:${DEBIAN_VERSION}
```

updatecli apply ghcr.io/updatecli/policies/autodiscovery/dockerfile

```
ARG DEBIAN_TAG=stable-slim@sha256:d10f0545d14bad5f4d230301f7e4fd904384f2dd16fda16d708f936c2fa1db3e
--FROM --platform=${BUILDPLATFORM} debian:${DEBIAN_TAG}
```

### Anything else?

_No response_

### LINKED ISSUE #2275: Feature Request: add source for `dockerfile` resource

I need to retrieve the value of a `FROM` instruction in a `Dockerfile` to use its value in my manifest, but the resource kind `dockerfile` does not have source support.

My manifest is there: https://github.com/jenkins-infra/docker-packaging/blob/048f979f954e8851db6f6c8abb0d6532ccc64186/updatecli/updatecli.d/jdk-build.yml#L38-L47

It tries to read the following instruction: https://github.com/jenkins-infra/docker-packaging/blob/048f979f954e8851db6f6c8abb0d6532ccc64186/Dockerfile#L6

### Solution you'd like

With the following `Dockerfile`:

```
FROM eclipse-temurin:${JAVA_VERSION}-jdk-jammy AS jdk
FROM --platform=linux/amd64 eclipse-mosquitto:2.0
```

I would expect:

- Snippet below to return `eclipse-temurin:${JAVA_VERSION}-jdk-jammy`:

```
sources:
  getImageName:
    name: Get the Image name
    kind: dockerfile
    spec:
      file: Dockerfile      
      instruction:
        keyword: "FROM"
        matcher: "eclipse-temurin"
```

- Snippet below to return `eclipse-mosquitto:2.0`:

```
sources:
  getImageName:
    name: Get the Image name
    kind: dockerfile
    spec:
      file: Dockerfile      
      instruction:
        keyword: "FROM"
        matcher: "mosquitto"
```

- Snippet below to return (multiline string)

```
eclipse-temurin:${JAVA_VERSION}-jdk-jammy
eclipse-mosquitto:2.0`:
```

```
sources:
  getImageName:
    name: Get the Image name
    kind: dockerfile
    spec:
      file: Dockerfile      
      instruction:
        keyword: "FROM"
        matcher: "eclipse-"
```

For now, we are using a `file` resource with a `matchpattern`: it works but it is tedious.

### Anything else?

_No response_

loispostula:
Fix #2022
Fix #2275 

`dockerfile` improvements

## Test

```shell
cd pkgs/plugins/dockerfile
go test
```

## Additional Information

This PR brings 4 changes to the dockerfile resources:
- Addition of the `stage` param for the Spec:
The default value (omit/empty) does not change the behaviour of the `condition` and `target`, all stages are considered for condition/targetting. But if a stage is specified, only that stage will be taken into account
- #2275 Creation of the `source`: `dockerfile now support a source attribute. Here the `stage` works a bit differently, if none is specified, then only the last stage of the dockerfile will be taken into account, otherwise we use the provided stage name. In the vast mojority of cases (non multi-stage dockerfile) this means that the argument can be safely disregarderd. 
- #2022 Refactoring of the `SimpleTextParser` to accomodate for the whole spec of the `FROM` keyword.

### Tradeoff
- The changes where only implemented to the `SimpleTextParser`, this means that the `moby` parser remains unchanged, no `source` and no `stage`
- As a docker image ref can be specified as `<image>[:<tag>][@<digest>]` we can have a situation where a dockerfile is defined with a ref such as `alpine@sha256...`. If a user uses the `dockertag` resource to update it, the image ref would become `alpine:3.1.4`. Maybe we should add a warning that the user is switching from digest specification of the ref to tag only.

### DIFF
--- a/pkg/plugins/resources/dockerfile/main.go
+++ b/pkg/plugins/resources/dockerfile/main.go
@@ -21,7 +21,11 @@ type Spec struct {
 	Instruction types.Instruction `yaml:"instruction,omitempty"`
 	// Value specifies the value for a specified Dockerfile instruction.
 	Value string `yaml:"value,omitempty"`
-	// Stage specifies the stage to extract the value from, if None specified, the latest stage will be used
+	// Stage can be used to further refined the scope
+	// For Sources:
+	// - If not defined, the last stage will be considered
+	// For Condition and Targets:
+	// - If not defined, all stages will be considered
 	Stage string `yaml:"value,omitempty"`
 }
 
--- a/pkg/plugins/resources/dockerfile/main.go
+++ b/pkg/plugins/resources/dockerfile/main.go
@@ -21,6 +21,9 @@ type Spec struct {
 	Instruction types.Instruction `yaml:"instruction,omitempty"`
 	// Value specifies the value for a specified Dockerfile instruction.
 	Value string `yaml:"value,omitempty"`
+	// Stage specifies the stage to extract the value from, if None specified, the latest stage will be used
+	// only for sources
+	Stage string `yaml:"value,omitempty"`
 }
 
 // Dockerfile defines a resource of kind "dockerfile"
--- a/pkg/plugins/resources/dockerfile/main.go
+++ b/pkg/plugins/resources/dockerfile/main.go
@@ -22,7 +22,6 @@ type Spec struct {
 	// Value specifies the value for a specified Dockerfile instruction.
 	Value string `yaml:"value,omitempty"`
 	// Stage specifies the stage to extract the value from, if None specified, the latest stage will be used
-	// only for sources
 	Stage string `yaml:"value,omitempty"`
 }
 
--- a/pkg/plugins/resources/dockerfile/mobyparser/main.go
+++ b/pkg/plugins/resources/dockerfile/mobyparser/main.go
@@ -27,7 +27,7 @@ var (
 	positionKeyRegex = regexp.MustCompile(`^(.*)\[[[:digit:]]*\]\[[[:digit:]]*\]$`)
 )
 
-func (m MobyParser) FindInstruction(dockerfileContent []byte) bool {
+func (m MobyParser) FindInstruction(dockerfileContent []byte, stage string) bool {
 
 	data, err := parser.Parse(bytes.NewReader(dockerfileContent))
 	if err != nil {
@@ -62,7 +62,7 @@ func (m MobyParser) FindInstruction(dockerfileContent []byte) bool {
 	return true
 }
 
-func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue string) ([]byte, types.ChangedLines, error) {
+func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue, stage string) ([]byte, types.ChangedLines, error) {
 
 	changed := make(types.ChangedLines)
 
@@ -118,10 +118,9 @@ func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue st
 	return []byte(newDockerfileContent), changed, nil
 }
 
-func (m MobyParser) GetInstruction(dockerfileContent []byte) []types.StageInstructionValue {
-	var stagesInstruction []types.StageInstructionValue
+func (m MobyParser) GetInstruction(dockerfileContent []byte, stage string) string {
 	// Not implemented
-	return stagesInstruction
+	return ""
 }
 
 func (m MobyParser) String() string {
--- a/pkg/plugins/resources/dockerfile/mobyparser/main.go
+++ b/pkg/plugins/resources/dockerfile/mobyparser/main.go
@@ -120,6 +120,7 @@ func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue, s
 
 func (m MobyParser) GetInstruction(dockerfileContent []byte, stage string) string {
 	// Not implemented
+	logrus.Warningf("Get Instruction is not yet supported for the MobyParser, if needed switch to the simpletextparser")
 	return ""
 }
 
--- a/pkg/plugins/resources/dockerfile/mobyparser/main.go
+++ b/pkg/plugins/resources/dockerfile/mobyparser/main.go
@@ -120,6 +120,7 @@ func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue st
 
 func (m MobyParser) GetInstruction(dockerfileContent []byte) []types.StageInstructionValue {
 	var stagesInstruction []types.StageInstructionValue
+	// Not implemented
 	return stagesInstruction
 }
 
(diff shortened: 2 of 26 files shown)


dduportal:
> That being said, alias should be unique and the last stage should never have one. 

I would not assume such a behavior: it makes sense but this rule is not written in the `Dockerfile` reference documentation (https://docs.docker.com/reference/dockerfile/#from) neither in https://docs.docker.com/build/building/multi-stage/ => assuming people won't name there last stage is a bit of a shortpath.

I'm only raising this as a nitpick concern: it is NOT challenging the PR content and choices of course :)

loispostula:
I stand corrected on the last stage not being able to be named, but stage do need to be unique: https://docs.docker.com/reference/build-checks/duplicate-stage-name/

How would the spec look like for the FROM-AS to say that the match should match the image and not the alias? Because for me, if you write a condition with a from as:
```yaml
conditions:
  builderPresent:
    kind: dockerfile
    spec:
      file: Dockerfile
      instruction:
        keyword: "FROM-AS"
        matcher: "builder"
```
It's about checking wether or not the alias is present. I'm not sure how we would represent, 'FROM-AS' but with image, maybe have another argument to the instructions, like `kind` that could any token from the from, so `platform|image|tag|digest|alias`?

dduportal:
> I stand corrected on the last stage not being able to be named, but stage do need to be unique: https://docs.docker.com/reference/build-checks/duplicate-stage-name/
> 
> How would the spec look like for the FROM-AS to say that the match should match the image and not the alias? Because for me, if you write a condition with a from as:
> 
> ```yaml
> conditions:
>   builderPresent:
>     kind: dockerfile
>     spec:
>       file: Dockerfile
>       instruction:
>         keyword: "FROM-AS"
>         matcher: "builder"
> ```
> 
> It's about checking wether or not the alias is present. I'm not sure how we would represent, 'FROM-AS' but with image, maybe have another argument to the instructions, like `kind` that could any token from the from, so `platform|image|tag|digest|alias`?

I agree with your proposal in the PR: the spec I proposed in the issue was mistaken. As you underline, the alias must be unique, as such what you say makes sense.

I'm wondering if my in initial issue is even valid. How would you feel about splitting the PR here, so we remove the `FROM-AS` logic and think about it in details (I have to reconsider my use case) and we could focus on the other code elements?

loispostula:
@dduportal agreed, move the from-as implem into https://github.com/updatecli/updatecli/pull/2600

pkg/plugins/resources/dockerfile/simpletextparser/keywords/label.go
```diff
@@ -0,0 +1,47 @@
...
+
+type Label struct{}
+
+func (a Label) ReplaceLine(source, originalLine, matcher string) string {
+
+	if !a.IsLineMatching(originalLine, matcher) {
+		return originalLine
+	}
+
+	// With an ARG instruction, we only need to use the 2nd "word"
```
olblak, line 16: > // With an LABEL instruction, we only need to use the 2nd "word"

pkg/plugins/resources/dockerfile/simpletextparser/keywords/label.go
```diff
@@ -0,0 +1,47 @@
...
+		return false
+	}
+	found := strings.ToLower(parsedLine[0]) == "label" && strings.HasPrefix(parsedLine[1], matcher)
+
+	return found
+}
+
+func (a Label) GetValue(originalLine, matcher string) (string, error) {
+	if a.IsLineMatching(originalLine, matcher) {
+		// With an ARG instruction, we just need the rest of the 2nd "word"
```
olblak, line 36:
```suggestion
		// With an LABEL instruction, we just need the rest of the 2nd "word"
```

pkg/plugins/resources/dockerfile/simpletextparser/keywords/label.go
```diff
@@ -0,0 +1,47 @@
...
+		parsedLine := strings.Fields(originalLine)
+		splitArgValue := strings.Split(parsedLine[1], "=")
+		if len(splitArgValue) < 2 {
+			// ARG without value
```
olblak, line 40:
```suggestion
			// LABEL without value
```

pkg/plugins/resources/dockerfile/simpletextparser/keywords/env.go
```diff
@@ -38,3 +38,23 @@ func (a Env) IsLineMatching(originalLine, matcher string) bool {
 
 	return found
 }
+
+func (a Env) GetValue(originalLine, matcher string) (string, error) {
+	if a.IsLineMatching(originalLine, matcher) {
+		// With an ARG instruction, we just need the rest of the 2nd "word"
```
olblak, line 44:
```suggestion
		// With an ENV instruction, we just need the rest of the 2nd "word"
```

pkg/plugins/resources/dockerfile/main.go
```diff
@@ -21,6 +21,9 @@ type Spec struct {
 	Instruction types.Instruction `yaml:"instruction,omitempty"`
 	// Value specifies the value for a specified Dockerfile instruction.
 	Value string `yaml:"value,omitempty"`
+	// Stage specifies the stage to extract the value from, if None specified, the latest stage will be used
+	// only for sources
```
olblak, line 25: @loispostula This is really awesome and I wonder if it would require a lot of effort to support condition/target as well?

pkg/plugins/resources/dockerfile/mobyparser/main.go
```diff
@@ -120,6 +120,7 @@ func (m MobyParser) ReplaceInstructions(dockerfileContent []byte, sourceValue st
 
 func (m MobyParser) GetInstruction(dockerfileContent []byte) []types.StageInstructionValue {
 	var stagesInstruction []types.StageInstructionValue
+	// Not implemented
```
olblak, line 122: Wouldn't it be better to show a warning message here?

pkg/plugins/resources/dockerfile/main.go
loispostula: @olblak I need to update the comment, it's been implemented for all

olblak, line 25:
```suggestion
```

olblak (dismissed):
Nice pull request, I really like all the various tests

olblak (approved):

PR #2593, opened 2024-08-31
chore(gitlab tests) more restrictive main regex

loispostula:
Gitlab source test may fail if the example repo has an updatecli worker branch in it

## Test

```shell
cd <to_package_directory>
go test
```


dduportal (approved):

dduportal:
Thanks @loispostula ! I changed the title to materialize the change (not a `fix` as it does not change anything to end users, and added a scope): is that ok for you?

PR #2600, opened 2024-08-31
feat(dockerfile): from-as keyword

### LINKED ISSUE #2276: Feature Request: with `dockerfile` resource, support `FROM <...> AS <...>` form

I would like  `updatecli` to manage the parent image alias, e.g. the `AS` secondary keyword  in the `FROM` instruction from dockerfile resource.

### Solution you'd like

With the following `Dockerfile`:

```
FROM eclipse-temurin:${JAVA_VERSION}-jdk-jammy AS jdk
FROM --platform=linux/amd64 eclipse-mosquitto:2.0
```

The following manifest should change the `Dockerfile` to 

```
FROM eclipse-temurin:${JAVA_VERSION}-jdk-jammy AS hello
FROM --platform=linux/amd64 eclipse-mosquitto:2.0 AS world
```

```
sources:
  sayHello:
    kind: shell
    spec:
      command: echo hello
  sayWorld:
    kind: shell
    spec:
      command: echo world

targets:
  setFromAliasToHello:
    sourceID: sayHello
    kind: dockerfile
    spec:
      file: Dockerfile
      instruction:
        keyword: "FROM-AS" # New keyword
        matcher: "eclipse-temurin"
  setFromAliasToWorld:
    sourceID: sayWorld
    kind: dockerfile
    spec:
      file: Dockerfile
      instruction:
        keyword: "FROM-AS" # New keyword
        matcher: "eclipse-mosquitto"
```

Using file matcher works but is tedious

### Anything else?

_No response_

loispostula:
Fix #2276 

- #2276 Accept new keyword `FROM-AS` to use the alias portion of an `FROM` line

## Test

```shell
cd pkgs/plugins/dockerfile
go test
```

## Additional Information

This PR requires #2591 

### Potential improvement

- Regarding #2276 the spec is a bit different, as we are in a `FROM-AS` context the matcher should is an alias. This means that the issue examples becomes

<details>
<summary>Dockerfile</summary>

```dockerfile
FROM eclipse-temurin:${JAVA_VERSION}-jdk-jammy AS jdk
FROM --platform=linux/amd64 eclipse-mosquitto:2.0 AS mqtt
```

</details>
<details>
<summary>manifest.yaml</summary>

```yaml
sources:
  sayHello:
    kind: shell
    spec:
      command: echo hello
  sayWorld:
    kind: shell
    spec:
      command: echo world

targets:
  setFromAliasToHello:
    sourceID: sayHello
    kind: dockerfile
    spec:
      file: e2e/updatecli.d/success.d/dockerfile/Dockerfile.from_as
      instruction:
        keyword: "FROM-AS"
        matcher: "jdk"
  setFromAliasToWorld:
    sourceID: sayWorld
    kind: dockerfile
    spec:
      file: e2e/updatecli.d/success.d/dockerfile/Dockerfile.from_as
      instruction:
        keyword: "FROM-AS"
        matcher: "mqtt"
```

</details>
Which seems ok for me, let me know if we should switch it. That being said, alias should be unique and the last stage should never have one. So In essence the expected Dockerfile in the issue is not valid. And the matching is still done using prefix, so if the end goal is to reflect some version in the alias, it can still be done.


olblak:
@loispostula  I am wondering if it wouldn't be better to use the parameter `stage` introduced in https://github.com/updatecli/updatecli/pull/2591 so we could just use "FROM"

olblak:
I am confused by this pullrequest as your other pullrequest also solve the current and IMHO in a better way.
I tried this manifest 

```
targets:
  setAppVersion:
    kind: dockerfile
    spec:
      file: e2e/updatecli.d/success.d/dockerfile/Dockerfile.source
#      stage: builder
      instruction:
        keyword: "FROM"
        #matcher: "UPDATECLI_VERSION"
        matcher: "alpine"
    sourceid: updatecli
```

both with and without a stage.

Without a stage 

```
TARGETS
========

setAppVersion
-------------

**Dry Run enabled**

⚠ The line #1, matched by the keyword "FROM" and the matcher "alpine", is changed from "FROM alpine:3.18 as builder" to "FROM alpine:v0.83.0 as builder".
⚠ The line #2, matched by the keyword "FROM" and the matcher "alpine", is changed from "FROM alpine:3.18 AS tester" to "FROM alpine:v0.83.0 AS tester".
⚠ The line #5, matched by the keyword "FROM" and the matcher "alpine", is changed from "FROM alpine:3.18" to "FROM alpine:v0.83.0".
⚠ - changed lines [1 2 5] of file "e2e/updatecli.d/success.d/dockerfile/Dockerfile.source"
```

It correctly update all my from instruction while with a stage it fix only one from instruction

loispostula:
I think it´s really the same, when using the `FROM` instruction, you are referencing the image in the FROM. So you are doing:
* Source: retrieve tag information for the image X
* Condition: Is the image present in the dockerfile
* Target: Update the image tag in the dockerfile
For it, the introduced `stage` parameter permits to further limit the scope.

When using the `FROM-AS` argument, you are referencing the alias in the FROM. So you are doing:
* Source: retrieve tag information for the aliased stage X
* Condition: Is the alias present in the dockerfile
* Target: Update the tag for the alias staged in the dockerfile
For it, the introduced `stage` parameter does not limit the scope further as they are referencing the same information.

So while i get where you are coming from, I don't think than a `FROM` instruction with a `stage` parameter defined is the same as a `FROM-AS` they are similar, but not referencing the same information.

One thing that I introduced in #2591 is the parsing of the `FROM` instructions into tokens, and if we check the dockerfile reference, we can see part of the spec as:
* __FROM__: `[--platform=<platform>] <image>[:<tag>][@<digest>] [AS <name>]`
* __LABEL__: `<key>=<value> ...`
* __ENV__: `<key>=<value> ...`
* __ARG__: `<name>[=<default value>]`
* __HEALTHCHECK__: `[--interval=<duration>] [--timeout=<duration>] [--start-period=<duration>] [--start-interval=<duration>] [--retries=<N>] CMD <command>`
* __RUN__: `--mount=[type=TYPE][,option=<value>[,option=<value>]...]`

So maybe we should not have a `FROM-AS` but have a `token` parameter that can be configured for each keyword. `FROM-AS` would become `FROM` with `token=alias` and it'd be more generic

loispostula:
@olblak @dduportal What do you think of my latest comment/proposition regarding the fromas?

PR #2601, opened 2024-09-01
feat(core): implement retry client for http fetching

loispostula:
Add exponiental retry mechanism for fetching http resource

## Test

```shell
cd pkg/core/check
go test
```

### Potential improvement

Use a library like https://github.com/cenkalti/backoff instead of the implemented RetryClient


olblak (approved):
Wow nice

Issue #2718, opened 2024-09-12
Feature Request: Descriptions of release notes between current and latest versions


rromic:
### Is your feature request related to a problem?

I would like to get release notes between current version and latest version so I can easily identify if there are some breaking changes between some version. Currently what I could get is only release notes of the latest version which could lead to some issues if I upgrade to that version and there was a breaking change that I haven't anticipated.

I would appreciate any other suggestion how to setup configuration to achieve same thing if possible!

Awesome tool!

olblak:
Thanks for your warm message and the feature request, I agree it's something I would like to see improved as well.
As part of https://github.com/updatecli/updatecli/issues/159 that something I would like to do

olblak:
I am switching this issue to an epic as all existing plugins need to be adapted.
My current focus is on plugin that already implement the Changelog interface.

Any other plugin need to be handled in a separated issue

PR #2800, opened 2024-10-07
feat(core/pipeline): allow dependency across source/conditions/targets

loispostula:
# Overview

This pull request introduces support for dependencies across sources, conditions, and targets, making the dependency model more flexible.
Current Behavior:

- Sources can depends on:
  - Other sources
- Conditions can depends on:
  - Other conditions
  - One source (by sourceID)
- Targets can depend on:
  - Other targets
  - Either:
    - All conditions
    - A subset of conditions
  - One source (by sourceID)

# New Behavior:

With this change, sources, conditions, and targets can depend on any other resource type. This is a stepping stone towards resolving issues like [issue #2718](https://github.com/updatecli/updatecli/issues/2718) and [issue #159](https://github.com/updatecli/updatecli/issues/159).

# Implementation

To achieve this, I've extended the current dependsOn specification. It can now be defined using the format:

```
(resourceType#)resourceId(:booleanOperator)
```

This change allows merging the three resource-specific dependency Directed Acyclic Graphs (DAGs) into a single unified DAG. The DAG is built using a custom node type:

```golang

type Dependency struct {
    ID       string
    Operator string
}

type Node struct {
    ID                string
    Category          string
    DependsOn         []Dependency
    DependsOnChange   bool
    Result            string
    Changed           bool
}
```

To ensure backward compatibility, the existing dependencies (i.e., DependsOn) are preserved. In addition, the sourceID is passed as a dependency for conditions and targets, and conditionIDs are fed into the targets as dependencies in the DAG.

A traversal of the DAG using `DescendantsFlow` processes each resource in the correct order once its dependencies are completed.

# Expected Impact

This PR is expected to be a no-op, meaning that while the core pipeline logic has been updated, there should be no impact on the output or behavior of existing pipelines.

# `updatecli manifest show -graph`
In order to help debugging the dependencies tree, 

I' ve added an option to the `show` command to display the DAG in graphviz format, this allows to get interesting information on a pipeline:
- e2e/updatecli.d/success.d/temurin/sources.yaml:
e2e/updatecli.d/success.d/temurin/sources.yaml
- e2e/updatecli.d/success.d/chained.yaml:
e2e/updatecli.d/success.d/chained.yaml
- e2e/updatecli.d/success.d/http.yaml:
e2e/updatecli.d/success.d/http.yaml


olblak:
Thanks for the pullrequest, this one is really awesome but I need a bit more time to review it properly, I am adding commits as I am reviewing it

olblak:
I am having many warning messages like 

`call of append copies lock value: github.com/updatecli/updatecli/pkg/core/pipeline.Pipeline contains sync.Mutex`

pkg/core/config/main.go
```diff
@@ -539,13 +539,13 @@ func (config *Config) validateTargets() error {
 			}
 		}
 
-		if t.DisableConditions && len(t.ConditionIDs) > 0 {
-			logrus.Errorf("target %q has 'disableconditions' set to true and 'conditionids' defined (%v), it's not possible to disable conditions and define conditions at the same time", id, t.ConditionIDs)
+		if t.DisableConditions && len(t.DeprecatedConditionIDs) > 0 {
```
olblak, line 542:
```suggestion
		if t.DisableConditions && len(t.DeprecatedConditionIDs) > 0 {
```

Why does it becomes deprecated?

olblak: Re-reading the code, I now understand 

```
Deprecated - please use DependsOn with `condition#conditionid` keys
```

loispostula: Exactly. In my opinion, the existence of `ConditionIds` is because, prior to this PR, a target by default depends on all of its conditions with an implicit `AND`. You could modify this behavior by enabling the `DisableConditions` option and adding a few IDs to `ConditionIDs`.

This behavior has now been generalized with the cross-resource DependsOn, so that:
```yaml
targets:
  conditionids:
     disableconditions: true
     conditionids:
        - 1
        - 3
  dependson:
     disableconditions: true
     dependson:
        - condition#1
        - condition#3
```
are equivalent. This is why I marked `ConditionIDs` as deprecated. From an implementation point of view, it's not critical—I've already implemented the logic for `ConditionIDs`, so we could mark them as non-deprecated. However, from a documentation and explanation point of view, I think it's easier to avoid having multiple ways to express the same behavior.

loispostula:
@olblak hiw did you get those warning? Could you share a pipeline or something so that I can investigate?

olblak:
> @olblak hiw did you get those warning? Could you share a pipeline or something so that I can investigate?

image

The problem is well explained on https://stackoverflow.com/questions/76648255/how-to-range-over-a-slice-of-structs-that-contain-mutexes-in-go

olblak:
I think this commit fixes the mutex issue 
https://github.com/updatecli/updatecli/tree/55b2fc247631de801646687c87c659d5fde90c8f

loispostula:
> Note for myself as it don't have to be done in this pullrequest but I am wondering if you can generated the graphviz graphs directly using https://github.com/goccy/go-graphviz

That's a good point. I think having a graph functionality could also be useful on generating some summary, with what ran and what didn't. But then maybe dot is not the ideal choice and we should use mermaid-js so that it's easy to but in github?

olblak:
> But this should be new issues / new pr. Here the goal is really to help debug and visualize dependencies

I think I would move the graph option behind the experimental flag so we can change it later.
Theh current pullrequrest is already big enough that I wouldn't add new content to it

pkg/core/config/main.go
olblak: Thanks for the clarification, it totally makes sense and you took the right approach

olblak (dismissed):

PR #2849, opened 2024-10-19
fix: show pipeline as skipped if all target are skipped

olblak:
This pullrequest fix a regression introduced in https://github.com/updatecli/updatecli/pull/2800
Where a pipeline would be displayed as succeeded if all targets are skipped.

## Test

```shell
cd <to_package_directory>
go test
```

### Potential improvement

We could apply the same logic if all conditions are skipped or all sources are skipped


loispostula (approved):

olblak:
Thanks, @loispostula, for adding some tests on this area, It seems like I broke them :D

loispostula:
Tests are always good, but maybe the tests are broken

olblak:
In this case tests were good :)

PR #2850, opened 2024-10-19
feat(gittag): Add git url param to gittag plugin

olblak:
Similarly to https://github.com/updatecli/updatecli/pull/2848

This pull request allows using the gitbranch plugin with a URL without having to specify a scmid
Using the scmid comes with the requirement that we need to know what branch the git repository use (master vs main) as Updatecli run git checkout operation before handling a resource, but in the context of the gitbranch all we need is running git clone to retrieve the branch locally.

The gitbranch plugin can now be used in three different ways

    using the scmid to clone the repository
    using the path parameter to analyze an already cloned git repository
    using the url parameter to clone and then analyze the repository, which is more convenient for autodiscovery plugin like 

https://github.com/updatecli/updatecli/pull/1964

## Test

```shell
cd <to_package_directory>
go test
```


olblak:
I now realize this plugin is missing a lot of tests...

Issue #2839, opened 2024-10-16
The Adoptium API has changed the location header


gounthar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the debug mode, I can see `DEBUG: [temurin] API client got the following Location header value: "".`
It makes me think of the [discussion](https://adoptium.slack.com/archives/C09NLQQAV/p1728223668491699?thread_ts=1728128855.456059&cid=C09NLQQAV) Mark Waite and the Temurin team got about the recent changes due to the CloudFlare usage.

```
CONDITIONS:
===========

checkTemurinAlpineRelease
-------------------------
DEBUG: [temurin] Performing an http GET request to "https://api.adoptium.net/v3/types/architectures"...
DEBUG: [temurin] API client returned the following response:
&{200 OK 200 HTTP/2.0 2 0 map[Alt-Svc:[h3=":443"; ma=86400] Cf-Cache-Status:[DYNAMIC] Cf-Ray:[8d365b2e7bd299d9-CDG] Content-Type:[application/json;charset=UTF-8] Date:[Wed, 16 Oct 2024 07:32:38 GMT] Server:[cloudflare] Strict-Transport-Security:[max-age=63072000; includeSubDomains; preload] X-Content-Type-Options:[nosniff] X-Frame-Options:[DENY] X-Pod-Hostname:[frontend-service-f6979b64b-zgv68]] 0xc00126c2a0 -1 [] false true map[] 0xc001034c80 0xc0011f20c0}
DEBUG: [temurin] API client got the following Location header value: "".
DEBUG: [temurin] Performing an http GET request to "https://api.adoptium.net/v3/types/operating_systems"...
DEBUG: [temurin] API client returned the following response:
&{200 OK 200 HTTP/2.0 2 0 map[Alt-Svc:[h3=":443"; ma=86400] Cf-Cache-Status:[DYNAMIC] Cf-Ray:[8d365b2f3c6d99d9-CDG] Content-Type:[application/json;charset=UTF-8] Date:[Wed, 16 Oct 2024 07:32:39 GMT] Server:[cloudflare] Strict-Transport-Security:[max-age=63072000; includeSubDomains; preload] X-Content-Type-Options:[nosniff] X-Frame-Options:[DENY] X-Pod-Hostname:[frontend-service-f6979b64b-zgv68]] 0xc0010c2fc0 -1 [] false true map[] 0xc001034f00 0xc0011f20c0}
[...]

PR #2852, opened 2024-10-21
fix: disabling temurin s390x test

olblak:
I am disabling this test until we find a solution. The root cause could be related to https://github.com/updatecli/updatecli/issues/2839


dduportal:
Note: related to https://github.com/updatecli/updatecli/issues/2914

PR #3054, opened 2024-11-15
Support maven mirror in autodiscovery

### LINKED ISSUE #3053: Feature Request: Support mirrors in maven autodiscovery

In many environments it is preferable to use an internal mirror of maven central.  Support autodiscovery against an internal mirror.

### Solution you'd like

It would be great if the mirror could be configured at the crawler level

```
autodiscovery:
  scmid: default
  crawlers:
    maven:
      mirror: https://my.internal.repo/central
```

I have considered adding a `--mirror` flag to `updatecli`, but this seems inappropriate for a single crawler.

### Anything else?

_No response_

AidanDelaney:
Allow the maven autodiscovery crawler to be configured to use a mirror instead of upstream maven central.

Fix #3053 

## Test

I have not yet provided a test.

### Tradeoff

The `maven` crawler now needs access to the `maven` plugin in order to set a default `MavenCentralRepository`.

### Potential improvement

Parse and understand the standard maven `settings.xml` file to identify any mirrors.

### DIFF
--- a/pkg/plugins/autodiscovery/maven/main.go
+++ b/pkg/plugins/autodiscovery/maven/main.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/plugins/resources/maven"
 	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
 )
 
@@ -42,6 +43,7 @@ type Spec struct {
 		and its type like regex, semver, or just latest.
 	*/
 	VersionFilter version.Filter `yaml:",omitempty"`
+	Mirror        string         `yaml:",omotempty"`
 }
 
 // Maven hold all information needed to generate helm manifest.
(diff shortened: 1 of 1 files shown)


pkg/plugins/autodiscovery/maven/main.go
```diff
@@ -42,6 +43,7 @@ type Spec struct {
 		and its type like regex, semver, or just latest.
 	*/
 	VersionFilter version.Filter `yaml:",omitempty"`
+	Mirror        string         `yaml:",omotempty"`
```
olblak, line 46:
```suggestion
	Mirror        string         `yaml:",omitempty"`
```

olblak:
Thanks for the pull request, I agree with you that ideally we should correctly handle settings.xml
Adding the parameter `mirror` sounds a good solution until we support the settings.xml

pkg/plugins/autodiscovery/maven/main.go
```diff
@@ -65,6 +67,10 @@ func New(spec interface{}, rootDir, scmID string) (Maven, error) {
 		return Maven{}, err
 	}
 
+	if s.Mirror != "" {
+		maven.MavenCentralRepository = s.Mirror
```
olblak, line 71: This will have unexpected side effects.
All Updatecli pipelines executed after this moment will use the mirror.

I think it's probably better to override the url in https://github.com/updatecli/updatecli/blob/9f9f34505083a8352c61a49e63206f981f550774/pkg/plugins/autodiscovery/maven/utils.go#L77

olblak:
Thanks @AidanDelaney for the pull request.
You motivated me enough to work on https://github.com/updatecli/updatecli/pull/3270 which I think cover more cases.
Feel free to comment to the other pullrequest if I am missing something

PR #3062, opened 2024-11-18
Add SkipPackaging option for helm autodiscovery

dshmelev:
SkipPackaging option allows users to bypass `helm dependency update` command during the helmchart target execution. 

For umbrella charts with many dependencies, the helm dependency update command is run repeatedly during dependency checks. This process is not always required and adds unnecessary overhead, especially in large projects.


olblak (approved):
Thanks for the pullrequest
It looks good to me, I didn't test it manually but I assume you did

PR #3277, opened 2024-12-12
Update SECURITY policy

olblak:


SECURITY.md
```diff
@@ -6,5 +6,5 @@ The Updatecli project only support the latest version.
 
 ## Reporting a Vulnerability
 
-Please disclose vulnerability via email to security@updatecli.io
+Please disclose vulnerability via [GitHub](https://github.com/updatecli/updatecli/security/advisories/new)
```
lemeurherveCB, line 9:
```suggestion
Please disclose vulnerability via [GitHub](https://github.com/updatecli/updatecli/security/advisories/new) or via email at security@updatecli.io
```
suggestion(optional): keep both in case some wants to avoid disclosing their GitHub username?

lemeurherveCB (approved):

SECURITY.md
dduportal: I would keep both with email by default. Relying only on a GitHub feature like this one, even if practical, looks dangerous to me as it creates a tight dependency top their business model.

dduportal (approved):

PR #3291, opened 2024-12-13
feat: allow to deepmerge multiple values file

### LINKED ISSUE #3284: Multiple values file merging behavior broken with same top level key

### Current Behavior

I am getting a broken template when I am providing two different values files which share the same key and multiple different child keys

### Expected Behavior

I would expect the same result of passing a single value file with keys merged.

### Steps To Reproduce

will provide a repo with code

### Environment

```markdown
- OS: linux/amd64
- updatecli: 0.89.0
```

### Pipeline Graph

```mermaid
```

### Anything else?

_No response_

gionn:
Fix #3284 

Replace the custom merging logic with mergo library

## Test

A quick test against https://github.com/gionn/updatecli-bug-3284 worked succesfully

More tests to come?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/updatecli/updatecli
 go 1.23.4
 
 require (
+	dario.cat/mergo v1.0.1
 	github.com/Masterminds/semver/v3 v3.3.1
 	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/aws/aws-sdk-go v1.55.5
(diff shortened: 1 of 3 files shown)


olblak:
Thanks, @gionn, for the pull request, I just did some manual testing and it works as expected.
I also open a pull request on https://github.com/gionn/updatecli/pull/1 to add unit test to catch any regression in the feature

olblak:
Once my pullrequest is merged on your fork, I think we are ready to move this pull requet to "ready"

olblak (approved):

PR #3347, opened 2024-12-20
feat: add version filter regex/semver

mcwarman:
I've had a couple of cases where i wanted to get latest semantic version, but the versions returned by the resource do not follow the convention.

The idea behind this to extract the version numbers using a regex then pass them on to the semver search.

[yarnpkg/berry](https://github.com/yarnpkg/berry) is a good example, as they do backport releases. Therefore latest of both `regex` and `latest` currently returns a `3.x`.

## Test

```shell
cd pkg/plugins/utils/version
go test
```

```shell
go build -o bin/updatecli
./bin/updatecli diff --config test.yaml
```

```yaml
sources:
  default:
    name: Get latest version
    kind: githubrelease
    spec:
      owner: yarnpkg
      repository: berry
      versionfilter:
        kind: regex/semver
        regex: "@yarnpkg/cli/(\\d*\\.\\d*\\.\\d*)"
        #pattern: "3.x"
```

### Tradeoff

Potentially confusing having `regex` parameter.

### Potential improvement

N/A

### DIFF
--- a/pkg/plugins/utils/version/filter.go
+++ b/pkg/plugins/utils/version/filter.go
@@ -17,16 +17,17 @@ const (
 	SEMVERVERSIONKIND string = "semver"
 	// LATESTVERSIONKIND specifies that we are looking for the latest version of an array
 	LATESTVERSIONKIND string = "latest"
+	// REGEXSEMVERVERSIONKIND use regex to extract versions and semantic version to find version
+	REGEXSEMVERVERSIONKIND string = "regex/semver"
 )
 
-var (
-	// SupportedKind holds a list of supported version kind
-	SupportedKind []string = []string{
-		REGEXVERSIONKIND,
-		SEMVERVERSIONKIND,
-		LATESTVERSIONKIND,
-	}
-)
+// SupportedKind holds a list of supported version kind
+var SupportedKind []string = []string{
+	REGEXVERSIONKIND,
+	SEMVERVERSIONKIND,
+	LATESTVERSIONKIND,
+	REGEXSEMVERVERSIONKIND,
+}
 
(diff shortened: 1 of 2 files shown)


pkg/plugins/utils/version/filter.go
```diff
@@ -17,16 +17,17 @@ const (
 	SEMVERVERSIONKIND string = "semver"
 	// LATESTVERSIONKIND specifies that we are looking for the latest version of an array
 	LATESTVERSIONKIND string = "latest"
+	// REGEXSEMVERVERSIONKIND use regex to extract versions and semantic version to find version
+	REGEXSEMVERVERSIONKIND string = "regex-semver"
```
mcwarman, line 21: I made this kebab case, this might not follow convention though.

olblak: To align with other plugin as you can see [here](https://github.com/updatecli/updatecli/blob/caa6cef9fe86cf172abe53a8e2c8484b2d81ed5c/pkg/core/pipeline/resource/main.go#L86), I would use "regex/semver"

olblak:
I like this pullr equest a lot, because I faced the same issues.
Like the kubernetes publish version with "kubernetes-" prefix

pkg/plugins/utils/version/filter.go
mcwarman: Perfect, it didn't feel right, and I didn't really engage my brain to resolve it.

mcwarman:
@olblak not sure on test failures seems unrelated.

olblak (approved):
Thanks for the pull request

PR #3371, opened 2024-12-27
feat: add versionfilter of kind lexicographical

acobaugh:
This change implements a `gosort` filter mode. This is intended to start a conversation around picking the "latest" version of a resource that uses the "calendar version" versioning method.

This seems really simple in the implementation (it just uses `slices.Sort()`), but should actually work for many of the calver use cases out there.

Thoughts?

## Test

```shell
cd <to_package_directory>
go test
```

### DIFF
--- a/pkg/plugins/utils/version/filter.go
+++ b/pkg/plugins/utils/version/filter.go
@@ -3,6 +3,7 @@ package version
 import (
 	"fmt"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -23,6 +24,8 @@ const (
 	REGEXTIMEVERSIONKIND string = "regex/time"
 	// TIMEVERSIONKIND represents versions use date format to identify the version
 	TIMEVERSIONKIND string = "time"
+	// LEXSORTVERSIONKIND uses the go sort package to find the latest version
+	LEXVERSIONKIND string = "lex"
 )
 
 // SupportedKind holds a list of supported version kind
@@ -33,6 +36,7 @@ var SupportedKind []string = []string{
 	REGEXSEMVERVERSIONKIND,
 	REGEXTIMEVERSIONKIND,
 	TIMEVERSIONKIND,
+	LEXVERSIONKIND,
 }
 
 // Filter defines parameters to apply different kind of version matching based on a list of versions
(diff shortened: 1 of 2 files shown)


olblak:
Thanks for the suggestion, it's definitely something I had in mind, but I never had the usecase.
I think a proper name for this kind of sorting would be "lexicographical", what do you think?

olblak:
I think a test case would be to check 

["v1.1", "v1.10", "v1.2"] where the latest version is "v1.2" and not "v1.10"

olblak:
The challenge about sorting date using this technique is that it's common for date to be written either YYYY-MM-DD or YYYY-DD-MM but it wouldn't work in the later format

olblak:
golangci-lint spotted a minor issue "pkg/plugins/utils/version/filter_test.go:179: File is not `gofmt`-ed with `-s` (gofmt)"

acobaugh:
> Thanks for the suggestion, it's definitely something I had in mind, but I never had the usecase. I think a proper name for this kind of sorting would be "lexicographical", what do you think?

I'm indifferent on the name.

> I think a test case would be to check
> 
> ["v1.1", "v1.10", "v1.2"] where the latest version is "v1.2" and not "v1.10"

Sure. Added.

> golangci-lint spotted a minor issue "pkg/plugins/utils/version/filter_test.go:179: File is not `gofmt`-ed with `-s` (gofmt)"

Fixed.


> The challenge about sorting date using this technique is that it's common for date to be written either YYYY-MM-DD or YYYY-DD-MM but it wouldn't work in the later format

True. Admittedly, this PR fixes our specific use case. I think if we wanted to implemented a true date sort, we would likely need the user to specify the pattern to use to extract the date. So this PR would work for the cases where the date format in use can sort naturally, otherwise we'd need something else.

olblak:
I am not a big fan of the name "gosort" as it doesn't really tell us what the sortting approach used. I feel like "lexicographical" would be more appropriated.

Also, you motivated me enough to add versionfilter based on time cfr https://github.com/updatecli/updatecli/pull/3405 feel free to review that pullrequest

acobaugh:
Okay, I renamed this to `lex`, since `lexicographical` is a lot to type out. Should be ready to review now.

olblak:
I just documented this change on https://github.com/updatecli/website/pull/2012

PR #3405, opened 2025-01-03
feat: add versionfilter of kind time

olblak:
- [x] https://github.com/updatecli/website/pull/2012

This pullrequest add two versionfilter kind:
* `time` that allow to retrieve the latest version sorted by time where the date format is defined by the parameter "pattern"
* `regex/time` that allow to retrieve the latest version sorted by time when the date must be extracted using a regular expression

Here are two examples:

```
sources:
  regextime:
    name: Get latest ubuntu version using regex/time versionfilter
    kind: dockerimage
    spec:
      image: ubuntu
      versionfilter:
        kind: 'regex/time'
        regex: '^focal-(\d*)$'
        pattern: "20060102"

  time:
    name: Get latest ubuntu version using time versionfilter
    kind: dockerimage
    spec:
      image: ubuntu
      versionfilter:
        kind: 'time'
        pattern: "06.01"

```

Where the first example sorts by date all tags matching `focal-YYYYMMDD` and the second example  sorts docker image tag by date matching `YY.MM`

It's worth mentioning that the date pattern must specified using the following conventions

```
Year: "2006" "06"
Month: "Jan" "January" "01" "1"
Day of the week: "Mon" "Monday"
Day of the month: "2" "_2" "02"
Day of the year: "__2" "002"
Hour: "15" "3" "03" (PM or AM)
Minute: "4" "04"
Second: "5" "05"
AM/PM mark: "PM"
```

## Test

```shell
go build -o bin/updatecli .
./bin/updatecli diff --config e2e/updatecli.d/success.d/versionfilter.yaml
cd pkg/plugins/utils/version/
go test .
```

### Tradeoff

I decided to use "time" instead of "date" to follow Golang convention.

### Potential improvement

It's worth mentioning that the version filter of kind "regex/time" returns the initial version and not just the date extracted from the version. I think it would make sense to adapt the version filter of kind "regex/semver" to follow the same pattern.


pkg/plugins/utils/version/filter.go
```diff
@@ -55,6 +61,8 @@ func (f Filter) Init() (Filter, error) {
 		f.Pattern = "*"
 	} else if f.Kind == REGEXVERSIONKIND && len(f.Pattern) == 0 {
 		f.Pattern = ".*"
+	} else if f.Kind == TIMEVERSIONKIND && len(f.Pattern) == 0 {
+		f.Pattern = "2006-01-02"
```
mcwarman, line 65: Looks good, was wondering is there a reason this date was chosen as default?

Seems random to me, but I might be missing the context.

mcwarman (dismissed):
Looks good to me, reviewing on phone so can't test it though.

pkg/plugins/utils/version/filter.go
mcwarman: Actually this is a pattern and date itself irrelevant?

mcwarman: If that's the case should this default also be applicable for regex time variant?

olblak: Good catch, I was setting a default pattern for "time" but not "regex/time"
Also, I agree with you that 2006-01-02 is an arbitrarily format which correspond to YYYY-MM-DD
I don't really have strong opinion across the different format like YYYY-MM-DD, YYYY-DD-MM, etc.

mcwarman: `YYYY-MM-DD` always feels correct to be region agnostic.

PR #3508, opened 2025-01-08
githubaction autodiscovery should allow for pinning version to commit hash 

### LINKED ISSUE #3507: Feature Request: githubaction autodiscovery should allow for pinning version to commit hash

Currently our company policies requires that github actions should be pinned by release commit hash.

e.g.:

> The commit hash for https://github.com/actions/setup-node/releases/v4.1.0 is 39370e3970a6d050c480ffad4ff0ed4d3fdee5af. When pinning, the Action's definition would be: - uses: `actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af`.

### Solution you'd like

The githubaction autodiscovery plugin should allow users to specify how to resolve an action version (branch, tag, release, commit)

- [x] `githubrelease` should allow output in hash or name (`key`)
- [x] `gitbranch` should allow output in hash or name (`key`)
- [x] `githubaction` autodiscovery should accept a `digest` parameter

_No response_

### Anything else?

_No response_

loispostula:
Fix #3507 
Introduce a `key` argument for github release to get the tag hash instead of the tag name.

## Test

```shell
cd <to_package_directory>
go test
```

## Additional Information

This pr introduces the following changes:
- `githubrelease` now supports an `key` argument like `gittag` that can either be `name` (default) or `hash`.
- `gitbranch` now supports an `key` argument like `gittag` that can either be `name` (default) or `hash`.
- `yaml` now accept a `comment` arg, when used in a `target` and with the `yamlpath` engine, it can be used to add a comment to a line.
  <details>
    <summary>Yaml comment</summary>
  
  ## Given the following `test.yaml`
  
  ```yaml
  ---
  owner: anakin
  ```
  ## Running `updatecli.yaml`
  
  ```yaml
  ---
  sources:
    default:
      kind: shell
      spec:
        command: echo "obiwan"
    comment:
      kind: shell
      spec:
        command: echo "is the best"
  targets:
    default:
      kind: yaml
      sourceid: default
      spec:
        engine: yamlpath
        key: owner
        file: test.yaml
        comment: '{{ source "comment" }}'
  ``` 
  ## Would yield `test.yaml`
  
  ```yaml
  ---
  owner: obiwan # is the best
  ```
  </details>

- `githubaction` autodiscovery can be setup to resolved `digest`:
```yaml
jobs:
  build:
    steps:
      - uses: "actions/checkout@v4`
```
gets resolved to
```yaml
jobs:
  build:
    steps:
      - uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"  # pinned from v4.2.2 by updatecli (do-not-remove-comment)`
```

The comment is required to be able to infer the update channel in the following updates, removing it would translate to a pinned version.

### DIFF
--- a/pkg/plugins/autodiscovery/githubaction/main.go
+++ b/pkg/plugins/autodiscovery/githubaction/main.go
@@ -90,6 +90,8 @@ type Spec struct {
 	//             token: '{{ requiredEnv "GITHUB_TOKEN" }}'
 	// ```
 	Credentials map[string]gitProviderToken `yaml:",omitempty"`
+	// Digest provides parameters to specify if the generated manifest should use a digest instead of the branch or tag.
+	Digest *bool `yaml:",omitempty"`
 }
 
 // GitHubAction holds all information needed to generate GitHubAction manifest.
@@ -108,6 +110,8 @@ type GitHubAction struct {
 	versionFilter version.Filter
 	// workflowFiles is a list of HelmRelease files found
 	workflowFiles []string
+	// digest holds the value of the digest parameter
+	digest bool
 }
 
 type gitProviderToken struct {
@@ -163,6 +167,10 @@ func New(spec interface{}, rootDir, scmID string) (GitHubAction, error) {
 		newFilter.Kind = defaultVersionFilterKind
 		newFilter.Pattern = defaultVersionFilterPattern
 	}
+	digest := false
+	if s.Digest != nil {
+		digest = *s.Digest
+	}
 
 	return GitHubAction{
 		credentials:   s.Credentials,
@@ -171,6 +179,7 @@ func New(spec interface{}, rootDir, scmID string) (GitHubAction, error) {
 		rootDir:       dir,
 		scmID:         scmID,
 		versionFilter: newFilter,
+		digest:        digest,
 	}, nil
 
 }
--- a/pkg/plugins/autodiscovery/githubaction/main_test.go
+++ b/pkg/plugins/autodiscovery/githubaction/main_test.go
@@ -13,6 +13,7 @@ func TestDiscoverManifests(t *testing.T) {
 		rootDir           string
 		expectedPipelines []string
 		credentials       map[string]gitProviderToken
+		digest            bool
 	}{
 		{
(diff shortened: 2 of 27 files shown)


olblak:
> Fix part of https://github.com/updatecli/updatecli/issues/3507
Do you need more than the current pullrequest?
It sounds almost complete to me

pkg/plugins/autodiscovery/githubaction/main_test.go
```diff
@@ -96,6 +97,136 @@ conditions:
...
+  release:
+    dependson:
+      - 'condition#release:and'
+    disableconditions: true
+    name: 'deps(github): bump Action release for actions/checkout from v4 to {{ source "release" }}'
+    kind: 'yaml'
+    sourceid: 'release'
+    transformers:
+      - addprefix: '"actions/checkout@'
+      - addsuffix: '"'
```
olblak, line 110: I think it would be better to also include the version as a comment.
Something like:

```suggestion
      - addsuffix: '"'
      - addprefix: "# {{ source "release" }}"
```

The goal is to quickly identify the version associated to the hash and comment are allowed in github action files

pkg/plugins/autodiscovery/githubaction/main.go
```diff
@@ -163,6 +167,10 @@ func New(spec interface{}, rootDir, scmID string) (GitHubAction, error) {
 		newFilter.Kind = defaultVersionFilterKind
 		newFilter.Pattern = defaultVersionFilterPattern
 	}
+	digest := false
```
olblak, line 170:
```suggestion
	digest := true
```

What would you think to enable digest by default? 
With the way comments are handled and It think it's the best of the two worlds

olblak (approved):
Minor comment but the pullrequest is ready to be merged with or without my suggestion

pkg/plugins/autodiscovery/githubaction/main.go
olblak: Keeping the none digest behavior by default for now

Issue #3219, opened 2024-12-03
Feature Request: Support standard enterprise environments


jcmcken:
### Is your feature request related to a problem?

In most enterprises, there is no direct access to the Internet. There are things like outbound HTTP proxies, private mirrors (e.g. Artifactory, Nexus), and so forth. Currently, it doesn't seem like ``updatecli`` considers these use cases nearly at all, or at least doesn't treat them as a first class citizen. The effect of this is this tool cannot be used at all in these kinds of environments.

The latest incarnation of this problem for me (although there are other users reporting similar or related problems such as #1623 and #1370) is that it doesn't seem like ``updatecli`` supports standard Go implementation for configuring a [HTTP proxy](https://pkg.go.dev/golang.org/x/net/http/httpproxy). For example, if I try to use ``githubrelease`` after setting HTTP proxy in my environment, it ignores those settings, and tries to reach out directly to Github's public IP addresses.

### Solution you'd like

At the least, any plugin that talks over HTTP should support HTTP proxy settings. This way even if we can't yet make use of private mirrors (e.g. ``GOPROXY`` or equivalent), we could reach out to the Internet.

### Alternatives you've considered

There's not really any alternative. Direct Internet access is blocked, and private mirrors aren't really supported. So it seems at the least we need to be able to configure HTTP proxy

olblak:
Thanks, @jcmcken , for the great feedback.

You are kind of right that "enterprise" environment wasn't the initial target for Updatecli.
Probably the main reason for that is that I've been working full-time on OSS projects for almost the past decade,
and Updatecli started as a side project to help me to maintain those projects. 

It turns out that Updatecli is not anymore a small side project.
We extended the scope of Updatecli by a lot, and we keep improving step by step
Hence, the various git provider support like Bitbucket/Stash, GitHub enterprise support, etc.

Updatecli is built with extensibility in mind. Many "enterprise" related features are either contributed by company's employees who need something or by sponsoring.
If something is urgent/important for your company, that could help your company deliver faster, then feel free to drop an email on info@updatecli.io.
There are Updatecli maintainers available for short term consulting missions.

Regarding the Maven and Go improvements, it's something I have on my radar, unless priority changes it's probably one of the next things I'll look into.

I like very much your suggestion about supporting HTTP_PROXY. While I doubt it's gonna be very difficult, it may take some time to review all plugins in order to support HTTP_PROXY

PR #3516, opened 2025-01-09
improvement(httpclient): add support for http proxy in retry client

ryancurrah:
The custom retry http client does not current support http proxies. This PR adds support for using proxies variables if set.

Issue: GH-3219

Fix GH-3219


olblak:
@ryancurrah Thanks for the pullrequest, I added one commit to consolidate on the httpclient so all http request can handle proxy. Do you have access to a restricted environment? To tests that all e2e works as expected? 

```yaml
go build -o bin/updatecli . 
./bin/updatecli diff --config e2e/updatecli.d/success.d/
```

ryancurrah:
I just ran the tests. They look like they are working for the URLs that are allow listed in the egress proxy.

olblak:
Awesome, thanks for looking into this

olblak (approved):

PR #3538, opened 2025-01-12
feat: by default create github commit via API

olblak:
No breaking changes should be introduced by this pullrequest, but pipeline using gpg key will now show the warning.
The warning can be removed by setting `commitusingapi`: false

By default, the GitHub SCM now creates commits using the GitHub API, so commits are signed by GitHub if Updatecli is executed from GitHub action using the token GITHUB_TOKEN.

While the purpose of this change is to increase security, this behavior can be disabled by setting the parameter `commitusingapi: false` such as:

```
scms:
  default:
    kind: github
    spec:
      owner: git_repo_owner
     repository: git_repo_name
     commitusingapi: false
```

When set to false, Updatecli creates commit in a more traditional way by running git command.
As a reminder, Updatecli also allow sign commits using a gpg key using the parameter `gpg`
with an example on  https://www.updatecli.io/docs/core/scm/ but this behavior is not compatible with commit over API

## Test

Nothing to test, as this feature is used in production for over 6 months

### DIFF
--- a/pkg/plugins/scms/git/sign/main.go
+++ b/pkg/plugins/scms/git/sign/main.go
@@ -39,3 +39,8 @@ func GetCommitSignKey(armoredKeyRing string, keyPassphrase string) (*openpgp.Ent
 
 	return key, nil
 }
+
+// IsZero returns true if the GPGSpec is empty
+func (g GPGSpec) IsZero() bool {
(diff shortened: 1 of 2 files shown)


olblak:
Due to #5323 
I am closing this issue.
I am not working on it at the moment

PR #3903, opened 2025-02-17
fix(maven): check all configured repositories before returning an error

### LINKED ISSUE #3889: Maven doesn't check all the repositories

### Current Behavior

Hi,

I've noticed that in case a Maven project has multiple repositories, updatecli fails if the first repo doesn't contain a given dependency. Say I have two repos: an internal one, containing only the artifacts for a few libraries, and an external one containing everything else. When Maven resolves the dependencies, if it doesn't find an element in the first repo it falls back to the second one; however, updatecli just fails. I think it should check all the repositories and fail only when the dependency is not found in any of them (in other words, I this [this line](https://github.com/updatecli/updatecli/blob/main/pkg/plugins/resources/maven/source.go#L239) should just log a warning).
If you think this behavior needs to be changed I can work on a PR.

Thanks,

lorenzo

### Expected Behavior

updatecli tries to get the dependencies from all the repositories

### Steps To Reproduce

Setup a Maven project with two repos where only one of them contains a given dependency and then run updatecli using Maven autodiscovery

### Environment

```markdown
- OS: Fedora 41
- updatecli: 0.93.1
```

### Pipeline Graph

```mermaid
```

### Anything else?

_No response_

lorenzobenvenuti:
Fix #3889

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist

### DIFF
--- a/pkg/plugins/resources/maven/source.go
+++ b/pkg/plugins/resources/maven/source.go
@@ -20,7 +20,7 @@ func (m *Maven) Source(workingDir string, resultSource *result.Source) error {
 
 		latestVersion, err := metadataHandler.GetLatestVersion()
 		if err != nil {
-			return fmt.Errorf("getting latest version: %w", err)
+			logrus.Warnf("getting latest version: %v", err)
 		}
 
 		if latestVersion != "" {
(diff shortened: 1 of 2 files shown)


olblak (approved):
Thanks for the fix

PR #4426, opened 2025-03-14
feat(shell): set default shell environments variable

### LINKED ISSUE #1768: Feature Request: shell resource should automatically includes common environment variables

While using `kind: shell` as a source I faced some unexpected failures in my script related to lack of $PATH being made available to the shell script run within updatecli.

### Solution you'd like

We should include a common set in the environment in which shell scripts are run, inherited from the environment in which updatecli run, without the need to explicitly list them.

e.g. PATH, HOME, USER

Documentation clearly state:
> By default the command doesn’t inherit any environment variables from the updatecli process. The motivation is to avoid environment variable leak into the command.

but this is probably NOT what the user expect happening when using the shell resource

### Anything else?

_No response_

olblak:
Fix #1768 

Share a set of default environment variable with the shell plugin.
Improve the shell plugin parameter documentation.

## Test

```shell
updatecli diff --config /tmp/updatecli.yaml
```

where updatecli.yaml contains

```
targets:
  default:
    kind: shell
    spec:
      command: updatecli version
```

### Checklist

### DIFF
--- a/pkg/plugins/resources/shell/condition.go
+++ b/pkg/plugins/resources/shell/condition.go
@@ -16,7 +16,8 @@ func (s *Shell) Condition(source string, scm scm.ScmHandler) (pass bool, message
 	// Ensure environment variable(s) are up to date
 	// either it already has a value specified, or it retrieves
 	// the value from the Updatecli process
-	err = s.spec.Environments.Load()
+	ignoreEnvironmentsNotFound := s.spec.Environments == nil
+	err = s.environments.Load(ignoreEnvironmentsNotFound)
 	if err != nil {
 		return false, "", err
 	}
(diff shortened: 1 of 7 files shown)


dduportal:
I got (non blocking!) usage questions:

- Is there a way to opt-out on of the "predefined" env. var? At first sight I would override the env. var I do not want  (with an empty value). Would it work?
- Is there a way to append to an existing default variable? I'm thinking about the `PATH`? (again not blocking and might be a separate topic when a use case comes on)

olblak:
> Is there a way to opt-out on of the "predefined" env. var? At first sight I would override the env. var I do not want (with an empty value). Would it work?

If you set environments to any dumb value, then the predefined environment variables are ignored

> Is there a way to append to an existing default variable? I'm thinking about the PATH? (again not blocking and might be a separate topic when a use case comes on)

Not, in the current pull request, if environments is undefined then it fallbacks to global env variables value.

Also, I realize now that  

```
targets:
  default:
    kind: shell
    spec:
      command: updatecli version
      environments:
        - name: "PATH"
          value: ""
```

is equivalent to 

```
targets:
  default:
    kind: shell
    spec:
      command: updatecli version
      environments:
        - name: "PATH"
```

which retrieve the PATH value from the global context instead of setting PATH to "" in the first example

olblak:
My last commit differentiate an empty env variable from an unset one

PR #4429, opened 2025-03-16
feat(flux): support multiple YAML documents per file

Antvirf:
## Purpose

Currently, only the first yaml doc in a file is processed by Flux autodiscovery. It is quite common - as well as personal preference - to often keep `HelmRepository` and `HelmRelease` close to each other in the same file.

This change allows for defining these resources in the same file. Any number of documents can be specified in one yaml file and will be picked up by `updatecli`.

## Test

```shell
(all tests OK)

❯ pwd && go test .
/updatecli/pkg/plugins/autodiscovery/flux
ok      github.com/updatecli/updatecli/pkg/plugins/autodiscovery/flux   0.010s
```


olblak:
Thanks for the contribution, I think it would sense, and I would be happy to merge it.

olblak:
For info there is something fishy with the golang e2e tests affecting all the tests, I am looking to it

pkg/plugins/autodiscovery/flux/aider.sh
```diff
@@ -0,0 +1,12 @@
...
+export OLLAMA_API_BASE=http://127.0.0.1:11434
+
+export
+
+#aider --model ollama_chat/deepseek-r1:1.5b
+#aider --model ollama_chat/deepseek-coder:6.7b
+
+
+source .env
+aider --model deepseek --api-key deepseek=$DEEPSEEK_API_KEY
```
olblak, line 12: Is this file needed by the pullrequest?
I didn't know that tool, "aider" but looking at the repository, I think it could be removed

Antvirf: Sorted!

olblak:
Just of note for the future myself.

While reviewing this pull request, I stumbled on a question, what if our single file contains multiple helmrelease definition.
Then the query `$.spec.chart.spec.version` will update all of them at once.

We could leverage the yamlpath queries in target to better update a specific document for example with something like: 

```
targets:
    helmrelease:
        name: 'deps(flux): bump Helmrelease "udash"'
        kind: yaml
        spec:
            engine: yamlpath
            file: helmrelease-helmrepository.yaml
            key: $.spec.chart.spec[?(@.chart == "udash")].version
        sourceid: helmrelease
```

After a quick tests, it appears that Updatecli will only write back to the file one document.

The yaml plugin either rely on `github.com/goccy/go-yaml` the default approach or `github.com/vmware-labs/yaml-jsonpath` that offer more advanced yamlpath queries but depends on `https://github.com/go-yaml/yaml` which comes with limitation. One of them, it doesn't handle multiple document out of the box, so some extra effort is need in the target code.
So I think for today, it's better to stick to the default yaml engine and not use

olblak (approved):

PR #4432, opened 2025-03-17
feat: gitlab support authorization with bearer token

alikhil:
This pull request adds support of [Authorization with bearer token](https://docs.gitlab.com/api/rest/authentication/#oauth-20-tokens) for Gitlab SCM plugin which is essential in cases when use of PAT tokens is not possible or not recommended for security reasons.

During testing of this new feature I stuck with strange behavior in untouched part of the code, specifically – updatecli started failing request to docker registry to pull latest helm chart with error like this: 
```
"failed to fetch anonymous token: unexpected status from GET request to https://auth.docker.io/token?scope=repository%3Abitnamicharts%2Fgrafana-operator%3Apull&service=registry.docker.io: 500 Internal Server Error"
```

After hours of debugging I found out that it was because of using and mutating default http client – `http.DefaultClient`, which is later is referenced in code of dependency libs (helm, opencontaiers, etc). To fix this bug and prevent feature problems I have replaced use of `http.DefaultClient` with `&http.Client{}`

## TL; DR;

- Add `Bearer` token support for Gitlab SCM plugin
- Replace usage of `http.DefaultClient` with `&http.Client{}` to fix and prevent bugs on mutated global client
- Fix typo in docs

### Checklist

- [ ] If I get it right, there is no need to update [documentation](https://github.com/updatecli/website/blob/master/content/en/docs/plugins/scm/gitlab.adoc) since it's generated from comment annotations. Tell me if I am wrong


olblak:
Thanks for the pull request.
It sounds good to me, once again I didn't test it, but I am sure you did.

The side issue you faced was really "interesting" :D

olblak:
Thanks for the pull request.
It sounds good to me, once again I didn't test it, but I am sure you did.

The side issue you faced was really "interesting" :D 

Regarding the documentation, you're right, it will be used in your IDE and on updatecli.io

olblak (approved):

PR #4568, opened 2025-03-31
ci: update to golangci v2 configuration && fix reported linting issues

olblak:
golangci-lint update to v2 is currently broken.

While the core of Updatecli shouldn't be modified, it's hard to guarantee that I didn't introduce a mistake in the process
golangci-lint started reporting issues in a many packages such as:

* ST1005: error strings should not end with punctuation or newlines
* QF1003: could use tagged switch on gr.spec.Key1
* ST1005: error strings should not be capitalized
* QF1001: could apply De Morgan's law 
* SA5001: should check error returned from os.Create() before deferring newFile.Close()
* SA4014: this condition occurs multiple times in this if/else if chain
* SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead
* QF1007: could merge conditional assignment into variable declaration
* SA6005: should use strings.EqualFold instead
* SA9003: empty branch
* QF1008: could remove embedded field "InlineKeyChain" from selector
* SA1001: template: :6: unexpected "}" in end 
* QF1004: could use strings.ReplaceAll instead

## Test

```shell
make test
```

### Checklist

### DIFF
--- a/pkg/core/config/template.go
+++ b/pkg/core/config/template.go
@@ -158,7 +158,7 @@ func mergeValueFile(valuesFiles ...map[string]interface{}) (results map[string]i
 
 	for _, values := range valuesFiles {
 		if err := mergo.Merge(&results, values, mergo.WithOverride); err != nil {
-			err = fmt.Errorf("Error while merging values files: %v", err)
+			err = fmt.Errorf("merging values files: %w", err)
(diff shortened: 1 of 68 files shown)


olblak:
I am merging this pr hoping that I am not introducing regression

PR #4793, opened 2025-04-09
feat(autodiscovery/cargo): switch cargo update to cargo-upgrade

### LINKED ISSUE #4792: Feature Request: Improve cargo resource

Currently the `cargopackage` resource reformat the `Cargo.toml` completely when doing a version update.

### Solution you'd like

The Cargo.toml should not be reformatted, only the version should be updated

_No response_

### Anything else?

_No response_

loispostula:
Fix #4792 

Change the update in the Cargo.toml from `toml` to `shell` with `cargo-upgrade`.

This allows for the change to not reformat the `Cargo.toml`

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist

### Tradeoff

[`cargo-edit`](https://crates.io/crates/cargo-edit) becomes a requirement for this to work


olblak:
This would be a breaking change today, what if the cargo update is not available? 
I think it would be better to have a second template if cargo is available then run this 
otherwise fallback to the toml plugin.

We uses this approach on the npm plugin which I like, it provides an easy fallback

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -40,17 +40,83 @@ sources:
...
+          ARGS="--dry-run"
+        fi
+        cargo upgrade $ARGS --manifest-path {{ .File }} --package {{ .DependencyName }}@{{"{{"}} source "{{ .SourceID }}" {{"}}"}}
+        cargo update $ARGS --manifest-path {{ .File }} {{ .DependencyName }}@{{"{{"}} source "{{ .SourceID }}" {{"}}"}} --precise {{"{{"}} source "{{ .SourceID }}" {{"}}"}}
+      changedif:
+        kind: file/checksum
+        spec:
+          files:
+            - "{{ .File }}"
+            - "Cargo.lock"
```
olblak, line 71: Is the `Cargo.lock` file always at the same location? Shouldn't it be next to the Cargo.toml file?

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -40,17 +40,83 @@ sources:
...
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      command: |
+        ARGS=""
+        if [ "$DRY_RUN" = "true" ]; then
+          ARGS="--dry-run"
+        fi
+        cargo upgrade $ARGS --manifest-path {{ .File }} --package {{ .DependencyName }}@{{"{{"}} source "{{ .SourceID }}" {{"}}"}}
```
olblak, line 64: Is it `cargo upgrade` or `cargo upgrades`
```suggestion
        cargo upgrades $ARGS --manifest-path {{ .File }} --package {{ .DependencyName }}@{{"{{"}} source "{{ .SourceID }}" {{"}}"}}
```

olblak: Thanks I just found the link.
I think it would be better to update the message to suggest installing 

* https://github.com/killercup/cargo-edit

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -40,17 +40,83 @@ sources:
...
+    kind: 'shell'
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      command: |
+        ARGS=""
+        if [ "$DRY_RUN" = "true" ]; then
+          ARGS="--dry-run"
+        fi
```
olblak, line 63:
```suggestion
        fi
        # Require cargo upgrade from https://github.com/killercup/cargo-edit
```

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
loispostula: I've done a better rewrite where I handle this correctly

olblak:
I just tested on https://github.com/updatecli-test/kwctl/pull/1
It looks great

olblak:
I think it would be nice to update the documentation to highlight that Updatecli leverages cargo-edit if it's installed

olblak (approved):

pkg/plugins/autodiscovery/cargo/dependencyManifest.go
```diff
@@ -40,17 +40,83 @@ sources:
...
+conditions:
+  {{ .ConditionID }}:
+    name: 'Test if version of "{{ .DependencyName }}" {{"{{"}} source "{{ .ExistingSourceID }}" {{"}}"}} differs from {{"{{"}} source "{{ .SourceID }}" {{"}}"}}'
+    kind: 'shell'
+    sourceid: '{{ .SourceID }}'
+{{- if .ScmID }}
+    scmid: '{{ .ScmID }}'
+{{- end }}
+    spec:
+      command: 'test {{"{{"}} source "{{ .ExistingSourceID }}" {{"}}"}} != '
```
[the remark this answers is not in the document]
loispostula: As the source is fetch from crates.io (or compatible registry) I think that we'll only get semver compatible version out of this.

I also think that we should be able to drop this condition if we use the cargo-edit. As before we were reformatting the whole file. I'd use the condition to ensure we were not updating the `Cargo.(toml|lock)`. Now we do in plce replacement so the condition is not really needed

PR #4957, opened 2025-04-28
feat(autodiscovery): Add Nomad integration

### LINKED ISSUE #2331: Feature Request: Add autodiscovery for Hashicorp Nomad

I would like to automate docker image update defined in Hashicorp Nomad configuration

### Solution you'd like

/

/

### Anything else?

/

olblak:
Fix #2331 

Add Nomad autodiscovery plugin to automate docker image update.
Please note that it only supports the Docker driver and HCL files.
We could improve it in the future to also handle JSON files

It handles the following scenario:

* simple scenario such as `image = "hashicorp/http-echo:latest"`
* handle scenario with one variable such as `image = "grafana/grafana:${var.image_tag}"`

In the second case, Updatecli will update the variable definition and not the task definition

## Test

```shell
cd pkg/plugins/autodiscovery/nomad/
go test 
```

I also did some testing on https://github.com/fhemberger/nomad-demo

### Checklist

- [x] [website#2151](https://github.com/updatecli/website/pull/2151) repository.

### Potential improvement

- [ ] Support credentials specified from config
- [ ] Support json manifest

### DIFF
--- a/pkg/plugins/autodiscovery/nomad/main.go
+++ b/pkg/plugins/autodiscovery/nomad/main.go
@@ -16,7 +16,7 @@ import (
 type Spec struct {
 	// digest provides parameters to specify if the generated manifest should use a digest on top of the tag.
 	Digest *bool `yaml:",omitempty"`
-	// rootDir defines the root directory used to recursively search for Helm Chart
+	// rootDir defines the root directory used to recursively search for Nomad files
 	// If rootDir is not provided, the current working directory will be used.
 	// If rootDir is provided as an absolute path, scmID will be ignored.
 	// If rootDir is not provided but a scmid is, then rootDir will be set to the git repository root directory.
--- a/pkg/plugins/autodiscovery/nomad/main.go
+++ b/pkg/plugins/autodiscovery/nomad/main.go
@@ -1,0 +1,139 @@
+package nomad
+
+import (
+	"path"
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/sirupsen/logrus"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/docker"
+	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
+)
+
+// Spec is a struct fill from Updatecli manifest data and shouldn't be modified at runtime unless
+// For Fields that requires it, we can use the struct DockerCompose
+// Spec defines the parameters which can be provided to the Helm builder.
+type Spec struct {
+	// digest provides parameters to specify if the generated manifest should use a digest on top of the tag.
+	Digest *bool `yaml:",omitempty"`
+	// rootDir defines the root directory used to recursively search for Helm Chart
+	// If rootDir is not provided, the current working directory will be used.
+	// If rootDir is provided as an absolute path, scmID will be ignored.
+	// If rootDir is not provided but a scmid is, then rootDir will be set to the git repository root directory.
+	RootDir string `yaml:",omitempty"`
+	// ignore allows to specify rule to ignore autodiscovery a specific Helm based on a rule
+	Ignore MatchingRules `yaml:",omitempty"`
+	// only allows to specify rule to only autodiscover manifest for a specific Helm based on a rule
+	Only MatchingRules `yaml:",omitempty"`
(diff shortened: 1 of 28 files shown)


pkg/plugins/autodiscovery/nomad/main.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/updatecli/updatecli/pkg/plugins/utils/version"
+)
+
+// Spec is a struct fill from Updatecli manifest data and shouldn't be modified at runtime unless
+// For Fields that requires it, we can use the struct DockerCompose
+// Spec defines the parameters which can be provided to the Helm builder.
+type Spec struct {
+	// digest provides parameters to specify if the generated manifest should use a digest on top of the tag.
+	Digest *bool `yaml:",omitempty"`
+	// rootDir defines the root directory used to recursively search for Helm Chart
```
olblak, line 19:
```suggestion
	// rootDir defines the root directory used to recursively search for Nomad files
```

pkg/plugins/autodiscovery/nomad/matchingRule.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// MatchingRule allows to specifies rules to identify manifest
+type MatchingRule struct {
+	// Path specifies a Helm chart path pattern, the pattern requires to match all of name, not just a substring.
```
olblak, line 12:
```suggestion
	// Path specifies a Nomad path pattern, the pattern requires to match all of name, not just a substring.
```

loispostula (dismissed):
This looks good, maybe it needs to be a touch clearer that it's only for the docker driver in hcl files, as you can have different drivers and also json spec

olblak:
> This looks good, maybe it needs to be a touch clearer that it's only for the docker driver in hcl files, as you can have different drivers and also json spec

Indeed good point

allroundtechie:
Just want to mention, docker is not the only driver for container related workloads, there is also a [podman](https://developer.hashicorp.com/nomad/plugins/drivers/podman) driver.
Also authentication against a private container registry would be awesome.

loispostula:
> Just want to mention, docker is not the only driver for container related workloads, there is also a [podman](https://developer.hashicorp.com/nomad/plugins/drivers/podman) driver. Also authentication against a private container registry would be awesome.

@allroundtechie usually this is left to the caller of updatecli, typically in ci what i do is have a login step (for docker registries, npm, cargo, ...) then call updatecli

olblak:
The way I see authentication today:

* Rely on the default docker configuration (usually my preferred approach)
* Specify credentials in the autodiscovery plugin, I must admit I never use this
* Not supported yet, but we could also load information from nomad configuration as explained [here](https://developer.hashicorp.com/nomad/docs/drivers/docker#authentication)

allroundtechie:
> The way I see authentication today:
> 
> * Rely on the default docker configuration (usually my preferred approach)
> * Specify credentials in the autodiscovery plugin, I must admit I never use this
> * Not supported yet, but we could also load information from nomad configuration as explained [here](https://developer.hashicorp.com/nomad/docs/drivers/docker#authentication)

Just my two cents:
I do not know what the majority of updatecli users do or how they use updatecli, I personally use it within CI pipelines only. That said you do not want to have credentials or other security related stuff in clear text in config files or in this case hcl/nomad files for authentication.
I struggled with this myself when deploying nomad job via a CI pipeline and do a replacement of  "dummy" credentials in the nomad/hcl file with the real credentials form the vault or secrets store e.g.
```
sed -i 's|REGISTRY_USERNAME|'${{ env.USERNAME }}'|g' service.hcl
sed -i 's|REGISTRY_PASSWORD|'${{ env.TOKEN_PACKAGE_READ }}'|g' service.hcl
```
So from a security point of view the last option @olblak mentioned is at least in context of CI pipelines actually a no go. But of course this could make sense for other people who do not use CI pipelines.
Basically @loispostula mentioning the docker login setup in the CI pipeline is all which is needed.

loispostula:
@allroundtechie back when I was using nomad, but I do use it for my k3s clusters, I actually set up `containerd` with authenticated mirrors on the hosts and don't even bother with credentials in the job definitions

PR #5094, opened 2025-05-09
feat(gitlab/mergerequest): configure additional fields during mr creation

alikhil:
This PR add support of setting additional Gitlab MR options during it's creation:
- `labels`
- `reviewers`
- `assignees`
- `squash`
- `removeSourceBranch`

## Additional Information

All interaction with Gitlab in gitlab/mergerequest action have been rewritten using [gitlab-api-client-go](https://gitlab.com/gitlab-org/api/client-go) libarary


pkg/plugins/resources/gitlab/mergerequest/spec.go
```diff
@@ -84,4 +84,58 @@ type Spec struct {
 			"body" is useful to provide additional information when reviewing mergerequest, such as changelog url.
 	*/
 	Body string `yaml:",omitempty"`
+
+	/*
+		"assignees" contains the list of assignee to add to the merge request
```
olblak, line 89: Nitpick but I find it better to use the comment syntax `//` over `/* */` because in the latter, our IDE add `\t` which kind of change the documentation visualisation such as:

https://www.updatecli.io/docs/plugins/resource/hcl/ 

vs 

https://www.updatecli.io/docs/plugins/resource/yaml/

olblak:
It looks good to me, I am wondering why the test is failing as it doesn't look like the usual flaky e2e tests

pkg/plugins/resources/gitlab/mergerequest/spec.go
alikhil: Noted, I will change it. Just was trying to do it in the same way as it already was

pkg/plugins/resources/gitlab/mergerequest/utils.go
```diff
@@ -2,68 +2,66 @@ package mergerequest
...
+		mergeRequests, resp, err := g.api.MergeRequests.ListProjectMergeRequests(
+			g.getPID(),
+			&optsList,
+			gitlabapi.WithContext(ctx),
 		)
 
-		if err != nil {
-			logrus.Debugf("RC: %d\nBody:\n%s", resp.Status, resp.Body)
-			return "", "", "", err
+		if err != nil || resp.StatusCode > 400 {
```
olblak, line 39:
```suggestion
		if err != nil || resp.StatusCode > 400 {
```

Both err and resp can't be tested at the same time because resp is nil if err is different than nil
cfr https://gitlab.com/gitlab-org/api/client-go/-/blob/v0.128.0/merge_requests.go#L791

We should split the condition into two different part

pkg/plugins/resources/gitlab/mergerequest/utils.go
```diff
@@ -102,26 +100,25 @@ func (g *Gitlab) isRemoteBranchesExist() (bool, error) {
...
 		)
 
-		if err != nil {
-			logrus.Debugf("RC: %d\nBody:\n%s", resp.Status, resp.Body)
-			return false, err
-		}
-
-		if resp.Status > 400 {
-			logrus.Debugf("RC: %d\nBody:\n%s", resp.Status, resp.Body)
+		if err != nil || resp.StatusCode > 400 {
```
olblak, line 118: Same here, resp and err can't be tested at the same time becaues resp is nil if err is not 

https://gitlab.com/gitlab-org/api/client-go/-/blob/v0.128.0/branches.go#L84

pkg/plugins/resources/gitlab/mergerequest/utils.go
olblak: Please note that it's useful to check for error code

olblak:
I had a quick look to the drone/go-scm library and there, they always return the response, which is useful to get additional information https://github.com/drone/go-scm/blob/a515870a3931d01562ab61723891462134a3277f/scm/driver/gitlab/pr.go#L80

Especially when the request fails for situation like authentication issues that are not catch by the err handling

pkg/plugins/resources/gitlab/mergerequest/utils.go
olblak: btw checking the StatusCode is useless with this library as it's already handled 
https://gitlab.com/gitlab-org/api/client-go/-/blob/v0.128.0/gitlab.go#L1031

alikhil: Removed redundant check

pkg/plugins/resources/gitlab/mergerequest/utils.go
alikhil: Fixed

olblak (approved):
Thanks for the pullrequest.

While I tested it, I didn't try to open a mergerequest

PR #5133, opened 2025-05-18
feat(bitbucket): support pull request updates

mcwarman:
Update bitbucket pull request title and description on change.

## Test

```shell
go build -o bin/updatecli
export BITBUCKET_TOKEN="xxx"
# update scm.yaml to hit test repo etc.
bin/updatecli --config e2e/updatecli.d/success.d/bitbucket/scm.yaml --debug apply
# update scm.yaml to change Target and PR Title
bin/updatecli --config e2e/updatecli.d/success.d/bitbucket/scm.yaml --debug apply
```

### Checklist

### Tradeoff

Required to remove the type alias (`type Client *scm.Client`) so the [`b.client.Do`](https://github.com/drone/go-scm/blob/a515870a3931d01562ab61723891462134a3277f/scm/client.go#L157) method could be used.

### Potential improvement

Further extension similar to https://github.com/updatecli/updatecli/pull/5094 with potentially moving away from drone scm, which isn't as feature rich as alternatives


pkg/plugins/resources/bitbucket/client/main.go
```diff
@@ -52,9 +52,7 @@ type Spec struct {
 	Repository string `yaml:",omitempty" jsonschema:"required"`
 }
 
-type Client *scm.Client
```
mcwarman, line 55: Removed the type alias so the [`b.client.Do`](https://github.com/drone/go-scm/blob/a515870a3931d01562ab61723891462134a3277f/scm/client.go#L157) method could be used.

olblak: Thanks for the explanation

olblak:
Have you tested on your bitbucket instance?

Does it erase the pullrequest description or merge the previous and new description

olblak:
I don't think you merge existing pullrequest content with the new one.
You can find some inspiration about how it is done on the GitHub or GitLab one
 
https://github.com/updatecli/updatecli/blob/09608ffe8c1c0486f5553df12cdc4df371a5ca7b/pkg/plugins/resources/gitlab/mergerequest/create.go#L40

Merging pullrequest body is needed when you have two different pipeline updating the same pullrequest (using pipelineid)

olblak (changes requested):
Thanks for the contribution.
It would be nice to also merge pullrequest content.

mcwarman:
> Does it erase the pullrequest description or merge the previous and new description

I had seen that it did this, but I don't think I quite realised it was a solved issue.

I'll look into it, might be a challenge with the markdown that Bitbucket requires but I'll have a go.

olblak:
Does it support HTML content?

If not we would need to build a new Markdown parser for the pullrequest body as we would need to have to store
some private Updatecli information such as the pipelineid, etc.

If the latter is required, then I would implement it in a following iteration.

olblak:
Your change is already a nice step forward.
What do you think?

mcwarman:
It doesn't support HTML, it's also has a very limited set of markdown support.

I'd started to look at the parser but it's a much bigger change. So possibly makes sense to do it separately as a follow up. From a feature separation and ease of review.

Probably the most annoying thing is it doesn't support comments to hide IDs in the body.

So we'll likely have to change the format to include the IDs so they can be parsed and merged. 

The change going as is would be useful for us, as we use the title for squash merge messages and at the moment if there's another update we have to remember to update the title to the new version number.

olblak (approved):
I created https://github.com/updatecli/updatecli/issues/5174 to keep track of the following work.
For now, I am happy to merge your pr, as I think it's already an improvement.

Thanks for your patience

Issue #5174, opened 2025-05-24
Feature Request: Allow to merge bitbucket pullrequest body


olblak:
### Is your feature request related to a problem?

Once https://github.com/updatecli/updatecli/pull/5133 is merged, two different Updatecli pipelines (different manifest) targeting the same pullrequest will override the pullrequest description depending on which one run last.
It's worth mentioning that the commit won't be overridden, so the real change will be applied but with a wrong pull request description.

### Solution you'd like

Bitbucket pull request description doesn't support HTML content.
This is unfortunate as we can't use the same technique used for GitHub/GitLab/Gitea where we use HTML div tag to store pull request metadata such as pipelineid.

Bitbucket supports a very limited version of Markdown so we need to define a new standard content

PR #5189, opened 2025-05-27
fix(reports): ensure latest title and description are used

mcwarman:
I'm looking at the #5174,

I noticed something around the behaviour of the current merge, that I'd think is bug, this tries to address that.

If there's match on Ids for Action or ActionTarget the Description and Title are not always updated with the new report. This is due to how the logic can flip which is the new report, based on the number of actions or targets.

## Test

```shell
cd pkg/core/reports
go test
```

### Checklist

### Tradeoff

N/A

### Potential improvement

N/A


olblak:
Interesting, I always assumed it was two different pipeline using the same pipelineid and both enforcing different pr title
Thanks for the catching it

olblak (approved):

PR #5299, opened 2025-06-01
fix: remove duplicate generate schema

qianlongzt:
remove duplicate `s.GenerateSchema(spec)`

## Test

```shell
go build
.\updatecli.exe jsonschema
```

## Question

Should I remove one of them?

I checked https://www.schemastore.org/json/, and only `policy/manifest` exists.

```go
        if err = generateSchema(baseSchemaID, schemaDir, "", config.Spec{}); err != nil {
		return fmt.Errorf("unable to generate schema - %s", err)
	}
	if err = generateSchema(baseSchemaID, schemaDir, "policy/manifest", config.Spec{}); err != nil {
		return fmt.Errorf("unable to generate schema - %s", err)
	}
```

### DIFF
--- a/pkg/core/engine/schema.go
+++ b/pkg/core/engine/schema.go
@@ -57,10 +57,6 @@ func GenerateSchema(baseSchemaID, schemaDir string) error {
 		return nil
 	}
 
-	if err = generateSchema(baseSchemaID, schemaDir, "", config.Spec{}); err != nil {
-		return fmt.Errorf("unable to generate schema - %s", err)
-	}
-
 	if err = generateSchema(baseSchemaID, schemaDir, "policy/manifest", config.Spec{}); err != nil {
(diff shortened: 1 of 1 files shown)


olblak:
Thank you very much, that's a good catch.
```
        if err = generateSchema(baseSchemaID, schemaDir, "", config.Spec{}); err != nil {
		return fmt.Errorf("unable to generate schema - %s", err)
	}
```

Can also be removed now, as it is not used anymore
I checked on https://github.com/SchemaStore/schemastore/blame/b741b07dcf60330e07d2d0af9c553a60c6d381f7/src/api/json/catalog.json#L6927 and it currently uses the one named "manifest/config.json"

qianlongzt:
I remove `generateSchema(baseSchemaID, schemaDir, "", config.Spec{})`

olblak (approved):

PR #5309, opened 2025-06-03
fix(reports): update target description to output correctly in markdown

mcwarman:
Again in my battle with markdown reports for bitbucket i noticed a bug in the descriptions.

We use `...\n\t* ...` throughout the code ([here](https://github.com/updatecli/updatecli/blob/79f248bedcbbbceb7a28e888ca4e359b949b7792/pkg/plugins/resources/yaml/target.go#L100), [here](https://github.com/updatecli/updatecli/blob/79f248bedcbbbceb7a28e888ca4e359b949b7792/pkg/plugins/resources/json/target.go#L151) and [here](https://github.com/updatecli/updatecli/blob/79f248bedcbbbceb7a28e888ca4e359b949b7792/pkg/plugins/resources/file/target.go#L188)), this doesn't output correctly in GitHub or Bitbucket (and probably GitLab) but haven't checked.

This can be replicated in common mark tester [here](https://spec.commonmark.org/dingus/?text=Something%20change%0A%09*%20value) 

<details><summary>GitHub Example</summary>
<p>

image

</p>
</details> 

Ref: https://github.com/updatecli/updatecli/pull/5291

<details><summary>Bitbucket Before</summary>
<p>

image

</p>
</details> 

<details><summary>Bitbucket After</summary>
<p>

image

</p>
</details> 

## Test

```shell
cd pkg/core/reports
go test
```

### Checklist

### DIFF
--- a/pkg/core/reports/action.go
+++ b/pkg/core/reports/action.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"sort"
+	"strings"
 	"text/template"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


mcwarman:
This might not fix it in GitHub, because it looks to do some very deliberate escaping.

PR #5350, opened 2025-06-08
feat(bitbucket): add support for pull request description merging

### LINKED ISSUE #5174: Feature Request: Allow to merge bitbucket pullrequest body

mcwarman:
Fix #5174

Merges markdown pipeline and targets output together.

Uses [goldmark](https://github.com/yuin/goldmark) ast to parse the markdown. 

It makes assumptions about the format, and possibly has edge cases missed. That might come out in the wild.

Also fix a bug with #5309 that doesn't support more than one item in the list.

## Test

```shell
cd pkg/core/reports
go test
```

```shell
go build -o bin/updatecli
export BITBUCKET_TOKEN="xxx"
# update scm.yaml to hit test repo etc.
bin/updatecli --config e2e/updatecli.d/success.d/bitbucket/scm.yaml --debug apply
# update scm.yaml to change Target and PR Title
bin/updatecli --config e2e/updatecli.d/success.d/bitbucket/scm.yaml --debug apply
```

### Checklist

### DIFF
--- a/pkg/core/reports/action.go
+++ b/pkg/core/reports/action.go
@@ -114,12 +114,16 @@ func (a *Action) sort() {
 // updateTargetDescriptions updates descriptions from being console friendly to markdown friendly
 func (a *Action) updateTargetDescriptions() {
 	for id, target := range a.Targets {
-		a.Targets[id].Description = strings.ReplaceAll(target.Description, "\n\t*", "\n\n*")
+		d := target.Description
+		d = strings.Replace(d, "\n\t*", "\n\n*", 1)
+		d = strings.ReplaceAll(d, "\n\t*", "\n*")
+		a.Targets[id].Description = d
 	}
 }
 
 // ToActionsString show an action report formatted as a string
 func (a Action) ToActionsString() string {
+	a.sort()
 	a.updateTargetDescriptions()
 
 	output, err := xml.MarshalIndent(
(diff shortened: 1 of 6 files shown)


mcwarman:
This is arguably a breaking/annoying change. because there's no way to merge previously created PRs:

image

Probably just a simple release note saying that it will cause outputs like this, either close and delete branch or ignore.

mcwarman:
> > This is arguably an annoying change. because there's no way to merge previously created PRs:
> 
> 
> 
> Just to clarify, any PR created before this change won't be handled, right? 
> 
> If that's what you mean, then it's totally fine.

Yeah, exactly if it's been created before this realise it will cause the output in the screenshot on subsequent updates.

PR #5486, opened 2025-06-25
fix: Branch reset causes initial Github API commit to be lost 

### LINKED ISSUE #5463: Github API - Diverging branch causes initial commit to be lost if branch already exists

### Current Behavior

We've been seeing intermittent issues with some file changes not actually being applied and I finally had some time to look through it further than having to do manual checks and commits.

If base branch and head branch diverges and `force` is enabled (which is required with API), the initial commit will be lost during the reset and subsequent push.

I'm gonna take a guess straight away that this might come down to;

1. Branch is shown as diverging and reset to base.
2. Commit is done using API.
3. Pull from remote fails as its not able to fast-forward.

Haven't had time to investigate this more but will try to do so during the week if time allows.

(I implemented the initial Github API feature so I'm probably responsible for introducing this bug but haven't been used enough to be bothered by it until now). 

### Expected Behavior

Commits should be kept when resetting diverging branch.

### Steps To Reproduce

1. Given a simple configuration of a Github SCM with `commitUsingApi: true` and a single YAML target:

    ```yaml
    name: "Bump helm charts versions"
    
    pipelineid: "example"
    
    scms:
      default:
        kind: github
        spec:
          ...
          hideCredit: true
          commitUsingApi: true
          force: true
    
    targets:
      HelmValuesImageTag:
        name: "Update"
        kind: yaml
        scmid: default
        spec:
          file: "values.yaml"
          key: "$.imageTag"
          value: "{{ requiredEnv .version }}"
    ````

2. Run `updatecli` to create the initial commit.
3. Make any change in the base branch.
4. Run `updatecli` again. The initial commit will be missing and a new one has not been pushed.
5. If ran once more, it will work as expected.

### Environment

```markdown
- OS: Same issue on Ubuntu 22.04 (Github Actions) and locally on Mac
- updatecli: v0.101.0
```

### Pipeline Graph

```mermaid
```

### Anything else?

_No response_

MattiasAng:
Fix #5463 

Introduces a new variable to track if the locally checked out branch was reset from based branch. This is then utilised to make sure that the reset is pushed to remote **before** Github API commit is done as in its current behaviour it would be force pushed away when subsequent push it done due to remote pull failing due to a no fast-forward error.

## Test

Follow steps to reproduce in corresponding issue once from latest release and then from this pull request.
In the latest release the initial commit got forced pushed away due to branch reset, while with the changes in this pull request the branch reset is pushed before any commit is done.

### Checklist


pkg/plugins/scms/github/scm.go
```diff
@@ -154,14 +163,13 @@ func (g *Github) CreateCommit(workingDir string, commitMessage string) error {
...
 		headOid = sourceBranchRepoRef.HeadOid
-
 		logrus.Debugf("Branch %s does not exist, creating it from commit %q", workingBranch, headOid)
-
 		if err := g.createBranch(workingBranch, repoRef.ID, headOid); err != nil {
 			return err
 		}
 	}
 
+	repositoryName := fmt.Sprintf("%s/%s", g.Spec.Owner, g.Spec.Repository)
```
MattiasAng, line 172: Moved just to have it closer to the actual usage of the variable.

pkg/plugins/scms/github/scm.go
```diff
@@ -140,7 +140,16 @@ func (g *Github) CreateCommit(workingDir string, commitMessage string) error {
 		return err
 	}
 
-	repositoryName := fmt.Sprintf("%s/%s", g.Spec.Owner, g.Spec.Repository)
+	if g.nativeGitHandler.IsForceReset() {
```
olblak, line 143: I am wondering if it's an operation we should only do if committing via the API is configured.
Otherwise, we stick to the default behavior similar to the other scm

olblak: We can just add a second condition.
What do you think?

MattiasAng: This function is only used for committing through API. I think you may be confusing it with the interface function Commit (below) which contains the logic around when to use `CreateCommit` (API) or the native Git handler to commit.

https://github.com/updatecli/updatecli/blob/7272f0a48cc5867507bf30142c6c7a7c824dd50c/pkg/plugins/scms/github/scm.go#L71-L85

olblak: Thank you for the clarification

Issue #5461, opened 2025-06-21
Feature Request: Add `labels` field to every action, even if the underlying SCM doesn't support it


AmandaCameron:
### Is your feature request related to a problem?

I've got some ideas for the policies repo to make it so they can be reused on non-github scms easily, the only real wrinkle in that is that some of the SCMs don't have a labels field. While the SCM I'm most interested in (Gitea) *does* have a labels field in the go-scm/scm library, it doesn't in the action, which would mean my PR to the policies would be broken for gitea, and bitbucket/stash as well. While I could make a PR to enable the labels in gitea, as go-scm does support labels for gitea, I feel that would be a half-assed solution.

### Solution you'd like

Adding a `labels` field to the three actions currently lacking it (gitea, bitbucket,stash). Document in bitbucket/stash that they are a no-op, only there for compatibiility.

### Alternatives you've considered

Another solution might be to take the scm.PullRequestInput struct, and make it a generic Spec that the PR impls all use, and document that usability of the fields in the Spec are dependent on go-scm supporting them.

As far as I can tell, go-scm doesn't throw any errors if you try and pass Labels to a scm driver that doesn't support labels (Correct me if I'm wrong!) so I see no harm in having a common "PRSpec" type that has a method to turn it into a go-scm PullRequestInput

### Anything else?

I'm open to making this PR myself if the maintainers don't have the time/resources to implement this, I just wanted to make a discussion Issue beforehand to get some input on the direction this should take.

I believe there's a good value in having a common, standard set of fields in the action's specs, to allow more interoperability of policies.

The whole thing that sent me down this rabbithole is a PR I plan to make to the policies to abstract out the github dependence, after all.

AmandaCameron:
Also another field which would be needed is `automerge`

olblak:
Happy to see that those policies could be useful to you.
You are right, those policies are for GitHub only at the moment, and it would be nice to make them working for other Git provider.

The way to make it working today is by using Go templating with `if else` statement.
I am wondering if having something like this from Helm project would be useful https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files
It could make those policies more maintainable by having a file named `_scm.tpl` with needed configuration. We would still need Go templating but we could easily copy/paste the same file in all policies.

For the time being, I wouldn't add `label` and `automerge` to all scm if it's not relevant. The challenge here is that different git provider accepts different settings, for example `automerge` is not supported on stash and will probably never be.
[...]

PR #5508, opened 2025-06-28
feat: add partial manifest

### LINKED ISSUE #430: Ability to globally define the `scms` block (v0.17+)
**Is your feature request related to a problem?**
Looking at the new syntax of updatecli manifest for version 0.17.0 and above like [in this PR](https://github.com/jenkins-infra/kubernetes-management/pull/1816/files), we can see the block `scms` is repeated word for word in every manifest.

I'd like the ability to define it globally so we don't have to repeat it on every manifest needing a pull request.

N/A

**Additional context**
N/A

olblak:
Fix #430

As discussed in https://github.com/updatecli/updatecli/issues/5461#issuecomment-3011939378
I propose to introduce partial manifest, similar to how the [Helm](https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files) project is doing.

A partial manifest is a specific type of Updatecli manifest starting with `_` such as `_scm.yaml` that is available to 
all the Updatecli manifest within the same directory.

```
WARNING: Partial file can implement any Go template but will not be considered as a standalone manifest.
Therefor this would be considered as a breaking change
```

## Test

```shell
go build -o bin/updatecli . 
./bin/updatecli manifest show --config e2e/updatecli.d/success.d/partial
```

You will notice that it only generate one Updatecli manifest by combining the files `_source.yaml` and `jenkins.yaml`
The file `stable/_source.yaml` is correctly ignore as being a partial file

### Checklist

- [x]  I have updated the documentation via pull request in [website#2279](https://github.com/updatecli/website/pull/2279) repository.

### Tradeoff

In the current implementation all the files are concatenated together so any YAML file starting with `---` will be considered as a different YAML document which may lead to some confusion on a large project.

By design, partial manifest are only available within the same directory 

### Potential improvement

/


gionn:
I like the idea of automatically loading any `_*.yaml` file in the current directory. However, a use case that just came to mind is maintaining a collection of partial configs that can be reused when creating new Updatecli pipelines. I can still make this work with the proposed implementation by using symlinks. For example, I might have a `partials/` directory containing shared `scm.yaml` files, and a separate `manifests/` directory with pipeline definitions like `pipeline1.yaml`. Inside `manifests/`, I could symlink one of the `scm.yaml` files from `partials/` to include it in a specific pipeline.

That said, I think it would offer even greater composability to support a dedicated keyword in the manifest to explicitly declare which files to load. For instance, an `imports` keyword could accept a list of optional files to include, defaulting to the `_*.yaml` pattern.

This would allow each pipeline to declare exactly which partial configs it depends on, resulting in a more DRY and modular setup.

I realize this overlaps somewhat with the Updatecli policy feature, which supports storing versioned config fragments in an OCI repository. Still, I believe this approach would be more straightforward for users, as it avoids external dependencies and works with just a Git repository.

Of course, what I'm proposing here could be moved into a separate task. The current PR could be merged as-is since it's a solid first step toward more composable updatecli pipeline definitions.

PR #5696, opened 2025-07-21
feat: support multiple yaml keys for 1 target

### LINKED ISSUE #1056: Feature Request: update multiple yaml keys with one target

I need to update two different yaml keys with the same value.  I need to have two targets to do that.

### Solution you'd like

It would be nice if the `key` field in the target yaml accept a list of keys. Therefore, there is no need to define multiple target to change multiple yaml keys with the same value.

_No response_

### Anything else?

_No response_

pkazi:
Fix #1056

## Test

```shell
cd <to_package_directory>
go test

```

### Checklist

### Potential improvement

# Note - This code is generated by GitHub Copilot

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,14 +30,14 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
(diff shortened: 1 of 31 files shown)


olblak:
Thanks for the PR.
At a first look, it looks good but I would like to spend more time to reviewing it.
Code generated by AI is usually pretty good a looking good but with wrong logic

PR #5757, opened 2025-07-30
feat: add Dasel v2 support for the JSON plugin

### LINKED ISSUE #5750: How to pick last item of a JSON array?
Hi there. I'm trying to pick the last item in this array:

https://artifacts.elastic.co/releases/stack.json

The problem is that UpdateCLI uses Dasel v1 (if I'm not mistaken), which seems to lack support for things like `releases[-1].version`. 

It seems something rather basic to me. How can I do this with UpdateCLI?

olblak:
Fix #5750 

## Test

```shell
cd pkg/plugins/resources/json/
go test
```

### Checklist

I made some comment refactoring to improve documentation based on the jsonschema

### Potential improvement

Probably worth to improve the toml plugin as well in a separated PR

### DIFF
--- a/pkg/plugins/resources/json/condition.go
+++ b/pkg/plugins/resources/json/condition.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/plugins/resources/json/condition.go
```diff
@@ -30,20 +31,33 @@ func (j *Json) Condition(source string, scm scm.ScmHandler) (pass bool, message
...
-		switch len(j.spec.Query) > 0 {
-		case true:
-			queryResults, err = j.contents[i].MultipleQuery(j.spec.Query)
-			if err != nil {
-				return false, "", err
-			}
+		switch j.engine {
+		case ENGINEDASEL_V1:
+			logrus.Debugf("Using engine %q", ENGINEDASEL_V1)
+			switch len(j.spec.Query) > 0 {
```
dduportal, line 37: nitpick: I would have kept the test on the query presence wrapping the whole switch to avoid a bug if default engine is changed in the future

dduportal (approved):
LGTM!

pkg/plugins/resources/json/condition.go
olblak: Thanks for the feedback, I hesitated following your approach but I think for now I'll proceed as is
For I'll do it differently for the toml plugin

PR #5811, opened 2025-08-01
fix: correct decoding logic for GitLab clientSpec to allow overwriting

hebestreit:
In this PR I switched the order of decoding to reference the `clientSpec` to  the action's sub spec when decoding.

I was trying to overwrite the username and token inside the GitLab Merge Request action's spec using below config but it was not passed to the `clientSpec` correctly. That's why the scm credentials were used instead.

```yaml
actions:
  default:
    kind: gitlab/mergerequest
    scmid: gitlab
    title: "my-title"
    spec:
      removesourcebranch: true
      spec:
        username: mr
        token: gplat-mr-token

scms:
  gitlab:
    kind: gitlab
    spec:
      owner: "owner"
      repository: "my-group/my-repository"
      username: gitlab-ci-token
      token: {{ requiredEnv "CI_JOB_TOKEN" }}
```

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist

### DIFF
--- a/pkg/plugins/resources/gitlab/mergerequest/main.go
+++ b/pkg/plugins/resources/gitlab/mergerequest/main.go
@@ -57,14 +57,14 @@ func New(spec interface{}, scm *gitlabscm.Gitlab) (Gitlab, error) {
 
 	// mapstructure.Decode cannot handle embedded fields
 	// hence we decode it in two steps
-	err := mapstructure.Decode(spec, &clientSpec)
+	err := mapstructure.Decode(spec, &s)
 	if err != nil {
 		return Gitlab{}, err
 	}
 
-	err = mapstructure.Decode(spec, &s)
+	err = mapstructure.Decode(s.Spec, &clientSpec)
 	if err != nil {
-		return Gitlab{}, nil
+		return Gitlab{}, err
 	}
 
 	if scm != nil {


olblak:
@hebestreit I tried your example both with your code and the latest Updatecli version and in both cases I have the same result running `updatecli manifest show --config /tmp/test.yaml`

```
name: TEST.YAML
pipelineid: 223989165bed2b42fd29f4aca958d868c934ef019a0656abed5ed4fb7c7c4f46
actions:
    default:
        title: my-title
        kind: gitlab/mergerequest
        spec:
            removesourcebranch: true
            spec:
                token: gplat-mr-token
                username: mr
        scmid: gitlab
scms:
    gitlab:
        kind: gitlab
        spec:
            owner: owner
            repository: my-group/my-repository
            token: xxx
            username: gitlab-ci-token
        disabled: false
```

pkg/plugins/resources/gitlab/mergerequest/main.go
```diff
@@ -57,14 +57,14 @@ func New(spec interface{}, scm *gitlabscm.Gitlab) (Gitlab, error) {
...
+	err := mapstructure.Decode(spec, &s)
 	if err != nil {
 		return Gitlab{}, err
 	}
 
-	err = mapstructure.Decode(spec, &s)
+	err = mapstructure.Decode(s.Spec, &clientSpec)
 	if err != nil {
-		return Gitlab{}, nil
+		return Gitlab{}, err
```
olblak: Nice catch

hebestreit:
@olblak do you mean that you were not able to reproduce it? 

I don't know if the output of the `updatecli manifest show` command is reliable, because when leaving the `actions.default.spec.spec` empty it's also not outputting the defaults from `scm`, right?

```yaml
name: TEST.YAML
pipelineid: 223989165bed2b42fd29f4aca958d868c934ef019a0656abed5ed4fb7c7c4f46
actions:
    default:
        title: my-title
        kind: gitlab/mergerequest
        spec:
            removesourcebranch: true
#            spec:
#                token: gplat-mr-token
#                username: mr
        scmid: gitlab
scms:
    gitlab:
        kind: gitlab
        spec:
            owner: owner
            repository: my-group/my-repository
            token: xxx
            username: gitlab-ci-token
        disabled: false
```


I was mainly checking the value inside the debugger where the `clientSpec` was empty with the latest Updatecli version and result into falling back to the `scm.Spec.Token`.

```go
	// mapstructure.Decode cannot handle embedded fields
	// hence we decode it in two steps
	err := mapstructure.Decode(spec, &clientSpec)
	if err != nil {
		return Gitlab{}, err
	}

	err = mapstructure.Decode(spec, &s)
	if err != nil {
		return Gitlab{}, nil
	}

	if scm != nil {

		if len(clientSpec.Token) == 0 && len(scm.Spec.Token) > 0 {
			clientSpec.Token = scm.Spec.Token
		}
	// ...
```

<details>
<summary>Before change</summary>
image
</details>

After changing the order and referencing to `s.Spec` the value got populated correctly which for me indicates it uses the correct token then.

<details>
<summary>After change</summary>
image
</details>

olblak:
Well, it was a combination of multiple issues here

The Gitlab action spec has an embedded struct for the client configuration.

In the file `pkg/core/pipeline/action/main.go` I was decoding the action spec.
I don't remember why I did it because it's misleading as we also do it in each action plugin, anyway.
But worth, it also removes the embedded client struct as the library `github.com/mitchellh/mapstructure`
doesn't support embedded struct

cfr https://github.com/go-viper/mapstructure/issues/81

By the way this current issue also highlighted that I need to migrate from `github.com/mitchellh/mapstructure`(archived) to `github.com/go-viper/mapstructure` (maintained) 

Then within the gitlab mergerequest I was decoding the wrong empty object...

olblak:
For simplicity, I am also add the fix for the other action Gitea/Bitbucket/Stash in this PR

PR #6032, opened 2025-09-11
feat: allow to squash commit

### LINKED ISSUE #4043: Feature Request: Single commit for multiple updates via multiple targets

I have a use case to update yaml file with new docker image tag. I have 3 services that  I want to update the tag. In this case, it is creating 3 commits, where as I am looking for only one commit.

### Solution you'd like

A single commit for a given SCM, it can be multiple targets.

_No response_

### Anything else?

_No response_

olblak:
Fix #4043 

Add the ability to squash all commits from a working branch.
I am introducing two new parameters:

* `squash` which by default is set to false to keep the current default behavior. But when set to true, all commits from a working branch will be squashed by Updatecli.
* `squashformat` which is used to define if Updatecli should reuse squashed commit information in the new commit.

```
scms:
    default:
        kind: gitlab
        spec:
            branch: main
            commitmessage:
                scope: deps
                squash: true
                title: bump all Golang modules
                type: chore
            email: updatecli-bot@updatecli.io
            owner: olblak
            repository: updatecli
            token: xxx
            user: olblak
            username: updatecli-bot
        disabled: false
```

```
squash defines if the commit should be squashed
	
default: false
	
important:
if squash is set to to true, then it's highly recommended to set the commit title and body
to a meaningful value as all other commit information will be lost during the squash operation.

if title and body are not set, then the commit title/message will be generated based on the most recent commit
message of the squashed commits.
```

## Test

Test manually

### Checklist

- [x] Documentation should be updated automatically once the new jsonschema is generated
### Tradeoff

While working on this issue, I made the following tradeoff

1. When squash is enabled, Updatecli squashes commits after each target, differently said, each time a commit is created. This means we never squash more than two commits at a time. This approach creates dangling commits which are garbage collected every 90 days 
3. Updatecli is not trying to be smart when creating a new squashed commit, either commitmessage is specified or we just reuse the previous git commit title. 

### Potential improvement

Commits created via the GitHub API can't be squashed at the moment as it requires a specific logic.


jvanz (commented):
Nice feature! It look fine to me.  I'll test it in the Kubewarden project to make the PR cleaner! :) 

> After additional thought, I am planning to remove the parameter SquashFormat string yaml:",omitempty"``
I don't think it's gonna be as valuable as I initially thoughts

Thinking about how I probably use in the Kubewarden project, I think that I prefer to use a written title and message in the script. Therefore, I think you are right.

PR #6202, opened 2025-10-05
feat: Add build-dependencies support to cargo autodiscovery

refi64:
Right now it handles regular and dev dependencies, but not build dependencies, which just get skipped.


olblak:
Thanks for the PR

PR #6428, opened 2025-10-22
feat(autodiscover/githubaction): handle composite actions

loispostula:
Add support for composite action to the githubaction autodiscovery

https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action
https://git.nicco.io/cupcakearmy/memoir/src/commit/64a76f137cc687333e1c9af3daa2169923fcdc9b/docs/infrastructure/github-actions/composite-actions.md
https://forgejo.org/docs/latest/user/actions/actions/

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist

### DIFF
--- a/pkg/plugins/autodiscovery/githubaction/main.go
+++ b/pkg/plugins/autodiscovery/githubaction/main.go
@@ -232,12 +232,12 @@ func (g GitHubAction) DiscoverManifests() ([][]byte, error) {
 		searchFromDir = filepath.Join(g.rootDir, g.spec.RootDir)
 	}
 
-	err := g.searchWorkflowFiles(searchFromDir, g.files)
+	err := g.searchWorkflowFiles(searchFromDir)
 	if err != nil {
 		return nil, err
 	}
 
-	err = g.searchCompositeActionFiles(searchFromDir, g.actions)
+	err = g.searchCompositeActionFiles(searchFromDir)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/plugins/autodiscovery/githubaction/main.go
```diff
@@ -218,7 +237,12 @@ func (g GitHubAction) DiscoverManifests() ([][]byte, error) {
 		return nil, err
 	}
 
-	manifests := g.discoverWorkflowManifests()
+	err = g.searchCompositeActionFiles(searchFromDir, g.actions)
```
olblak: Is there any reason to pass `g.actions` by parameter if you already have access to it via the object `g`?

pkg/plugins/autodiscovery/githubaction/main.go
```diff
@@ -107,8 +114,10 @@ type Spec struct {
 type GitHubAction struct {
 	// credentials defines the credentials to use to authenticate to the git provider
 	credentials map[string]gitProviderToken
-	// files defines the accepted GitHub Action workflow file name
+	// files defines the accepted Action workflow file name
 	files []string
+	// actions defines the accepted Composite Action name
+	actions []string
```
olblak, line 120: What's the purpose of this parameter? 
Does this means that by default it doesn't handle detected action? 

I am wondering if it would make sense to have this parameter in the matching rule instead 
https://github.com/updatecli/updatecli/blob/main/pkg/plugins/autodiscovery/githubaction/matchingRule.go

So we could use "only/ignore"

loispostula: I've followed what was done for the workflowfiles, but happy to do a more thorough pass

olblak: Ho I am sorry, I misread the code, I thought it was the configuration exposed to end user.
Nevermind forgot question

PR #6437, opened 2025-10-23
fix: switch dockerfile matching to exact match instead of prefix

### LINKED ISSUE #5954: updatecli docker autodiscovery is updating hash of `FROM` referencing previous stage

### Current Behavior

updatecli docker autodiscovery is updating hash of `FROM` referencing previous stage

Image

### Expected Behavior

The FROM referencing previous stage should be left alone

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- updatecli:
```

### Pipeline Graph

```mermaid
```

### Anything else?

_No response_

loispostula:
Fix #5954 

Switch from `HasPrefix` to `==` for Dockerfile `FROM` matching

## Test

```shell
cd pkg/plugins/resources/dockerfile
go test
```

### Checklist

### Tradeoff

This is a behavioral breaking change


olblak:
If I understand correctly, this will be a breaking change with the declarative pipeline but a bug fix for the autodiscovery plugin.
I am not sure anymore, but I think we were using the hasPRefix because at that time, Updatecli couldn't handle stage information such as.

```
FROM golang as build
```

Which we do now.

TBH, I don't understand how this change would introduce a breaking change for the autodiscovery plugin, normally it should handle every image independently. In fact, using string comparison over checking prefix is even a bug fix for the autodiscovery because a multi-stage Dockerfile containing both `python and python_runtime` wouldn't be idempotent as Updatecli could use the python tag to update the python_runtime depending on the order of the generated manifests.

To play extra safe, we could use regex over string comparison, but then we would always need to specify `$xxx^` which imho bring a bad UI 

So my take on this would be to consider this as a bug and use string comparison


To me

olblak:
Sorry for the stupid question, but why do you want to ignore stage images? I don't understand
Right now, we detect all of them. I just did a quick check and the pipelines are executed in the order of appearance which is a good think there is still the problem to update the wrong docker image tag

loispostula:
The behaviour I see is

```Dockerfile
FROM python:3.14 AS python-runtime

RUN apk add git

FROM python-runtime AS builder
...
FROM python-runtime AS dev
...
```

So we have base runtime, that we call python-runtime, it's reused multiple time.

When i run autodiscovery, I expect to only update the first one, the others are not docker images per se

olblak (approved):
Thanks for the fix

olblak:
And the dockerfile e2e tests are not passing for the autodiscovery

PR #6527, opened 2025-11-03
fix(json): Don't check for query + versionFilter with dasel/v2

refi64:
With dasel/v2, only `key` is for single *and* multiple results, so this check no longer makes sense.

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist


olblak (approved):
Thanks for the improvement

PR #6785, opened 2025-11-17
feat: add body support to HTTP source

jklap:
Using HTTP source w/POST doesn't actually use/send the passed in body


olblak:
Thanks for the PR

PR #7036, opened 2025-12-06
feat: add option to disable automerge in Gitlab

d0mitoridesu:
## Summary

When creating a Merge Request in GitLab, updatecli currently always attempts to enable auto-merge, and this behavior cannot be disabled.
This PR introduces a new option for the GitLab Merge Request resource, similar to the existing option in the GitHub Pull Request resource.
It allows users to enable/disable automatic merging for GitLab, making the behavior consistent across both integrations.

I believe that similar modules should follow the same default behavior to reduce surprises and improve user experience.

### Tradeoff

This change modifies the default behavior for GitLab by disabling auto-merge unless explicitly enabled.
Although it differs from the previous default, it now aligns GitLab behavior with GitHub and provides users more control.

### DIFF
--- a/updatecli/updatecli.d/updatecli.yaml
+++ b/updatecli/updatecli.d/updatecli.yaml
@@ -63,23 +63,3 @@ targets:
             replacepattern: 'ghcr.io/updatecli/updatecli:{{ source "latestVersion" }}'
         scmid: default
         sourceid: latestVersion
-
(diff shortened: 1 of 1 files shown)


olblak:
I wondering if we should remove this setting "merge_when_pipeline_succeeds" by default at least not expose it to Updatecli enduser

PR #7223, opened 2025-12-19
Fix usage of versionfilter regex kind in autodiscovery plugins

kevinleturc:
I'm trying to use updatecli on a maven project to update only its parent version.

The parent dependency is released in two formats, a three digits one and a two digits one.
I want updatecli to update the parent version with the two digits format, so I used the regex versionfilter kind:

```yaml
---
name: "Maven nuxeo-parent update"

autodiscovery:
  crawlers:
    maven:
      only:
        - groupIds:
            - org.nuxeo
          artifactIds:
            nuxeo-parent: ""
      versionfilter:
        kind: regex
        pattern: '^\d+.\d+$'

```

During my tests of this configuration I was not able to get an update difference and notice this error in debug logs:

```
DEBUG: parsing file "pom.xml"
DEBUG: template: manifest:23:19: executing "manifest" at <.SourceVersionFilterRegex>: can't evaluate field SourceVersionFilterRegex in type struct { ActionID string; ManifestName string; ConditionID string; ConditionGroupID string; ConditionGroupIDName string; ConditionGroupIDPath string; ConditionGroupIDValue string; ConditionArtifactID string; ConditionArtifactIDName string; ConditionArtifactIDPath string; ConditionArtifactIDValue string; SourceID string; SourceName string; SourceKind string; SourceGroupID string; SourceArtifactID string; SourceRepositories []string; SourceVersionFilterKind string; SourceVersionFilterPattern string; TargetID string; TargetName string; TargetXMLPath string; File string; ScmID string }
Manifest detected: 0
```

It happens only on 0.112.0, everything seems to work on 0.111.0.

I had a look to the maven autodiscovery plugin that causes this error and noticed that the template would render the regex field in this context:

```
      versionfilter:
        kind: '{{ .SourceVersionFilterKind }}'
        pattern: '{{ .SourceVersionFilterPattern }}'
{{- if or (eq .SourceVersionFilterKind "regex") (eq .SourceVersionFilterKind "regex/semver") (eq .SourceVersionFilterKind "regex/time") }}
        regex: '{{ .SourceVersionFilterRegex }}'
{{- end }}
```

According to the maven autodiscoverry plugin documentation https://www.updatecli.io/docs/plugins/autodiscovery/maven/, the regex field should be set when `regex/semver` and `regex/time` are used.
The current code includes the regex field with the `regex` kind, the first commit of this PR fixes that to respect the documentation for all autodiscovery plugins.

The second commit is about adding the link between the autodiscovery plugin configuration and the templating by giving versionFilterRegex parameters to template rendering.

### Potential improvement

There're default in some autodiscovery plugins that I set Regex to `*` which shouldn't be a problem for most usage as it is printed only for `regex/semver` and `regex/time` Kind. But there's might be room for improvements.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -49,7 +49,7 @@ jobs:
           install-mode: "binary"
       - uses: "pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061" # v4.2.2
         with:
-          version: "10.25.0"
+          version: "10.26.1"
           run_install: false
       - name: Install Node.js
         uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0


pkg/plugins/autodiscovery/argocd/manifestTemplate.go
```diff
@@ -18,7 +18,7 @@ sources:
       versionfilter:
         kind: '{{ .SourceVersionFilterKind }}'
         pattern: '{{ .SourceVersionFilterPattern }}'
-{{- if or (eq .SourceVersionFilterKind "regex") (eq .SourceVersionFilterKind "regex/semver") (eq .SourceVersionFilterKind "regex/time") }}
```
olblak, line 21: Is there any reason to ignore the verison filter kind "regex"

kevinleturc: According to the documentation the regex field is for regex/semver and regex/time, for regex kind the pattern field should be used.
I'm using the kind regex and set the regex on the pattern field on 0.111.0 and it works as the documentation states.

I may be wrong on my understanding, I'm New with updatecli

kevinleturc: According to the documentation the regex field is for regex/semver and regex/time, for regex kind the pattern field should be used.
I'm using the kind regex and set the regex on the pattern field on 0.111.0 and it works as the documentation states.

I may be wrong on my understanding, I'm New with updatecli.

kevinleturc: According to the documentation the regex field is for regex/semver and regex/time, for regex kind the pattern field should be used.
I'm using the kind regex and set the regex on the pattern field on 0.111.0 and it works as the documentation states.

I may be wrong on my understanding, I'm New with updatecli.

PR #7358, opened 2025-12-31
fix(terragrunt): strip double slash path separator from git URLs

eugenestarchenko:
## Problem

Terragrunt autodiscovery was failing with "repository not found" errors for all git-based module sources. The generated git URLs contained Terragrunt's `//` path separator, creating invalid repository URLs like:
```
https://github.com/seedcx/tf-aws-network//
```

## Root Cause

Terragrunt uses `//` as a special path separator for subdirectories in module sources:
```hcl
source = "git::https://github.com/org/repo//subdir?ref=v1.0.0"
```

When Updatecli parsed these sources, it included the `//` in the `baseUrl`, which was then used directly in git clone operations, causing failures.

See https://community.gruntwork.io/t/relative-paths-in-terragrunt-modules/144
## Solution

Strip the `//` path separator and everything after it from the `baseUrl` when parsing git sources. The full source (including `//subdir`) is preserved in `rawSource` and `evaluatedSource` for other uses, but the `baseUrl` now contains only the clean repository URL needed for cloning.

## Testing

Added test cases covering:
- Git URL with trailing `//`
- Git URL with `//subdir` path

All tests pass.

## Impact

Fixes repository cloning for all git-based Terragrunt modules during autodiscovery.

### DIFF
--- a/pkg/plugins/autodiscovery/terragrunt/utils.go
+++ b/pkg/plugins/autodiscovery/terragrunt/utils.go
@@ -176,7 +176,13 @@ func parseSourceUrl(evaluatedSource string, rawSource string, allowNoVersion boo
 			parts := strings.Split(fmt.Sprintf("%s:%s", u.Scheme, u.Opaque), "://")
 			if len(parts) == 2 {
 				source.protocol = parts[0]
-				source.baseUrl = parts[1]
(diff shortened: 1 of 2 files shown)


olblak (approved):
Thanks for the pull request with the tests

PR #7359, opened 2025-12-31
feat(terragrunt): add GitHub authentication support

eugenestarchenko:
## Problem

Terragrunt autodiscovery had no support for authenticating to private GitHub repositories. While other autodiscovery plugins (githubaction, etc.) support GitHub tokens, terragrunt plugin would fail when encountering private repos.

## Solution

Added generic authentication support that works with any Git provider:

### Configuration
```yaml
autodiscovery:
  crawlers:
    terragrunt:
      rootdir: "infrastructure"
      token: "ghp_xxxxxxxxxxxx"  # must be explicitly set
```

### Token Resolution
Token must be explicitly set for private repositories. No auto-detection.

Template functions can be used to read from environment:
  token: "{{ requiredEnv \"GITHUB_TOKEN\" }}"
  token: "{{ requiredEnv \"GITLAB_TOKEN\" }}"

 ### Implementation

- Uses simple Token *string field (flat, provider-agnostic)
- No external dependencies (removed github/token package)
- Injects authentication into generated SCM configurations
- Only adds password field when token is set (doesn't break public repos)

## Testing

Added TestGetToken with test coverage for:
- Nil token (no authentication)
- Empty token (no authentication)
- GitHub token (uses specific token)
- GitLab token (uses specific token)

### DIFF
--- a/pkg/plugins/autodiscovery/terragrunt/main_test.go
+++ b/pkg/plugins/autodiscovery/terragrunt/main_test.go
@@ -15,7 +15,7 @@ sources:
     name: 'Get latest version of the terraform-aws-modules/rdss/aws module'
     kind: 'terraform/registry'
     transformers:
-      - addprefix: 'tfr://terraform-aws-modules/rdss/aws?version='
+      - addprefix: 'tfr:///terraform-aws-modules/rdss/aws?version='
     spec:
       versionfilter:
         kind: 'semver'
@@ -40,7 +40,7 @@ sources:
     name: 'Get latest version of the terraform-aws-modules/aurora/aws module'
     kind: 'terraform/registry'
     transformers:
-      - addprefix: 'tfr://terraform-aws-modules/aurora/aws?version='
+      - addprefix: 'tfr:///terraform-aws-modules/aurora/aws?version='
     spec:
       versionfilter:
         kind: 'semver'
@@ -89,7 +89,7 @@ sources:
     name: 'Get latest version of the terraform-aws-modules/auroravpc/aws module'
     kind: 'terraform/registry'
     transformers:
-      - addprefix: 'tfr://${local.module}?version='
+      - addprefix: 'tfr:///${local.module}?version='
     spec:
       versionfilter:
         kind: 'semver'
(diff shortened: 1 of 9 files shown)


pkg/plugins/autodiscovery/terragrunt/spec.go
```diff
@@ -12,6 +27,17 @@ type Spec struct {
...
+	//
+	// default:
+	//   Token is automatically detected from UPDATECLI_GITHUB_TOKEN or GITHUB_TOKEN environment variables
+	//
+	// example:
+	// ```
+	//   github:
+	//     token: "ghp_xxxxxxxxxxxx"
+	// ```
+	GitHub GitHubSpec `yaml:",omitempty"`
```
olblak: I wonder the benefit of "github.token" over simply "token"
Terragrunt module could be hosted elsewhere than GitHub so your contribution could be generic.

Even if we have the parameter "token" we can still make it default to "GitHub" default environment.

pkg/plugins/autodiscovery/terragrunt/spec.go
```diff
@@ -12,6 +27,17 @@ type Spec struct {
 	Ignore MatchingRules `yaml:",omitempty"`
 	// `only` specify required rule to restrict `.terraform.lock.hcl` update.
 	Only MatchingRules `yaml:",omitempty"`
+	// `github` specifies the github credentials to use for accessing private repositories
+	//
+	// default:
+	//   Token is automatically detected from UPDATECLI_GITHUB_TOKEN or GITHUB_TOKEN environment variables
```
olblak: Another thing that I wonder is what happen if the repository is public, but we find an environment variable "UPDATECLI_GITHUB_TOKEN" used for scm authentication, which is the default behavior for Github scm. Then the terragrunt autodiscovery plugin would fail because it would try to authentication with that credential.
So I think it's better to use a pointer like 

```suggestion
	GitHub *GitHubSpec `yaml:",omitempty"`
```

And then check if pointer is nil then we assume default configuration by trying to retrieve a token from environment variable and if GitHub.Token is set to empty, then we disable authentication

eugenestarchenko:
I've refactored the implementation based on your feedback

pkg/plugins/autodiscovery/terragrunt/spec.go
eugenestarchenko: > I wonder the benefit of "github.token" over simply "token" Terragrunt module could be hosted elsewhere than GitHub so your contribution could be generic.
> 
> Even if we have the parameter "token" we can still make it default to "GitHub" default environment.

I've refactored the implementation based on your feedback

PR #7480, opened 2026-01-10
feat(replaceAll): add replaceAll method for converting different semver conventions

### LINKED ISSUE #5864: Feature Request: Allow transformation of versions before versionfilter

Some packages like [curl](https://github.com/curl/curl/releases) use semver versioning, but the github tags in the repository use underscore separated semver, for example - `curl-8_15_0`.
This makes it impossible to use versionFilter semver constraints for those use cases.
The existing transformers system runs after the version filtration thus doesn't help it this case.

For example
```
sources:
  curl:
    name: Get latest Helm release version
    kind: githubrelease
    spec:
      owner: "curl"
      repository: "curl"
      token: '{{ requiredEnv .github.token }}'
      username: olblak
      versionfilter:
        kind: regex/semver
        regex: "curl-(\d+\.\d+\.\d+)"
        pattern: "8.15.x"
```

### Solution you'd like

It would be very helpful if there was a hook between upstream version fetching and the version filtering, so we could transform the tags back to regular semver and apply the version filter afterwards.

Using release title/name as key (instead of the default tag_name) also didn't help to solve the issue as the SearchReleases function in github source doesn't search by the `key` property.
Even if it did use the key, not all packages have the regular semver in the release title.

### Anything else?

Examples of packages with underscore separated versions:
https://github.com/libexpat/libexpat/releases/tag/R_2_7_1
https://github.com/curl/curl/releases/tag/curl-8_15_0
https://github.com/postgres/postgres/tags
https://github.com/ruby/ruby/releases/tag/v3_4_5

josill:
This pull request adds a `ReplaceAll` feature to the version filter functionality, allowing users to apply regex-based transformations to version strings before filtering. This is particularly useful for normalizing version formats (e.g., converting `curl-8_15_0` to `curl-8.15.0`) before regex extraction and semantic version matching.

Closes #5864 

## Test

```
cd pkg/plugins/utils/version
go test -v
```

Or run the full test suite:

```
make test-short
```

## Additional Information

- Added `ReplaceAll` struct with `Pattern` and `Replacement` fields to the version filter
- Implemented `applyReplaceAll()` method that applies regex replacements before filtering
- Integrated `ReplaceAll` functionality into `REGEXVERSIONKIND`, `REGEXSEMVERVERSIONKIND`, and `REGEXTIMEVERSIONKIND` filter types
- Added validation for `ReplaceAll` patterns in the `Validate()` method
- Added unit tests covering various scenarios including:
  - Valid and invalid regex patterns
  - Different filter kinds (regex, regex/semver, regex/time)
  - Edge cases (empty replacement, no matches, capture groups)

### Checklist

- [x] I have added unit tests for the new functionality
- [x] I have validated that tests pass with `go test`

### Tradeoff

The `ReplaceAll` transformation is applied before regex extraction, which means:
- **Pros**: Allows normalization of version strings that don't match expected formats (e.g., underscores vs dots)
- **Cons**: Adds an extra processing step, but the performance impact is minimal as it only applies when `ReplaceAll` is configured

The validation happens at filter initialization time, catching invalid regex patterns early rather than at runtime.

### DIFF
--- a/pkg/plugins/autodiscovery/npm/dependency.go
+++ b/pkg/plugins/autodiscovery/npm/dependency.go
@@ -197,6 +197,9 @@ func (n Npm) discoverDependencyManifests() ([][]byte, error) {
 					SourceVersionFilterKind:    sourceVersionFilterKind,
 					SourceVersionFilterPattern: sourceVersionFilterPattern,
 					SourceVersionFilterRegex:   sourceVersionFilterRegex,
+					SourceNpmrcPath:            n.npmrcPath,
+					SourceURL:                  n.url,
+					SourceRegistryToken:        n.registryToken,
 					TargetID:                   "npm",
 					TargetName:                 fmt.Sprintf("Bump %q package version to {{ source \"npm\" }}", dependencyName),
 					// NPM package allows dot in package name which has a different meaning in Dasel query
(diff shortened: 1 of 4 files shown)


olblak (approved):
Thanks for the  pull request including the tests

PR #7485, opened 2026-01-11
fix: remove lowercase 'l' from compiler flags pattern to fix -latest false positive

### LINKED ISSUE #2804: Adjust check-spelling pattern
It's coming from: https://github.com/updatecli/updatecli/blob/d71451b35ef47a7e89b200f404538c5d544aa35b/.github/actions/spelling/patterns.txt#L53

You'll want to adjust that pattern (the simplest thing is to remove the `l`)
```diff
-(?:^|[\t ,>"'`=(])-[DPWXYLlf](?=[A-Z]{2,}|[A-Z][a-z]|[a-z]{2,}) 
+(?:^|[\t ,>"'`=(])-[DPWXYLf](?=[A-Z]{2,}|[A-Z][a-z]|[a-z]{2,}) 
```

It's possible for check-spelling to identify which patterns produced a fractional term, but it's somewhat expensive (it'd involve running all the patterns against the line until it identifies just the ones that contribute).

I'm not sure if `@prerelease` would improve this case (I'm currently cleaning up all of the commits I have in prerelease... hoping to have a clean test to release really soon).

_Originally posted by @jsoref in https://github.com/updatecli/updatecli/issues/2758#issuecomment-2404081162_

josill:
Fix #2804

This PR fixes a false positive in the spelling check where `-latest` was incorrectly recognized as `atest`. 

**Root Cause:**
The compiler flags pattern in `.github/actions/spelling/patterns.txt` included a lowercase `l` in the character class `[DPWXYLlf]`. When encountering `-latest`, the pattern would match `-l`, leaving `atest` to be flagged as a misspelling.

**Solution:**
Removed the lowercase `l` from the pattern, changing `-[DPWXYLlf]` to `-[DPWXYLf]`. This change:
- Prevents false positives with words like `-latest`
- Still matches legitimate compiler flags (`-D`, `-P`, `-W`, `-X`, `-Y`, `-L`, `-f`)
- The uppercase `L` remains for library path flags like `-L/path/to/lib`

**Files Changed:**
- `.github/actions/spelling/patterns.txt` (line 53)

## Test

- The spelling check will run automatically when this PR is created
- To verify manually, create a test file with:
```
echo "version: -latest" > test-spelling.txt
echo "flag: -latest-tag" >> test-spelling.txt
```

Before the fix: -latest would match -l, leaving 'atest' as false positive
After the fix: -latest is not matched by the pattern, so 'latest' is checked as a whole word (valid)**Verification:**
- Created test file with `-latest` to verify the fix
- Confirmed the pattern change is syntactically correct
- Verified that legitimate compiler flags still match correctly

**Reference:**
- Issue: #2804
- Suggested fix by @jsoref in issue comments
- Pattern documentation: https://github.com/check-spelling/check-spelling/wiki/Configuration-Examples:-patterns

### DIFF
--- a/.github/workflows/updatecli_test.yaml
+++ b/.github/workflows/updatecli_test.yaml
@@ -28,5 +28,5 @@ jobs:
         run: "updatecli compose diff"
         env:
           # This step is executed in untrusted context. We use a GitHub token with minimal permissions.
-          GITHUB_ACTOR: ${{ github.actor }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          UPDATECLI_GITHUB_USERNAME: ${{ github.actor }}
+          UPDATECLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/e2e/updatecli.d/deprecated.d/githubPullrequest.yaml
+++ b/e2e/updatecli.d/deprecated.d/githubPullrequest.yaml
@@ -8,7 +8,6 @@ scms:
     spec:
       owner: updatecli
       repository: updatecli
-      token: '{{ requiredEnv "GITHUB_TOKEN" }}'
       branch: main
 
 sources:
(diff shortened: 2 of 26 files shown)


olblak:
nice catch

olblak:
> Im not sure if the e2e tests are failing because of my changes. They seem unrelated..

Yes I recently broke the e2e tests but this PR #7507 should fix them

PR #7506, opened 2026-01-12
fix(terragrunt): add missing username for token authentication

eugenestarchenko:
## Summary
Fixes token authentication from #7359 by adding required username field.

PR #7359 added token support but authentication fails with:
  ```
  ERROR: authentication required: Repository not found.
  It appears that a password has been specified without a username, could it be related?
  ```

Root cause: go-git requires non-empty username for HTTP BasicAuth even with token-only auth.

## Solution
- Add Username field with 'oauth2' default (matches GitHub SCM plugin)
- Generated manifests now include both username and password

### DIFF
--- a/pkg/plugins/autodiscovery/terragrunt/git_test.go
+++ b/pkg/plugins/autodiscovery/terragrunt/git_test.go
@@ -55,47 +55,64 @@ func TestGetToken(t *testing.T) {
 }
 
 func TestGetUsername(t *testing.T) {
+	tokenValue := "ghp_token"
 	emptyString := ""
 	oauth2Username := "oauth2"
-	gitUsername := "git"
+	customUsername := "custom"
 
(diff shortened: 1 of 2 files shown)


andreyreshetnikov-zh (approved):

olblak:
Thanks I assumed the other pr was enough but I am happy to merge this one as well

eugenestarchenko:
> Thanks I assumed the other pr was enough but I am happy to merge this one as well

I missed that username "can be anything except an empty string" when I did PR #7359.  
https://github.com/go-git/go-git/blob/master/_examples/clone/auth/basic/access_token/main.go#L24

olblak:
Could you have a look the "github.com/updatecli/updatecli/pkg/plugins/autodiscovery/terragrunt" test is not passing

PR #7527, opened 2026-01-13
refactor: migrate away from Golang module github.com/vmware-labs/yaml-jsonpath

### LINKED ISSUE #4929: refactor: migrate away from Golang module github.com/vmware-labs/yaml-jsonpath
Looking at the various forks of that module, https://github.com/helm-unittest/yaml-jsonpath seems to be a good alternative.

So we should update our dependency

ErickdelaCruzCas:
Fix #4929 -> https://github.com/updatecli/updatecli/issues/4929

## Context / Problem

While trying to use the maintained fork of `yaml-jsonpath`, Go’s module resolution and type system exposed two separate issues:

1) **Module path mismatch at the dependency level**
- The maintained fork lives at `github.com/helm-unittest/yaml-jsonpath`, but its `go.mod` still declares:
  `module github.com/vmware-labs/yaml-jsonpath`
- If we try to `require github.com/helm-unittest/yaml-jsonpath`, Go fails with:
  `module declares its path as github.com/vmware-labs/yaml-jsonpath but was required as github.com/helm-unittest/yaml-jsonpath`

2) **YAML node type mismatch at compile time**
- The fork’s public API expects YAML nodes from `go.yaml.in/yaml/v3`.
- Our code was constructing nodes using `gopkg.in/yaml.v3`.
- Even though both packages represent “yaml.v3”, they are **different import paths**, therefore **different Go types**.
  This breaks calls like `yamlpath.Find(...)` because `*yaml.Node` from one package is not assignable to `*yaml.Node` from the other.

## Motivation

- Prefer the maintained fork to avoid depending on an unmaintained upstream, while keeping our codebase buildable and type-safe.
- Fix the module mismatch in the idiomatic Go way (without rewriting imports across the repo).
- Ensure our YAML resource passes *exactly* the node types expected by the jsonpath library to avoid fragile adapters or unsafe conversions.

### 1) go.mod — use `replace` to redirect while keeping the declared module path

- Added:
  `replace github.com/vmware-labs/yaml-jsonpath => github.com/helm-unittest/yaml-jsonpath v0.4.0`

**Why this is the correct fix**
- In Go modules, the *declared* module path in the dependency’s `go.mod` is authoritative.
- Since the fork still declares itself as `github.com/vmware-labs/yaml-jsonpath`, we must keep the original module path in our dependency graph, and use `replace` to fetch the fork’s source at build time.
- This avoids the “module declares its path as… but was required as…” error while still consuming the forked implementation.

**Result**
- We continue importing `github.com/vmware-labs/yaml-jsonpath/...` (consistent with the declared module path).
- The actual code used during builds comes from `github.com/helm-unittest/yaml-jsonpath@v0.4.0`.

### 2) pkg/plugins/resources/yaml/*.go — align YAML v3 package to match fork types

Files updated:
- `pkg/plugins/resources/yaml/condition.go`
- `pkg/plugins/resources/yaml/source.go`
- `pkg/plugins/resources/yaml/target_yamlpath.go`

Change:
- `gopkg.in/yaml.v3` → `go.yaml.in/yaml/v3`

**Why**
- `yamlpath.Find` (from the fork) expects nodes from `go.yaml.in/yaml/v3`.
- Using `gopkg.in/yaml.v3` produces identical-looking types, but Go treats them as unrelated.
- Switching our import guarantees the YAML nodes we parse/unmarshal are the ones expected by the dependency, removing the compile-time mismatch cleanly.

### 3) go.sum — dependency graph normalization

- Added checksums for the replace target (`github.com/helm-unittest/yaml-jsonpath v0.4.0`)
- Removed checksums for old/unreferenced versions (e.g., `github.com/vmware-labs/yaml-jsonpath v0.3.2`) and other pruned transitive deps.

**Why**
- `go mod tidy` updates the module graph and prunes unused entries so `go.sum` matches the effective dependency set.
- It also selected `github.com/sergi/go-diff v1.4.0` due to the fork’s transitive requirements.

## Alternatives considered

- **Requiring the fork directly (`require github.com/helm-unittest/yaml-jsonpath`)**
  - Not possible due to the module path declared in the fork’s `go.mod`.
- **Forking again just to change the `module` line**
  - Works, but adds maintenance burden and moves us away from upstream updates.
- **Writing adapters to translate between YAML node types**
  - Would be brittle and unnecessary; type-alignment is the simplest and safest fix.

## Risk / Compatibility notes

- The YAML import change is scoped to the yaml resource implementation where we interact with yaml-jsonpath; it’s not a broad refactor across unrelated packages.
- Using `replace` is explicit and documented to avoid “mystery dependency” surprises.

## How to test

- `go test ./...`
- Specifically verify yaml resources still parse YAML and jsonpath queries behave as expected.

---

*Footnote:* This is my first contribution here. If anything in the style / structure / approach should be adjusted to match project conventions, happy to iterate — gentle reviews appreciated 🙂

### DIFF
--- a/pkg/plugins/resources/yaml/target_yamlpath.go
+++ b/pkg/plugins/resources/yaml/target_yamlpath.go
@@ -54,7 +54,7 @@ func (y *Yaml) goYamlPathTarget(valueToWrite string, resultTarget *result.Target
 
 		// Process each key for this file
 		for _, key := range keys {
-			urlPath, err := yamlpath.NewPath(y.spec.Key)
+			urlPath, err := yamlpath.NewPath(key)
 			if err != nil {
 				return 0, 0, fmt.Errorf("crafting yamlpath query for key %q: %w", key, err)
 			}


olblak:
@ErickdelaCruzCas thank your for the clear explanation, I think the replace approach is the only way to as you did

Your pull request looks great as I mentioned in the previous comment but I am wondering about the motivation for this snippet:

```
		urlPath, err := yamlpath_old.NewPath(y.spec.Key)
		urlPath1, err := yamlpath_new.NewPath(y.spec.Key)
		if urlPath != urlPath1 {
			fmt.Errorf("%s different from %s", urlPath, urlPath1)
		}
```

PR #7570, opened 2026-01-19
feat: Deprecate and remove scm commit title configuration

### LINKED ISSUE #1699: Feature Request: Deprecate and remove scm commit title configuration

I don't see the value anymore of being able to define a commit title here https://github.com/updatecli/updatecli/blob/49e8280cce1b106f08c314cac21f5f5913f3641f/pkg/plugins/scms/git/commit/main.go#L67

By default the commit title will use the target name if defined which imo should be required

### Solution you'd like

Remove commit title setting and enforce resource name 

_No response_

### Anything else?

_No response_

josill:
Fix #1699

Remove the `title` field from SCM commit message configuration. Commit titles now always use the target name (or description as fallback), removing the ability to override the commit title.

**Breaking Change:** Configurations using `commitMessage.title` will need to remove that field. Ensure targets have meaningful `name` fields (which should already be the case).

## Test

```
cd pkg/plugins/scms/git/commit
go test
```

## Tradeoff
Removing the title configuration field reduces flexibility but simplifies the API and enforces consistency. Users who relied on custom commit titles will need to ensure their target names are descriptive enough for commit messages.

## Potential improvement
Consider adding validation to ensure target names are always set and meaningful, since they now directly impact commit messages.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -45,7 +45,7 @@ jobs:
           # Required: the version of golangci-lint is required
           # and must be specified without patch version:
           # we always use the latest patch version.
-          version: v2.7.2
+          version: v2.8.0
           install-mode: "binary"
       - uses: "pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061" # v4.2.2
         with:
--- a/pkg/plugins/autodiscovery/terragrunt/git_test.go
+++ b/pkg/plugins/autodiscovery/terragrunt/git_test.go
@@ -53,3 +53,68 @@ func TestGetToken(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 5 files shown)


olblak:
> Consider adding validation to ensure target names are always set and meaningful, since they now directly impact commit messages.

Already the case, the pipeline will fail if the target name is not specified
cfr https://github.com/updatecli/updatecli/blob/4027f0cc2478c1780816f0123da053ac1913665a/pkg/core/pipeline/target/main.go#L85

Something we could improve (in a different pullrequest) is to tell jsonschema to require the name on 
https://github.com/updatecli/updatecli/blob/4027f0cc2478c1780816f0123da053ac1913665a/pkg/core/pipeline/resource/main.go#L70
to provide a hint

Issue #1280, opened 2023-04-17
Feature Request: allow include file with update script steps or allow template scripts


jvanz:
### Is your feature request related to a problem?

I have an update scenario where I need to set the `semverinc` transformer depending on the value from a environment variable. The overall scenarios are:

1. If I have value X in the env var, do X version increment
1. If I have value Y in the env var, do NOT increment the version

But in both cases, the script is almost the same. The only difference is how the version is incremented. I would like to avoid duplicate the file and delete just couple of lines removing the transformer. 

### Solution you'd like

I see two possible solutions:

1. Allow include other files inside a update script. This still requires multiple files. But I can share the common steps between scripts without the need to duplicate them.
1. Allow template scripts and render the final update script on runtime.

### Alternatives you've considered

I've tried to pass an empty value to the transformer. But the tool forces me to set a valid value.

olblak:
I think I would need to have a better understanding on how the scripts work but what comes to my mind is 

**Using go template function from sprig something like**
 
-> http://masterminds.github.io/sprig/

You could have something like 
```
sources:
  default:
    kind: shell
    spec:
      command: echo {{ default  (env "WORLD") ( env "GOODBUY") }}
```

`GOODBUY="" WORLD="hello world" updatecli diff --config /tmp/updatecli.yaml`

** Having a small scripts like 

```
sources:
  script:
    kind: shell
    spec:
      command: |
        if ! [ -z "$WORLD" ]; then
          echo "$WORLD"
        else
          echo "$GOODBUY"
        fi
      environments:
        - name: "WORLD"
        - name: "GOODBUY"
```

It worth mentioning that by default environment variable are not available in the shell command.
They must be specified using the environments parameters. If the value is not defined then it inherits the value from the context where the command updatecli is executed.

PR #7591, opened 2026-01-21
feat(file): add templating file path

mallendem:
This PR adds a templating option to the file target so we can avoid messy escaping. It should fix:

https://github.com/orgs/updatecli/discussions/1243
https://github.com/updatecli/updatecli/issues/1280
https://github.com/updatecli/updatecli/issues/1329

## Test

```
cd pkg/plugins/resources/file/ && go test -v .
```

### Checklist

### Potential improvement

It allows more fine tuned templating bypassing the initial loading done by updatecli, so templating rendering is supported. Something like:

```
targets:
  useTemplate:
    kind: file
    sourceid: blablabla
    spec:
      file: version-info.txt
      template: version.tmpl
```

version.tmpl:
```
{{- $versions := .source }}
{{- range $versions }}
version installed: {{ . }}
{{- end}}
```

should work

### DIFF
--- a/e2e/updatecli.d/success.d/file_template.yaml
+++ b/e2e/updatecli.d/success.d/file_template.yaml
@@ -18,7 +18,8 @@ targets:
     kind: file
     spec:
       file: version.txt
-      template: templates/version.tmpl
+      template: e2e/updatecli.d/success.d/templates/version.tmpl
+      forcecreate: true
     sourceid: appVersion
 
   updateConfigWithSprig:
@@ -26,7 +27,8 @@ targets:
     kind: file
     spec:
       file: config.yaml
-      template: templates/config.tmpl
+      forcecreate: true
+      template: e2e/updatecli.d/success.d/templates/config.tmpl
     sourceid: appVersion
 
   templateWithTemplateData:
@@ -34,7 +36,8 @@ targets:
     kind: file
     spec:
       file: release.txt
-      template: templates/release.tmpl
+      forcecreate: true
+      template: e2e/updatecli.d/success.d/templates/release.tmpl
       templatedata:
         Environment: Production
         Date: '{{ source "date" }}'
@@ -45,10 +48,9 @@ targets:
     kind: file
     spec:
       file: info.txt
-      template: templates/info.tmpl
+      template: e2e/updatecli.d/success.d/templates/info.tmpl
       templatedata:
         ServiceName: MyService
         Version: '{{ source "appVersion" }}'
       forcecreate: true
     disablesourceinput: true
-


olblak:
Thank you for the pull request, it's an interesting approach that I never thought about.
I need to think about it

pkg/plugins/resources/file/target.go
```diff
@@ -57,6 +60,32 @@ func (f *File) Target(source string, scm scm.ScmHandler, dryRun bool, resultTarg
...
+		}
+
+		tmpl, err := template.New("file").
+			Funcs(sprig.FuncMap()).
+			Parse(templateContent)
+		if err != nil {
+			return fmt.Errorf("failed to parse template from %q: %w", f.spec.Template, err)
+		}
+
+		templateData := map[string]interface{}{
```
olblak, line 78: I am wondering if we shouldn't expose the template data as plugin parameter something like 

```
sources:
  sourceid: 
    ..
targets:
  default:
    kind: file
    spec:
      file: foo.txt
      template: foo.tpl
      templatedata: | 
        source: {{ source "sourceid" }}
``` 

This could allow us to have more than one source input.
That being said your approach could be the default behavior if `templatedata` is empty

mallendem: What about keeping the current source mandatory to keep the spec (mostly) consistent, and expose an additional data plugin parameter with the approach that you showed? 

Would that work?

I don't entirely like the mandatory source approach, but I think I dislike more the arbitrary spec. Thoughts?

mallendem: I have implemented the additional `templateData`:

```
  templateWithTemplateData:
    name: Update info.txt using Go template with .TemplateData from file
    kind: file
    spec:
      file: release.txt
      template: templates/release.tmpl
      templatedata:
        Environment: Production
        Date: '{{ source "date" }}'
    sourceid: appVersion
```

The field `sourceid` is still mandatory and can't be overriden, but you can reference other sources in `templatedata`.

olblak: > The field sourceid is still mandatory and can't be overriden, but you can reference other sources in templatedata.

Source input is the default behavior but not mandatory. Quite often we disable the source input with the parameter `disablesourceinput` like [here](https://github.com/updatecli/policies/blob/d7fdd1e461653632dfa18585c49c412fa53f82f2/updatecli/policies/nodejs/netlify/updatecli.d/manifest.yaml#L38)

In this case `.source` would be set to empty string so I think it's fine.
So something like this should work

```
  templateWithTemplateData:
    name: Update info.txt using Go template with .TemplateData from file
    kind: file
    spec:
      file: release.txt
      template: templates/release.tmpl
      templatedata:
        Environment: Production
        Date: '{{ source "date" }}'
        AppVersion: '{{ source "appVersion" }}'
    disablesourceinput: true
```

pkg/plugins/resources/file/main.go
```diff
@@ -129,6 +129,40 @@ type Spec struct {
 
 	*/
 	SearchPattern bool `yaml:",omitempty"`
+	/*
```
olblak: Could I ask you to replace the comment `/* */` with `//` like here https://github.com/updatecli/updatecli/blob/0995d5dc41c0ee92849760ea4fff5376daf965ad/pkg/plugins/resources/json/spec.go#L15

The reasoning is because those comments are used to generate the jsonschema that is then used from 
updatecli.io and your IDE to display documentation. And I noticed that IDE often replace space with \t
which lead to rendering issues like [here](https://www.updatecli.io/docs/plugins/autodiscovery/nomad/)

In the past I was using `/* */` but I am now switching to `//` each time I work on a plugin hence why we still find this old comment syntax in Updatecli codebase

mallendem: Done

pkg/plugins/resources/file/target.go
mallendem: Tested with `disablesourceinput`, it works as expected

PR #7614, opened 2026-01-26
fix: file target bug with searchpattern and matchpattern

### LINKED ISSUE #1860: Inconsistent behaviour of file target

### Current Behavior

`file target` fails(exit non-zero) if pattern isn't matched:

```yaml
targets:
  a-target:
    ...
    kind: file
    spec:
      file: 'myfile'
      matchpattern:  'mypattern*'
```

`file target` deosn't fail(exit 0) if pattern isn't matched in multiple files:

```yaml
targets:
  a-target:
    ...
    kind: file
    spec:
      file: 'myfile*'
      matchpattern:  'mypattern*'
      searchpattern: true
```

### Expected Behavior

I think updatecli should fail in unmatched cases, even though `searchpattern` is true.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Mac sonoma
- updatecli: v0.70.0
```

### Anything else?

_No response_

josill:
Fix #1860

Fix inconsistent behavior when using `searchpattern: true` with `matchpattern` where no files match the content pattern.

## Changes

- **Fixed**: When `searchpattern: true` is used with `matchpattern` and all files are filtered out (no matches), the target now returns an error instead of silently skipping
- **Consistency**: Both `searchpattern: false` and `searchpattern: true` now fail consistently when no files match the pattern

## Test

```shell
cd pkg/plugins/resources/file
go test -v
```

### Tradeoff

The change modifies behavior for users who relied on the previous silent skip when no files matched with `searchpattern: true`. However, this was inconsistent with the behavior when `searchpattern: false`, and the issue maintainer confirmed this was a bug that should be fixed.

### Potential improvement

A future enhancement could add a `matchatleast` parameter to allow specifying a minimum number of files that must match before the operation succeeds, providing more flexibility for users who need partial matching scenarios.

### DIFF
--- a/pkg/plugins/resources/file/target_test.go
+++ b/pkg/plugins/resources/file/target_test.go
@@ -2,6 +2,8 @@ package file
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


pkg/plugins/resources/file/target_test.go
```diff
@@ -425,9 +427,109 @@ func TestFile_TargetMultiples(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
+			// Create temporary files for SearchPattern tests
+			var tempDir string
+			var mockSCM scm.ScmHandler
+			if tt.spec.SearchPattern {
+				tempDir = t.TempDir()
+				// Create the files that will be matched by the pattern
+				for fileName := range tt.files {
+					filePath := filepath.Join(tempDir, fileName)
+					if err := os.WriteFile(filePath, []byte(""), 0644); err != nil {
```
olblak, line 509:
```suggestion
					if err := os.WriteFile(filePath, []byte(""), 0600); err != nil {
```

PR #7629, opened 2026-01-29
fix(file): file target bug with searchpattern and matchpattern

### LINKED ISSUE #1860: Inconsistent behaviour of file target

josill:
Fix #1860

Fix inconsistent behavior when using `searchpattern: true` with `matchpattern` where no files match the content pattern.

## Changes

- **Fixed**: When `searchpattern: true` is used with `matchpattern` and all files are filtered out (no matches), the target now returns an error instead of silently skipping
- **Consistency**: Both `searchpattern: false` and `searchpattern: true` now fail consistently when no files match the pattern

## Test

```shell
cd pkg/plugins/resources/file
go test -v
```

### Tradeoff

The change modifies behavior for users who relied on the previous silent skip when no files matched with `searchpattern: true`. However, this was inconsistent with the behavior when `searchpattern: false`, and the issue maintainer confirmed this was a bug that should be fixed.

### Potential improvement

A future enhancement could add a `matchatleast` parameter to allow specifying a minimum number of files that must match before the operation succeeds, providing more flexibility for users who need partial matching scenarios.

### DIFF
--- a/pkg/plugins/resources/file/target_test.go
+++ b/pkg/plugins/resources/file/target_test.go
@@ -760,7 +760,7 @@ func TestFile_TargetMultiples(t *testing.T) {
 			var mockSCM scm.ScmHandler
 			mockedContents := tt.mockedContents
 			wantedContents := tt.wantedContents
-			
+
 			if tt.spec.SearchPattern {
 				tempDir = t.TempDir()
 				// Create the files that will be matched by the pattern


josill:
Sorry for the confusion I closed my previous PR and opened a new one because of the unrelated changes that I needed to revert. Tried to revert them with AI but that was a bad idea and screwed up the branch history. Did this manually now and fixed the comments from the CR: #7614

olblak:
I understand could I ask you to cleanup the git history? 
For example this commit https://github.com/updatecli/updatecli/pull/7629/changes/43b49994b2cf51e88dd81e581c26f3430e5b1572 is related to the pull request

https://github.com/updatecli/updatecli/pull/7613

PR #8437, opened 2026-04-15
feat: allow Updatecli manifest dependson

### LINKED ISSUE #8436: Feature Request: Pipeline manifest ordering

Currently, there is no way to specify Updatecli manifest execution dependency because Updatecli
will read every manifest and load them as a array. As defined in the files `pkg/core/engine/main.go`
at line 21. Then for each configuration, it looks for autodiscovery configuration and any generated 
manifest are happened to the list of configuration. This prevents us to order manifest execution for example using 
file names starting with number.

### Solution you'd like

Since the configuration is an array we could sort them based on a parameter but right now we don't have such parameter.
We have the parameter `pipelineid` but it's used to group configuration from different files together.

We could introduce two parameters 
* `id` that store a unique manifest id, which is by default set to the hash of the manifest.
* `dependson` similar to the depends on the resource which always to reference id to define ordering.

I could think about a best way to handle this

### Anything else?

_No response_

olblak:
Fix #8436 

Introduce two parameters:

* `id` that store a unique manifest id.
* `dependson` similar to the `dependson` the resource which allows to reference id to define ordering.

Please note that `id` don't need to be unique so an `id` specified in an autodiscovery configuration will be herited to any generated manifest by the autodiscovery configuration.

## Test

```shell
cd pkg/core/engine
go test
```

### Checklist


olblak:
The commits from this pull request were accidentally merged from  https://github.com/updatecli/updatecli/pull/8472 using automerge

PR #2847, opened 2024-10-18
feat(scm): add bitbucket cloud scm and action support

mcwarman:
Add support for Bitbucket Cloud:

* a new scm `bitbucket`
* a new action `bitbucket/pullrequest`

Also updates documentation for stash to make it clearer its Bitbucket Server.

## Test

To test this pull request, you can run the following commands:

```shell
go build -o bin/updatecli
./bin/updatecli diff --config ./e2e/updatecli.d/success.d/bitbucket/scm.yaml
```

### Tradeoff

Bitbucket doesn't support any HTML in PR descriptions, so I had to add Markdown only views. There's a potential the format could be improved going forward

### Potential improvement

Markdown format of PRs.

Issue #1905, opened 2024-01-31
Action Failed: Branches do not exist


kemra102:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We have a GitLab CI job that runs to update the Helm deps of our Helm based API deployments. Each API is a GitLab repo with a branch per environment (prefixed **env/**, e.g. **env/test**). The CI job iterates for each branch calling `updatecli` for each branch (each iteration has it's own config file due to the different branch names).

On a random branch in the iteration we get an error like this one:

```sh
ACTIONS
========
Pipeline "Bump dependency \"simple-api\" for Helm chart \"example-app\"" failed
Skipping due to:
	action stage:	"remote branches \"updatecli_02dd09819960503297b29201e46ffd28df93fe95b2e1cea5c31810cdcc87445b\" and \"env/test\" do not exist, we can't open a mergerequest"
```

However both branches mentioned in the error DO exist. The error causes the pipeline to correctly fail.

### Expected Behavior

MRs are created for each **env/** branch.

### Environment

```markdown
- OS: RHEL8
- updatecli: v0.71.0
```

### Anything else?

[example-updatecli-config.txt](https://github.com/updatecli/updatecli/files/14110994/example-updatecli-config.txt)

olblak:
Hi @kemra102,
Is it still something you are experiencing?
We made quite a lot of improvement since Updatecli version v0.71.0
We are about to release v0.114.0

kemra102:
@olblak I think the team I implemented this for has since moved onto to other solutions (for reasons besides this error - I think they used `updatecli` for some time).

Given my current circumstances I don't have easy access to the original code so it may be worth assuming this is either fixed unless someone else reports otherwise.

chrxmvtik:
@olblak can we reopen this issue? This error still persist for us.

Updatecli version: v0.118.0
Gitlab selfhosted version: v18.6.0

It happens randomly, for example:

First job in the pipeline works fine, it creates changes and publish MR to repo X, second job is triggered right after first one finish - it is also supposed to create changes and publish MR to repo X, but it fails, and says that branch <updatecli_temporary_branch> and "main" does not exist.

When I set a job to retry few times, it finally succeeds, but it shouldn't work this way.

When I downgraded to version v0.100.0, it works fine.

olblak:
> When I downgraded to version v0.100.0, it works fine.
That's a good hint

PR #8584, opened 2026-04-24
fix(bitbucket): use FindBranch and paginate PR listing in pullrequest…

roulettedares:
---
  Problem

  isRemoteBranchesExist() and isPullRequestExist() both fetch only the first page of results (Page: 1, Size: 30) with no follow-through pagination. On repositories with more than 30 branches, if either the source branch or target branch (main)
  sorts alphabetically past position 30, the function silently returns false — causing CreateAction to log "skipping pull request creation" at Debug level and exit with nil. The job appears green with no PR created and no visible explanation in
  normal log output.

  Root cause / history

  The Stash plugin (PR #1246, March 2023) used ListBranches with a manual name-match loop. The Bitbucket Cloud plugin (PR #2847, October 2024) copy-pasted this approach verbatim; no reviewer questioned it. FindBranch has been present in both
  go-scm Bitbucket drivers since 2018 and was never unavailable. Branch names are never wildcards or patterns — there was no reason to list all branches. PR #5133 silently switched to FindBranch elsewhere in a refactor, but the Bitbucket Cloud
  plugin was not updated.

  Fix

  Replace the ListBranches scan in isRemoteBranchesExist() with two direct FindBranch calls — one per branch. This is O(1) regardless of branch count, matches the approach used by GitHub's GraphQL ref(qualifiedName:) lookup, and treats a 404
  response as "branch not found" logged at Warning level so the failure is visible without --debug.

  Also fix isPullRequestExist() which has the same Size: 30 truncation: it now paginates using resp.Page.Next until all open pull requests have been checked.

  Relates to #1905 (same class of bug observed in GitLab). Introduced in #2847.

## Test

```shell
cd <to_package_directory>
go test
```

### Checklist

### DIFF
--- a/pkg/plugins/resources/bitbucket/pullrequest/utils.go
+++ b/pkg/plugins/resources/bitbucket/pullrequest/utils.go
@@ -8,7 +8,6 @@ import (
 	"github.com/drone/go-scm/scm"
 	"github.com/sirupsen/logrus"
 	"github.com/updatecli/updatecli/pkg/core/result"
-	"github.com/updatecli/updatecli/pkg/plugins/resources/bitbucket/client"
 )
 
 type pullRequestDetails struct {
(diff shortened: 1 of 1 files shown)


olblak:
Thanks for the fix

olblak:
And thank you for the clear explanation.

olblak:
@roulettedares Could you confirm that you tested this pullrequest on a bitbucket repository?

roulettedares:
it's been consistently working on repos with more than 30 branches

olblak:
Thank you awesome, then I'll skip the manual testing this time

Issue #8951, opened 2026-05-20
Feature Request: Autodiscovery for Podman Quadlets


ChristianCiach:
### Is your feature request related to a problem?

Podman Quadlets are similar to Docker Compose service definitions, but they are using a systemd unit like syntax. Quadlets are usually placed in `/etc/containers/systemd/` and the simpliest kind of a Quadlet uses the `.container` file suffix, which I would like for updatecli to support.

Example:

**/etc/containers/systemd/wg-easy.container**

```ini
[Unit]
Description=A simple wireguard management server

[Install]
WantedBy=default.target

[Container]
Image=ghcr.io/wg-easy/wg-easy:15.3.0
Environment=INSECURE=true
Volume=/lib/modules:/lib/modules:ro
Volume=/etc/wg-easy:/etc/wireguard:rw
PublishPort=51820:51820/udp
PublishPort=51821:51821/tcp
```

### Solution you'd like

I would like to just execute `updatecli apply` inside the `/etc/containers/systemd/` directory and let updatecli do its thing :) Images should be automatically updated and pinned using digests.

olblak:
This is a good idea, we could parse systemd configuration file using https://github.com/coreos/go-systemd/tree/main

We would need a few plugin https://github.com/coreos/go-systemd/tree/main

ScMofeoluwa:
I'd like to work on this. For the first PR, should I limit the scope to `.container` files and autodiscovery of `Image=` values using go-systemd, or do you expect a dedicated Quadlet resource plugin too?

ChristianCiach:
I don't even know what a quadlet resource plugin would do. Sorry, I never used the full feature set of updatecli. I usually just `updatecli apply` my docker compose files and I am happy that it works 😅

There doesn't seem to be a dedicated "docker compose" resource plugin, probably because you can just use the "yaml" resource plugin. In that sense, there should probably (if at all) a systemd-unit resource plugin instead of a specialized quadlet plugin. But this is out of scope of my feature request.

ScMofeoluwa:
Thanks, that matches my understanding. I'll wait for @olblak confirmation, but autodiscovery only for .container files sounds like the right first scope

olblak:
Just to clarify here, an autodiscovery plugin is nothing more than a generator of Updatecli manifest.
For example here the docker compose autodiscovery plugin: https://github.com/updatecli/updatecli/blob/main/pkg/plugins/autodiscovery/dockercompose/manifestTemplate.go

[...]

PR #8970, opened 2026-05-28
feat(systemd): add resource plugin for systemd unit files

ScMofeoluwa:
Adds a systemd resource plugin for reading and updating options in systemd unit files using go-systemd. Defaults to Container/Image for Quadlet compat.

## Test
```shell
go test ./pkg/plugins/resources/systemd/
```

Checklist
- Tests pass
- Documentation update needed (will do with autodiscovery plugin)

Refs #8951


olblak:
I tested the pullrequest with the following example, 

```
sources:
  default:
    kind: systemd
    spec:
      file: "/tmp/wg.container"
      section: "Container"
      option: Image

conditions:
  default:
    kind: systemd
    spec:
      file: "/tmp/wg.container"
      section: "Container"
      option: Image

targets:
  default:
    kind: systemd
    disablesourceinput: true
    spec:
      file: "/tmp/wg.container"
      section: "Container"
      option: Volume
      value: "xxx"
```

where /tmp/wg.container is the file mentioned in the issue.
One of things that I am wondering is how to handle multiple option like `Volume` or `PublishPort`


Another question that I am wondering in the case of the example shared in the linked issue.
Because right now it will update the first option it find

ScMofeoluwa:
> I tested the pullrequest with the following example,
> 
> ```
> sources:
>   default:
>     kind: systemd
>     spec:
>       file: "/tmp/wg.container"
>       section: "Container"
>       option: Image
> 
> conditions:
>   default:
>     kind: systemd
>     spec:
>       file: "/tmp/wg.container"
>       section: "Container"
>       option: Image
> 
> targets:
>   default:
>     kind: systemd
>     disablesourceinput: true
>     spec:
>       file: "/tmp/wg.container"
>       section: "Container"
>       option: Volume
>       value: "xxx"
> ```
> 
> where /tmp/wg.container is the file mentioned in the issue. One of things that I am wondering is how to handle multiple option like `Volume` or `PublishPort`
> 
> Another question that I am wondering in the case of the example shared in the linked issue. Because right now it will update the first option it find


Hmm, good point. I think a safer approach might be: update normally when there is only one match, but return an error when multiple matching options are found unless the user provides an explicit selector like `index` (would need to add that). If needed later, we could also add an explicit `all: true`, but I don’t think that should be the default

ScMofeoluwa:
> > I tested the pullrequest with the following example,
> > ```
> > sources:
> >   default:
> >     kind: systemd
> >     spec:
> >       file: "/tmp/wg.container"
> >       section: "Container"
> >       option: Image
> > 
> > conditions:
> >   default:
> >     kind: systemd
> >     spec:
> >       file: "/tmp/wg.container"
> >       section: "Container"
> >       option: Image
> > 
> > targets:
> >   default:
> >     kind: systemd
> >     disablesourceinput: true
> >     spec:
> >       file: "/tmp/wg.container"
> >       section: "Container"
> >       option: Volume
> >       value: "xxx"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where /tmp/wg.container is the file mentioned in the issue. One of things that I am wondering is how to handle multiple option like `Volume` or `PublishPort`
> > Another question that I am wondering in the case of the example shared in the linked issue. Because right now it will update the first option it find
> 
> Hmm, good point. I think a safer approach might be: update normally when there is only one match, but return an error when multiple matching options are found unless the user provides an explicit selector like `index` (would need to add that). If needed later, we could also add an explicit `all: true`, but I don’t think that should be the default

What do you think about this @olblak

olblak:
sorry I missed the ball, to align the behavior with other plugins like the [YAML](https://www.updatecli.io/docs/plugins/resource/yaml/) one
I think the default should be to update only the first occurence unless an index value is defined

cfr documentindex in the yaml documentation

olblak:
@ScMofeoluwa Do you need some help to finalize this pullrequest?

ScMofeoluwa:
> @ScMofeoluwa Do you need some help to finalize this pullrequest?

Sorry, this got buried in my mail. Thanks for the clarification. I’ll update the resource to follow that behavior and push an update shortly.

ScMofeoluwa:
@olblak updated this now. I also addressed copilot's review but I left the target write path using `go-systemd` serialization for now rather than adding custom line/regex handling just to preserve comments, since that felt more fragile. Happy to adjust if you’d prefer a different approach

olblak:
While doing some manual testing, I spotted a few issues that I fixed in my last commit.
On top of this I also implemented the default index which update all options if unset

olblak:
I also forgot to specify in my comment, I removed the default value for section, and option as I think this plugin should be generic to any systemd configuration

ScMofeoluwa:
> I also forgot to specify in my comment, I removed the default value for section, and option as I think this plugin should be generic to any systemd configuration

Thanks for the fixes. The explicit `section` / `option` requirement makes sense for keeping this generic

PR #9064, opened 2026-06-02
fix: skip version check for generator commands

### LINKED ISSUE #9028: Version-check banner is written to stdout in PersistentPostRun, corrupting generated shell completions
## Summary

The "new release available" version-check banner is emitted via `logrus.Infof` to **stdout**, and the check runs in `rootCmd.PersistentPostRun`, which executes after **every** subcommand — including `completion`. As a result, when shell completions are generated (e.g. `updatecli completion bash`), the banner is captured into the completion script itself, producing an invalid file that breaks the user's shell on startup.

This is especially impactful for the Homebrew formula, which generates completions at install time. End users get a broken bash completion file and a syntax error on every new shell.

## Version

- `updatecli` v0.117.1 (latest release at time of writing)
- Reproduced via the Homebrew-generated completion; the relevant code on `main` is unchanged.

## Steps to reproduce

```console
$ updatecli completion bash > updatecli.bash
$ tail -n 3 updatecli.bash
# ex: ts=4 sw=4 et filetype=sh
| A new release is available: "unknown" -> "v0.117.0"
| More information on https://www.updatecli.io/changelogs/updatecli/changelogs/v0.117.0/
$ bash -n updatecli.bash
updatecli.bash: line 427: syntax error near unexpected token `|'
```

When this file is installed as a bash completion (e.g. under `/opt/homebrew/etc/bash_completion.d/`), every new shell prints:

```
bash: /opt/homebrew/etc/bash_completion.d/updatecli: line 427: syntax error near unexpected token `|'
bash: /opt/homebrew/etc/bash_completion.d/updatecli: line 427: `| A new release is available: "unknown" -> "v0.117.0"'
```

## Root cause

1. `cmd/root.go` calls `logrus.SetOutput(os.Stdout)`, so info-level logs (including the version banner) go to **stdout** rather than stderr.
2. The version check is wired into `rootCmd.PersistentPostRun`, so it runs for **all** subcommands, including `completion` / `__complete` / `docs` / `man`:
   ```go
   rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
       err := engine.CheckLatestPublishedVersion()
       ...
   }
   ```
3. `CheckLatestPublishedVersion()` in `pkg/core/engine/version.go` emits the banner with `logrus.Infof("| A new release is available...")`, which therefore lands on stdout and is captured by any command that redirects stdout (such as completion/doc generation).

## Suggested fixes (any one, ideally a combination)

- **Route the banner to stderr.** Informational/diagnostic output should not pollute stdout, which is reserved for the command's actual output (the completion script, generated docs, etc.).
- **Skip the version check for non-interactive/generator commands** in `PersistentPostRun` — e.g. `completion`, `__complete`, `__completeNoDesc`, `docs`, `man`, `jsonschema`.
- **Add an opt-out env var / flag**, e.g. `UPDATECLI_DISABLE_VERSION_CHECK` (mirroring the existing `UPDATECLI_DISABLE_CHANGELOG`), so it can be disabled in CI and by packagers.

## Impact / workarounds

- Affects anyone generating completions or docs from a release binary; Homebrew users hit it automatically.
- Current workaround is to strip the banner from generated output, e.g. `updatecli completion bash | sed '/^| /d'`, or manually delete the trailing `| ...` lines from the installed completion file. This does not survive package upgrades.

Happy to open a PR if a preferred direction is confirmed.

ambikeesshh:
Fix #9028

## Summary

The version check in `PersistentPostRun` runs for every subcommand, including `completion`. Since `logrus` output goes to stdout, the "new release available" banner gets captured into generated shell completion scripts, producing an invalid file that breaks the user shell on startup.

This hits Homebrew users especially hard since completions are generated at install time, so every new shell gets a syntax error.

## What changed

- Added `--disable-version-check` flag and `UPDATECLI_DISABLE_VERSION_CHECK` env var (same pattern as `--disable-changelog`)
- Skip version check for generator commands: `completion`, `__complete`, `__completeNoDesc`, `docs`, `man`, `jsonschema`
- Walk up the command tree with `cmd.Parent()` so nested subcommands like `completion bash` are also correctly skipped
- Added tests for flag registration, env var defaults, flag override, and skip list contents

## Test

```shell
go test ./cmd/...
go build -o /tmp/updatecli . && /tmp/updatecli completion bash > test.bash && bash -n test.bash
```

Note: `bash -n` only catches the issue if a newer version of updatecli is available (which is what triggers the banner). If you are already on the latest version, no banner is emitted so the test passes regardless.

### Checklist

### Tradeoff

- The version check is skipped entirely for generator commands rather than routing the banner to stderr. This avoids the corruption but means users will not see version notifications when running these commands.
- New generator commands added in the future need to be added to the skip list.
- The root cause (`logrus.SetOutput(os.Stdout)` sending diagnostic output to stdout) is not addressed here. A follow-up could route the banner to stderr as a more general fix.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -10,6 +10,8 @@ import (
 
 const DisableChangelogEnvVar = "UPDATECLI_DISABLE_CHANGELOG"
 
+const DisableVersionCheckEnvVar = "UPDATECLI_DISABLE_VERSION_CHECK"
+
 // getEnvBoolOrDefault reads a boolean environment variable.
 // It returns defaultValue when the variable is unset or invalid.
 func getEnvBoolOrDefault(envVar string, defaultValue bool) bool {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,9 +87,10 @@ func init() {
 		if disableVersionCheck {
 			return
 		}
-		cmdName := cmd.Name()
-		if slices.Contains(skipVersionCheckCommands, cmdName) {
-			return
+		for c := cmd; c != nil; c = c.Parent() {
+			if slices.Contains(skipVersionCheckCommands, c.Name()) {
+				return
+			}
 		}
 		err := engine.CheckLatestPublishedVersion()
 		if err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -27,20 +27,20 @@ import (
 )
 
 var (
-	pipelineIds      []string
-	labels           []string
-	manifestFiles    []string
-	valuesFiles      []string
-	valuesInline     []string
-	secretsFiles     []string
-	policyReferences []string
-	e                engine.Engine
-	verbose          bool
-	experimental     bool
-	disableTLS       bool
-	disableChangelog     bool
-	uniqueTmpDir         bool
-	disableVersionCheck  bool
+	pipelineIds         []string
+	labels              []string
+	manifestFiles       []string
+	valuesFiles         []string
+	valuesInline        []string
+	secretsFiles        []string
+	policyReferences    []string
+	e                   engine.Engine
+	verbose             bool
+	experimental        bool
+	disableTLS          bool
+	disableChangelog    bool
+	uniqueTmpDir        bool
+	disableVersionCheck bool
 
 	rootCmd = &cobra.Command{
 		Use:   "updatecli",
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -38,8 +38,9 @@ var (
 	verbose          bool
 	experimental     bool
 	disableTLS       bool
-	disableChangelog bool
-	uniqueTmpDir     bool
+	disableChangelog     bool
+	uniqueTmpDir         bool
+	disableVersionCheck  bool
 
 	rootCmd = &cobra.Command{
 		Use:   "updatecli",
@@ -66,11 +67,30 @@ func Execute() {
 	}
 }
 
+// skipVersionCheckCommands lists commands where the version check is skipped
+// to avoid unnecessary network requests and prevent banner output from
+// corrupting generated content.
+var skipVersionCheckCommands = []string{
+	"completion",
+	"__complete",
+	"__completeNoDesc",
+	"docs",
+	"man",
+	"jsonschema",
+}
+
 func init() {
 
 	logrus.SetOutput(os.Stdout)
 
 	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
+		if disableVersionCheck {
+			return
+		}
+		cmdName := cmd.Name()
+		if slices.Contains(skipVersionCheckCommands, cmdName) {
+			return
+		}
 		err := engine.CheckLatestPublishedVersion()
 		if err != nil {
 			logrus.Debugf("Unable to check for the latest version of updatecli: %v", err)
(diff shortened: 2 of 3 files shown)


ambikeesshh:
@olblak, PTAL

This PR fixes the immediate problem (banner corrupting completion scripts) by skipping the version check for generator commands and adding the --disable-version-check flag. But the root cause from #9028 is still open: logrus.SetOutput(os.Stdout) means the banner goes to stdout, so any command where someone pipes stdout could still get the banner mixed into their output

The issue suggested routing the banner to stderr as one of the fixes. I held off on that here to keep the scope small, but happy to include it in this PR if you'd prefer. Otherwise I can open a follow-up.

thanks!!

olblak:
Thanks for the pullrequest, I like the environment variable and flag to totally disable the version checking as it can reduce http queries to updatecli.io

I am struggling to test this change manually or other changes.

I am curious what you think about my suggestion on https://github.com/updatecli/updatecli/issues/9028#issuecomment-4593008643.
The idea would to only run the check version in a defer function so in theory it wouldn't be executed on auto completation.

Regarding moving the message to stderr, I don't know if it's the time to do it as I would like to revamp to UI and move away from logrus completely, but I never took the time

ambikeesshh:
Thanks for the feedback @olblak!

Regarding the `defer` approach in `run()` - I think that's cleaner than the skip list. The skip list works but every new generator command needs to be manually added. With `defer` inside `run()`, the version check naturally only runs for commands that go through that function, so completion/docs/man/etc. are excluded without any extra logic.

The tradeoff is that some commands in `run()` like `udash/login`, `udash/logout`, `manifest/pull` probably don't need the version check either, but since they don't redirect stdout it's not harmful, just an extra network request. The skip list is more surgical, but `defer` is simpler and harder to get wrong.

I'd lean toward `defer` for simplicity, and keep the `--disable-version-check` flag either way so users can opt out regardless of which command they're running. That also helps with reducing HTTP queries to updatecli.io.

As for only checking on `updatecli version` - that feels a bit too restrictive since most users don't run that directly, so they'd never see the update notification. The current approach surfaces it during normal usage which is more useful.

Happy to refactor to the `defer` approach if that's your preference. And understood on stderr/logrus, makes sense to handle that as part of a larger UI revamp.

olblak:
I could manually test it running 

```
go build -ldflags="-X github.com/updatecli/updatecli/pkg/core/version.Version=v0.1.0" -o bin/updatecli .
./bin/updatecli completion bash > /tmp/completions.bash
grep -n 'new release' /tmp/completions.bash
bash -norc /tmp/completions.bash
```

Where we shouldn't see `427:| A new release is available: "0.1.0" -> "0.117.1"`

olblak:
After additional testing, your approach is better, as not every subcommand calls the run function
So, for example, my suggestion wouldn't work with `updatecli version`

ambikeesshh:
Good catch on `updatecli version` @olblak, that would have been an ironic miss. Thanks for fixing the formatting too. Skip list it is, I'll keep an eye on CI.

ambikeesshh:
looks like the analyse check hit a gitHub rate limit. mind re-running it when you get a chance @olblak?
thanks!

ambikeesshh:
all checks passed @olblak. ready for merge whenever you are :)

olblak:
Thank you for the pullrequest today I learn

PR #9095, opened 2026-06-02
fix: return error when {{ source }} references a skipped source

ambikeesshh:
Fix #7919

## Summary

`{{ source "id" }}` returns the literal string `{{ source "id" }}` when the referenced source is skipped via `depends_on`. Since a skipped source never transitions to SUCCESS, this template literal never resolves. It silently leaks into config values causing confusing downstream failures with no clear indication of what went wrong.

## What changed

Sources initialized with `Result: SKIPPED`, same state as an explicitly skipped source, so the template function couldn't tell them apart. Two changes fix this:

- Changed initial source result from `result.SKIPPED` to empty string, so "not yet run" and "skipped" are distinguishable
- Set source result to SKIPPED explicitly when `shouldSkipResource` skips a source

Then the template function can safely error on SKIPPED:

- Added `case result.SKIPPED` in the runtime `source` template function that returns a clear error, consistent with existing `FAILURE` handling

## Test

```shell
go test ./pkg/core/config/... -run TestUpdate -v -count=1
go test ./pkg/core/pipeline/... -run TestRunSources -v -count=1
```

New test cases:
- `pkg/core/config/main_test.go` - unit test for SKIPPED source in template function
- `pkg/core/pipeline/sources_test.go` - end-to-end test with a source skipped via `depends_on` and a target referencing it via `{{ source "..." }}`


ambikeesshh:
hey @olblak PTAL
thanks!!

ambikeesshh:
@olblak can you review it ? when you get a chance

olblak:
Of course, I'll try to find some time thanks for the fix

pkg/core/pipeline/main.go
```diff
@@ -144,7 +144,7 @@ func (p *Pipeline) Init(config *config.Config, options Options) error {
 		p.Sources[id] = source.Source{
 			Config: config.Spec.Sources[id],
 			Result: &result.Source{
-				Result: result.SKIPPED,
+				Result: "",
```
olblak: Why is it changed from result.SKIPPED to empty string?

olblak: Ok, now I understand.

ambikeesshh: empty string = not yet run (retry later). SKIPPED = explicitly skipped (error)
both were the same value before this change

ambikeesshh:
@olblak rebased on latest main

ambikeesshh:
@olblak friendly ping

ambikeesshh:
@olblak

olblak:
@ambikeesshh Sorry very busy days at work.
Have you seen my review?

ambikeesshh:
@olblak yep, i replied inline on the `Result: ""` change. The empty string distinguishes "not yet run" from "explicitly skipped" so the template function can error on the latter. 

lemme know if you'd like any other changes

ambikeesshh:
@olblak

olblak:
> @olblak

Hi @ambikeesshh,
I am reposting here what I post earlier in this pullrequest

---

I am a bit annoyed by this change because if something unexpected happened, the resource wouldn't have a clear state.
It's used, for example, on this webapp https://app.uda.sh/updatecli/scm/dashboard to monitor Updatecli dependencies.

I think I would prefer to add a new Boolean named "IsRun" on the source/condition/target like
https://github.com/updatecli/updatecli/blob/094d96c892877b56a09aecccb34be2edfc5a620e/pkg/core/pipeline/source/main.go#L22

Once The source function is exectued like here https://github.com/updatecli/updatecli/blob/094d96c892877b56a09aecccb34be2edfc5a620e/pkg/core/pipeline/source/main.go#L112
then we set  it to true

ambikeesshh:
hey @olblak 
done, reworked it the way you suggested

- dropped the empty `Result`, sources now init to `SKIPPED` like conditions/targets so the state stays valid
- added `IsRun` to source/condition/target, set to true inside each `Run()` where you pointed

one note: a skipped source never hits `Run()`, so i set `IsRun = true` on the skip path too. so it means "ran or skipped". if you'd rather it strictly mean "executed", i'll use a separate flag instead. let me know

thanks for the review!

olblak:
thanks for the update, I'll review this pr in the coming days

olblak:
sorry for the confusion on this pullrquest, I now realize that all the comments that I made on this pr where only visible to me because they were "pending"

cfr https://github.com/orgs/community/discussions/10369

ambikeesshh:
@olblak 
haha, no worries. thanks for the review tho :)

olblak (approved):
I manually tested using the following manifest 

```
name: Test sourceid {{ source "shell" }}

sources:
  shell:
    name: Show 1.2.3
    kind: shell
    dependson:
      - condition#default
    spec:
      command: 'echo "1.2.3"'

  shell2:
    name: Show source output {{ source "shell" }}
    kind: shell
    spec:
      command: 'echo {{ source "shell" }}'

conditions:
  default:
    name: "False"
    kind: shell
    disablesourceinput: true
    spec:
      command: |
        false
```

And it works as expected.
While testing it,  I spotted a minor issue which I'll fix in a follow up pr

ambikeesshh:
@olblak 
the e2e (`Run End to End tests`) is red on the latest run, and it's from this change.

turns out a bunch of the repo's own `deps: bump` manifests hit the new behavior: a source gets skipped via `dependson` (e.g. `actions/checkout` when release `v4` isn't found → `source#branch` skipped), but the targets still reference `{{ source "branch" }}`. on main that rendered as a literal silently, now it errors with "parent source ... was skipped" so ~45 pipelines fail and the e2e assertion trips.

is that covered by the follow-up you mentioned, or would you rather skipped sources also skip their dependents instead of erroring? happy to go either way, just lmk which you prefer

olblak:
This is not related to my follow up.
Mine is about initializing report data.

I think it would make more sense to not error on skipped source but just to skip dependents resources

PR #9266, opened 2026-06-25
fix(pyproject): track uv.lock relative to target workdir

### LINKED ISSUE #9265: pyproject autodiscovery: uv.lock path is doubled for projects in subdirectories, target fails with "Missing files"

### Current Behavior

When the pyproject autodiscovery crawler (`autodiscovery.crawlers.pyproject`, or the policy `ghcr.io/updatecli/policies/autodiscovery/pyproject`) discovers a `pyproject.toml` in a **subdirectory** (e.g. `backend/`), the generated `shell` target fails even though the underlying `uv lock` command succeeds.

The generated target sets:

- `workdir: backend` — the `uv` command runs here and succeeds, and
- a `changedif` (`file/checksum`) tracking `backend/uv.lock`.

But the shell resource resolves the tracked file **relative to `workdir`**, producing `backend/backend/uv.lock`. That doubled path does not exist, so the post-command file check reports `Missing files` and the target is marked failed.

When the project is at the scan root, `workdir` is `.` and the tracked path is `uv.lock`, so nothing doubles and the target passes — which is why this is only seen for projects in subfolders.

`--debug` output for one target (`Faker`):

```
target: Faker
DEBUG: open backend/backend/uv.lock: no such file or directory
DEBUG:     command: /bin/sh /tmp/updatecli/bin/<hash>.sh
The shell 🐚 command "/bin/sh /tmp/updatecli/bin/<hash>.sh" ran successfully with the following output:
DEBUG: command stderr output was:
Using CPython 3.13.13 interpreter at: /.../python3.13
Resolved 174 packages in 788ms          <-- uv succeeded
DEBUG: open backend/backend/uv.lock: no such file or directory
DEBUG: no checksum for file "backend/uv.lock" before running shell command
DEBUG: no checksum for file "backend/uv.lock" after running shell command
DEBUG: Missing files "backend/uv.lock"
    Shell command exited on error.
```

Note the contradiction: the command `ran successfully` (uv resolved 174 packages), yet the target fails on the missing `backend/backend/uv.lock`.

This looks like a core-crawler bug, not a policy bug: it reproduces identically with a hand-written manifest and with the published policy.

### Expected Behavior

The generated `shell` target should track the lockfile at a path relative to its own `workdir`, so that for a project at `backend/` running with `workdir: backend` the `changedif` file is simply `uv.lock` (resolving to `backend/uv.lock`), and the target succeeds when `uv lock` updates it.

### Steps To Reproduce

1. Create a repo with a Python project in a subdirectory, e.g. `backend/pyproject.toml` plus a generated `backend/uv.lock`.
2. Run pyproject autodiscovery against the repo root, either via a hand-written manifest:

   ```yaml
   autodiscovery:
     crawlers:
       pyproject: {}
   ```

   or via the policy `ghcr.io/updatecli/policies/autodiscovery/pyproject:0.1.0`.
3. Run `updatecli diff --config <manifest> --debug` (or `apply`).
4. Observe the `Faker`/any target fail with `DEBUG: Missing files "backend/backend/uv.lock"` despite the `uv lock` command reporting success.

### Environment

```markdown
- OS: Ubuntu 24.04 arm64 (GitHub-hosted runner); also reproduced on Linux x86_64
- updatecli: v0.118.0 (also reproduced on v0.117.0)
- uv: 0.11.21 (also 0.11.9)
- Source: autodiscovery.crawlers.pyproject, and policy
  ghcr.io/updatecli/policies/autodiscovery/pyproject:0.1.0
  (@sha256:115651e7596b3ad3c00d30163ff3475380a15070c2899c7591185a8ea2ed6e2b)
```

### Anything else?

The root cause appears to be in the generated manifest: the `workdir` is applied **twice** to the lockfile path. The crawler computes the tracked lockfile as `filepath.Join(workdir, "uv.lock")` and emits the same `workdir` into the target, but the shell resource already resolves `changedif` files relative to `workdir`.

Relevant code:

- `pkg/plugins/autodiscovery/pyproject/dependencies.go:201` — `relLockFile := filepath.Join(workdir, "uv.lock")`
- `pkg/plugins/autodiscovery/pyproject/manifestTemplate.go:40,43` — emits both the `changedif` `files:` entry (`{{ .LockFile }}`) and `workdir: '{{ .Workdir }}'`.

A likely fix is to track the lockfile as `uv.lock` (relative to the target's own `workdir`) rather than joining `workdir` in. The existing tests (`pkg/plugins/autodiscovery/pyproject/main_test.go`) only exercise the scan-root case (`workdir: '.'`, `files: ["uv.lock"]`), so a subdirectory test case would catch this regression. Happy to open a PR if useful.

nguyeda:
## Current Behavior

The pyproject autodiscovery crawler generates a broken `shell` target for any `pyproject.toml` discovered in a **subdirectory**. The generated target sets:

- `workdir: <subdir>` — the `uv lock` command runs here and succeeds, and
- a `changedif` (`file/checksum`) tracking `<subdir>/uv.lock`.

Because the shell resource resolves `changedif` files **relative to `workdir`**, the path doubles (e.g. `foo/bar/foo/bar/uv.lock`). That path doesn't exist, so the post-command file check reports `Missing files` and the target fails — even though `uv lock` succeeded. Projects at the scan root are unaffected because `workdir` is `.`.

## Fix

Record the lockfile path relative to the target's own `workdir` (just `uv.lock`) instead of joining `workdir` into it. The `workdir` is already emitted into the target, and the shell resource prepends it, so joining it twice was the bug. This also matches how sibling crawlers (`golang`, `npm`, `cargo`) reference their lock files as plain literals.

```diff
-	relLockFile := filepath.Join(workdir, "uv.lock")
-	...
-		LockFile:                   relLockFile,
+		LockFile:                   "uv.lock",
```

## Tests

Added a regression scenario (`Scenario 12`) using a nested subdirectory project (`testdata/subdir_project/foo/bar`) to prove the fix holds at arbitrary depth. With the bug present this scenario fails (`files: - "foo/bar/uv.lock"`); with the fix it passes (`files: - "uv.lock"`, `workdir: 'foo/bar'`). Existing scan-root scenarios are unchanged.

Verified end-to-end against a real subdirectory project: `updatecli apply` runs `uv lock` in the subdir and rewrites `<subdir>/uv.lock` correctly, with no `<subdir>/<subdir>/` ever created.

Fixes #9265

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/plugins/autodiscovery/pyproject/dependencies.go
+++ b/pkg/plugins/autodiscovery/pyproject/dependencies.go
@@ -198,8 +198,6 @@ func (p Pyproject) buildTemplateParams(
 		}
 	}
 
-	relLockFile := filepath.Join(workdir, "uv.lock")
-
 	return manifestTemplateParams{
 		ManifestName:               manifestName,
 		ActionID:                   p.actionID,
@@ -214,7 +212,7 @@ func (p Pyproject) buildTemplateParams(
 		TargetName:                 fmt.Sprintf("deps(pypi): bump %q to {{ source %q }}", dep.Name, dep.Name),
 		ScmID:                      p.scmID,
 		UvEnabled:                  lockSupport.uv,
-		LockFile:                   relLockFile,
+		LockFile:                   "uv.lock",
 		Workdir:                    workdir,
 	}
 }
--- a/pkg/plugins/autodiscovery/pyproject/main_test.go
+++ b/pkg/plugins/autodiscovery/pyproject/main_test.go
@@ -547,6 +547,39 @@ targets:
 `,
 			},
 		},
+		{
+			name:        "Scenario 12 -- nested subdirectory project tracks uv.lock relative to its workdir",
+			rootDir:     "testdata/subdir_project",
+			uvAvailable: true,
+			expectedPipelines: []string{
+				`name: 'deps(pypi): bump "requests" for "nested-project" project'
+sources:
+  requests:
+    name: 'Get latest "requests" package version'
+    kind: 'pypi'
+    spec:
+      name: 'requests'
+      versionfilter:
+        kind: 'pep440'
+        pattern: '>=2.28'
+targets:
+  requests:
+    name: 'deps(pypi): bump "requests" to {{ source "requests" }}'
+    kind: 'shell'
+    spec:
+      command: 'uv lock --upgrade-package requests=={{ source "requests" }}'
+      changedif:
+        kind: file/checksum
+        spec:
+          files:
+            - "uv.lock"
+      environments:
+        - name: PATH
+      workdir: 'foo/bar'
+    disablesourceinput: true
+`,
+			},
+		},
 	}
 
 	for _, tt := range testdata {
--- a/pkg/plugins/autodiscovery/pyproject/testdata/subdir_project/foo/bar/pyproject.toml
+++ b/pkg/plugins/autodiscovery/pyproject/testdata/subdir_project/foo/bar/pyproject.toml
@@ -1,0 +1,6 @@
+[project]
+name = "nested-project"
+requires-python = ">=3.12"
+dependencies = [
+    "requests>=2.28",
+]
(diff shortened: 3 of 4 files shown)


loispostula (approved):
thanks @nguyeda for your first contribution!

Lgtm

olblak (approved):
Thank for the pullrequest

PR #9385, opened 2026-07-07
Fix YAML quoted scalar reads

### LINKED ISSUE #4510: YAML parsing seems to erroneously distinguish quoted and non-quoted version strings

### Current Behavior

For example, if I have a YAML that looks like:

```yaml
version: "4.3.1"
```

If I run an ``updatecli`` manifest like:

```yaml
name: foo
sources:
  current:
    kind: "yaml"
    spec:
      file:
        file: "path/to/file"
        key: "$.version"
    transformers:
    - semverinc: "minor"
```

I get an error ``wrong semantic version input: "\"4.3.1\""``

### Expected Behavior

Every YAML parser I've used treats ``"4.3.1"`` and ``4.3.1`` as strings, so it's not clear what distinction ``updatecli`` is making. 

### Steps To Reproduce

See the example in the 'current behavior' description

### Environment

```markdown
- OS: Rocky Linux release 9.5. (Blue Onyx) -- a RHEL9 clone

- updatecli:


Application: 0.95.0
Golang:      1.24.1 linux/amd64
Build Time:  2025-03-11T16:30:40Z
```

### Pipeline Graph

```mermaid
graph TD
    source#current(["current (yaml)"])
```

### Anything else?

This issue may be similar to what I posted in [this discussion](https://github.com/orgs/updatecli/discussions/4491), but with this config I'm not using any templating.

ahfoysal:
Fix #4510

## Test

```shell
go test ./pkg/plugins/resources/yaml
make test-short
```

## Additional Information

This keeps YAML source and condition reads from returning quote syntax for quoted scalar values.


olblak:
Thank for the pullrequest, I'll try to find some times in the coming days to review it

olblak (approved):
Thank you for the fix, it seems to be working as expected

PR #9590, opened 2026-07-14
feat: add opt-in singleBranch option to speed up git scm checkout

kuisathaverat:
## Problem

Updatecli pipelines targeting repositories with a very large number of branches, tags, and other refs (hundreds of branches/tags, 350k+ commits) can spend a very long time (measured up to **~19 minutes**) just checking out the git scm, even when `depth` is set to a low value.

Two issues were found in the shared native git implementation used by every git-based scm (`github`, `gitlab`, `gitea`, `bitbucket`, `stash`, `azuredevops`, `git`) — `pkg/plugins/utils/gitgeneric` (`GoGit`, built on `go-git`):

1. **The initial clone fetches every branch on the remote.** `git.CloneOptions` is built without `SingleBranch`/`ReferenceName`, so go-git falls back to its default refspec `+refs/heads/*:refs/remotes/<remote>/*` (all branches) — `depth` only limits the commit history *per ref*, not the number of refs fetched.
2. **A forced, unconditional "mirror" fetch runs after every clone**, using `RefSpecs: []config.RefSpec{"refs/*:refs/*"}`. This matches **every ref namespace** on the remote — not just branches and tags, but also `refs/pull/*`, `refs/notes/*`, and any other custom ref — on **every single Updatecli execution**, whether the local clone was fresh or reused.

For a repository with hundreds of branches/tags and a long history, this second point in particular is extremely expensive and explains slow checkouts even with `depth` configured.

## Solution

This PR adds a new **opt-in** `singleBranch` option to every git-based scm's spec (`github`, `gitlab`, `gitea`, `bitbucket`, `stash`, `azuredevops`, `git`), following the same pattern as the existing `submodules`/`depth` options.

When `singleBranch: true` is set:
- The initial clone only fetches the configured `branch` (`SingleBranch: true` + `ReferenceName` on `git.CloneOptions`), instead of every branch on the remote.
- The post-clone `refs/*:refs/*` reconciliation fetch is **skipped entirely**.

When the option is left unset (default), behavior is **100% unchanged** — this is purely opt-in.

### Trade-off

Skipping the full ref reconciliation fetch means Updatecli may, in some edge cases, fail to detect an already-published working branch and create a duplicate pull request. This is documented on the new option and is an acceptable trade-off compared to multi-minute checkouts on very large repositories.

### Example

```yaml
scms:
  default:
    kind: github
    spec:
      owner: updatecli
      repository: updatecli
      branch: main
      depth: 1
      singleBranch: true
```

## Impact

Measured on a repository with hundreds of branches/tags and ~350k commits, enabling `singleBranch` reduced the scm checkout time from **~19 minutes to ~43 seconds**.

## Validation

- `go build ./...`
- `go vet ./pkg/plugins/scms/... ./pkg/plugins/utils/gitgeneric/...`
- `go test ./pkg/plugins/scms/... ./pkg/plugins/utils/gitgeneric/...` — all pass (one pre-existing, unrelated failure in `TestGoGit_RemoteURLs` confirmed present on `main` as well, due to the local worktree environment)


olblak (dismissed):
Thank you for the great pull request, covering all scm provider plus the documentation.
I assume you tested it so I am happy to merge as is

kuisathaverat:
I made a test with the repository that hit the issue, we moved from 19 minutes to process the pipeline to 43 seconds

pkg/plugins/resources/shell/changeIf.go
kuisathaverat, line 1: This fixes the check Spell workflow, but the changes are not related to my Pr so I will revert them

pkg/plugins/resources/shell/main.go
kuisathaverat, line 1: This fixes the check Spell workflow, but the changes are not related to my Pr so I will revert them

kuisathaverat:
There are some issues with the Check Spell, but they are in files that I did not touch, so I will not fix them. As a reference, this is the fix https://github.com/updatecli/updatecli/pull/9590/changes/6183807d07cac1b58c191b9a5fdad15d6f90ef8d. I will revert now.

olblak:
It now complains because of this pullrequest that I merged this morning https://github.com/updatecli/updatecli/pull/9565/checks

I can fix it in a different pr

olblak (approved):

PR #9609, opened 2026-07-16
feat(gitcommit): add Git commit source resource

### LINKED ISSUE #2052: Feature Request: Git Commit Source

We build and push container images on each commit to specific Git branches. I attempted to use the `dockerimage` source to update a values file and deploy these images after creation but could not reliably get the most recent tag from a Dockerhub repository due to the tag list not being in the order of the most recent push.

### Solution you'd like

A new `git commit` source that returns the latest commit on a git branch. 

- Using dockerimage source 
- Adding a tag step to our workflow for each commit and using the git tag source

### Anything else?

_No response_

ihopenre-eng:
Fix #2052

Add a `gitcommit` resource. The source returns the latest commit hash for a Git branch; the condition checks that a commit exists in a repository. Both support repository URLs, local paths, and SCM working directories. Branch lookup resolves local and `origin` refs without checking out the branch or modifying the working tree.

The condition hash defaults to the source output and can be overridden with `spec.hash`; abbreviated hashes are resolved with the Git revision syntax.

The resource is registered in the pipeline mapping and includes an end-to-end manifest covering URL, path, and SCM inputs plus two condition scenarios.

## Test

```shell
go test ./pkg/plugins/resources/gitcommit -count=1
go test ./pkg/plugins/utils/gitgeneric -run '^TestGetCommitHash$|^TestIsCommitExist$' -count=1
go test ./pkg/core/pipeline/resource -count=1
go vet ./pkg/plugins/resources/gitcommit ./pkg/plugins/utils/gitgeneric ./pkg/core/pipeline/resource
```

### Checklist

- [x] Documentation companion: [updatecli/website#3563](https://github.com/updatecli/website/pull/3563)

### Tradeoff

The target stage returns an explicit unsupported error because a commit hash is immutable. The condition checks commit existence, defaulting to the source output with `spec.hash` as an override.

### Potential improvement

A future changelog implementation could expose commits between two source revisions.


olblak:
Thanks for the pullrequest, I'll try to find some time in the coming days to review it

pkg/plugins/resources/gitcommit/condition.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/updatecli/updatecli/pkg/core/pipeline/scm"
+)
+
+// Condition is not supported for the Git Commit resource.
+func (gc *GitCommit) Condition(_ context.Context, source string, scm scm.ScmHandler) (bool, string, error) {
+	return false, "", fmt.Errorf("condition not supported for the Git Commit resource")
```
olblak: I think it would be valuable to have a git commit condition to see if the commit exist for a specific repository.

pkg/plugins/resources/gitcommit/source.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	hash, err := gc.nativeGitHandler.GetCommitHash(gc.directory, gc.spec.Branch)
+	if err != nil {
+		return fmt.Errorf("retrieving latest Git commit: %w", err)
+	}
+
+	branch := gc.spec.Branch
+	if branch == "" {
+		branch = "HEAD"
```
olblak, line 35: Interesting I am wondering if it's a way to have git repositories that either use master or main as default branch

olblak:
I just gave a quick test and it works as expected thanks for the e2e tests.
I am happy to merge it as it is or wait for a condition if you are interested in it

pkg/plugins/resources/gitcommit/condition.go
ihopenre-eng: Good idea — implemented in 9b60f1d. The condition checks that the commit exists in the selected repository: the commit hash defaults to the source output and can be overridden with `spec.hash` (abbreviated hashes are supported, resolved via the Git revision syntax). Repository selection follows the same order as the source: `spec.url`, then `spec.path`, then the `scmid` working directory. A missing commit reports "git commit not found" while lookup failures (e.g. an unreadable repository) surface as errors.

pkg/plugins/resources/gitcommit/source.go
ihopenre-eng: Yes, exactly. When `spec.branch` is omitted, the source resolves the repository's `HEAD`, so it follows whatever default branch the repository uses (`main`, `master`, or anything else) without hardcoding a name. When `spec.branch` is set, the lookup tries the local branch first and then falls back to the `origin` remote-tracking branch.

ihopenre-eng:
Thanks for testing it! I went ahead and added the condition in this PR (9b60f1d).

The condition checks that a commit exists in the repository: the commit hash defaults to the source output and can be overridden with `spec.hash` (abbreviated hashes are supported). Repository selection follows the same order as the source: `spec.url`, then `spec.path`, then the `scmid` working directory.

The e2e manifest now covers two condition scenarios — checking the source commit against the SCM working directory, and checking a pinned hash against a cloned URL — and the website documentation PR (updatecli/website#3563) is updated accordingly.

```
go test ./pkg/plugins/resources/gitcommit -count=1
go test ./pkg/plugins/utils/gitgeneric -run '^TestIsCommitExist$' -count=1
go run . diff --config e2e/updatecli.d/success.d/gitCommit.yaml
```

olblak (approved):
Thanks for the pull request

PR #9611, opened 2026-07-16
fix(terraform): support direct OpenTofu registry h1 hashes fetch (#9213)

### LINKED ISSUE #9213: terraform/lock: only requested-platform h1: hashes are recorded, so tofu init rewrites the lock file (OpenTofu registry exposes all platforms)
> Related: #8968 (same crawler, same OpenTofu-registry mismatch family — distinct bug)

### Summary

The `terraform/lock` resource (and the `terraform` autodiscovery crawler) records `h1:` hashes only for the platforms listed in `spec.platforms`, while recording `zh:` hashes for *all* published platforms. When the target registry is **OpenTofu** (`registry.opentofu.org`), a subsequent `tofu init` rewrites the lock file to add `h1:` hashes for **every** published platform — producing a spurious diff on every fresh `tofu init` after pulling a branch UpdateCli updated.

UpdateCli and `tofu init` therefore never agree on the lock file, so it is perpetually "dirty" for anyone running `tofu init` locally or in CI.

### Steps to reproduce

1. Configure the terraform autodiscovery crawler with a platform subset (the documented default `linux_amd64, linux_arm64, darwin_amd64, darwin_arm64` reproduces it equally):
   ```yaml
   autodiscovery:
     crawlers:
       terraform:
         platforms:
           - linux_arm64
           - darwin_arm64
   ```
2. Let UpdateCli bump a provider (e.g. `hashicorp/aws`) and commit the resulting `.terraform.lock.hcl`. The `aws` block has **2** `h1:` hashes and **15** `zh:` hashes.
3. Run `tofu init` (OpenTofu, against `registry.opentofu.org`) in that module.
4. `.terraform.lock.hcl` is modified: the `aws` block now has **15** `h1:` hashes (the 2 originals kept, 13 added).

### Expected vs. actual

- **Expected:** a lock file produced by UpdateCli is not modified by a subsequent `tofu init`.
- **Actual:** `tofu init` always expands the `h1:` set to the full published platform list.

### Root cause

`tofu init` against the OpenTofu registry records `h1:` for **all** published platforms, deterministically and machine-independently. Clean-room proof (empty plugin cache, fresh `HOME`), using `null` which publishes 10 platforms:

```
$ tofu init -backend=false      # OpenTofu v1.12, single platform, clean machine
=> .terraform.lock.hcl contains 10 h1: hashes (all 10 published platforms), not 1
```

UpdateCli only computes `h1:` for the requested platforms:

- `pkg/plugins/resources/terraform/lock/main.go` — `getProviderHashes` passes `spec.Platforms` straight through:
  `t.lockIndex.GetOrCreateProviderVersion(ctx, provider, version, t.spec.Platforms)` → `pv.AllHashes()`
- the underlying `minamijoyo/tfupdate` (`lock/index.go`) loops over only those platforms, with a now-outdated assumption:
  ```go
  for _, platform := range platforms {
      ...
      // Currently the Terraform Registry returns the zh hash for all platforms,
      // but not the h1 hash, so the h1 hash has to be calculated separately.
  }
  ```
  This holds for the legacy Terraform registry, but the **OpenTofu registry does expose `h1` for all platforms**, so for OpenTofu the full `h1:` set is available without downloading every platform.

Net result: `zh:` = all platforms (from signed `SHA256SUMS`), `h1:` = requested platforms only → permanent mismatch with what `tofu init` writes.

### Proposed fixes (any one resolves it)

1. **Record the registry-provided `h1` set for OpenTofu registries** — match `tofu init` exactly, no extra downloads. (Preferred.)
2. **Support an "all platforms" option** for `spec.platforms`, so the lock can be complete without hand-enumerating platforms (which differ per provider and drift over time).
3. **Document the limitation** and recommend regenerating with `tofu init` as a post-step.

### Environment

- UpdateCli: `v0.118.0`
- OpenTofu: `v1.12.x`
- Registry: `registry.opentofu.org`
- Resource: `terraform/lock` via terraform autodiscovery

prajaktaukirde:
Fix #9213

This PR fixes the issue where running UpdateCli against the OpenTofu provider registry (`registry.opentofu.org`) only records the requested platform's `h1` hashes. Because the OpenTofu registry serves `h1` hashes for all published platforms directly in its download metadata response (under the `packages` dictionary), a subsequent local `tofu init` expands the lock file's `h1` set to match the full platform list, leading to a perpetual mismatch and a "dirty" lock file.

To resolve this, we:
- Intercept lock queries targeting `registry.opentofu.org`.
- Fetch the provider versions metadata and pick a supported platform.
- Query download metadata for that platform, which yields the `packages` dictionary containing both `h1` and `zh` hashes for **all** published platforms.
- Sort, deduplicate, and record these hashes to match `tofu init`'s behavior exactly, avoiding any binary downloads.
- Fall back to the default `tfupdate` downloader behaviour in case of failure.

## Test

```shell
cd pkg/plugins/resources/terraform/lock
go test -v .

### CODE REVIEW

File: pkg/plugins/resources/terraform/lock/main.go [hunk 1]
```diff
@@ -195,6 +222,78 @@ func (t *TerraformLock) getProviderHashes(version string) ([]string, error) {
...
 }
 
+func (t *TerraformLock) getOpenTofuProviderHashes(version string) ([]string, error) {
+	scheme := "https"
+	if strings.HasPrefix(t.provider.Hostname, "127.0.0.1") || strings.HasPrefix(t.provider.Hostname, "localhost") {
+		scheme = "http"
+	}
+
+	urlVersions := fmt.Sprintf("%s://%s/v1/providers/%s/%s/versions", scheme, t.provider.Hostname, t.provider.Namespace, t.provider.Type)
+	req, err := http.NewRequestWithContext(context.Background(), "GET", urlVersions, nil)
```
olblak, line 235: Could I ask you to use the Updatecli custom httpclient from https://github.com/updatecli/updatecli/blob/main/pkg/core/httpclient/main.go

It will setup a few things like retry, user agent, etc.

olblak (changes requested):

pkg/plugins/resources/terraform/lock/main.go
prajaktaukirde: Hi @olblak, thank you for the feedback! 

I have updated the implementation to import and use the custom Updatecli HTTP client (`httpclient.NewRetryClient()`) for querying the OpenTofu registry. The updated commit has been pushed to the branch.

loispostula (changes requested):
Built and tested this locally against registry.opentofu.org/hashicorp/random 3.9.0 (15
published platforms). The approach works: before, the lock has 1 h1 and `tofu init`
adds the other 14; with this branch updatecli writes all 15 and `tofu init` leaves them
alone. Fixes #9213.

**1. Doesn't compile.** `Provider.Hostname` is `svchost.Hostname`, not `string`:

    main.go:230:23: cannot use t.provider.Hostname (variable of string type
    svchost.Hostname) as string value in argument to strings.HasPrefix

Needs a `string(...)` conversion. CI never ran here (fork PR), hence not caught.

**2. Empty hashes return as success.** If `packages` is empty, the function returns
`(nil, nil)`, the caller only falls back on `err != nil`, and `Apply` writes
`hashes = []`, emptying the block. Needs a `len(uniqueHashes) == 0` guard.

**3. `spec.Platforms` is silently ignored.** Ran with `platforms: [darwin_arm64]`, got
all 15. Probably the right lock file, but the field is documented as controlling this
and `ReportConfig` still reports it. Needs a log line and a docs update.

Minor:

- The localhost http downgrade (228-231) is unreachable in production, the only caller gates on the opentofu hostname. An unexported `registryBaseURL` field set in `New()` would let the test inject instead, and makes `httpClient` mockable.
- Hostname is hardcoded, so self-hosted registries get nothing.
- `v.Platforms[0]` is arbitrary, prefer an entry from `spec.Platforms`.
- `http.MethodGet`, and `slices.Sort`/`slices.Compact` over `sort.Strings` plus the manual dedup.

Tests cover the happy path only, and call the unexported function directly, so the routing condition in `getProviderHashes` is never exercised. Worth adding the hostname gate, non-200, and empty `packages`.

olblak:
@prajaktaukirde Without pressure, would you have some time to look at @loispostula comment?

PR #9688, opened 2026-07-19
fix: guard against short paths in changelog GitHub URL redirect

AruneshDwivedi:
redirectToGitHubRawContent splits the changelog URL path on "/" and reads index 3 without checking the length, so a github.com changelog URL with fewer than four path segments panics with index out of range. Since the URL comes from the org.opencontainers.image.changelog image annotation, a short value like https://github.com/CHANGELOG.md crashes the run.

Split once, check the slice has more than three elements before touching index 3, and leave short paths unchanged.

### DIFF
--- a/pkg/plugins/resources/dockerimage/changelog.go
+++ b/pkg/plugins/resources/dockerimage/changelog.go
@@ -112,8 +112,8 @@ func getChangelogAnnotation(desc v1.Descriptor) string {
 // redirectToGitHubRawContent tries to redirect a github url to its associated file raw content
 func redirectToGitHubRawContent(u *url.URL) {
 	beforePath := u.Path
-	if strings.Split(u.Path, "/")[3] == "tree" {
-		s := strings.Split(u.Path, "/")
+	s := strings.Split(u.Path, "/")
+	if len(s) > 3 && s[3] == "tree" {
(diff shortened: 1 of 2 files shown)


olblak:
Hi @AruneshDwivedi,
I am not familiar with those short changelog url.
Do you have an example of a failing pipeline due to this?

AruneshDwivedi:
@olblak good question it's a real panic, not a theoretical one.
An image annotated with `org.opencontainers.image.changelog: https://github.com/CHANGELOG.md` (short URL, changelog at repo root) hits `redirectToGitHubRawContent`, which does `strings.Split(u.Path, "/")[3]`. That path only has two segments, so it panics with `index out of range [3] with length 2` and kills the whole run. Any pipeline doing dockerimage changelog autodiscovery over such an image crashes.  put up a fix in #9689 with a test that repros it. The example in my earlier message used a 4-segment URL by mistake  that one doesn't panic; the 2-segment `github.com/CHANGELOG.md` is the actual trigger.

olblak:
Thanks for the explanation, why did you created two differernt pullrequest with the same content 
#9688 vs #9689

olblak (approved):
Thanks for the pullrequest

PR #9699, opened 2026-07-22
chore: cleanup autodiscovery spec comment

olblak:
This pullrequest is an attempt to fix the comment used to generate the jsonschame then used by the documentation website.
Especially `pkg/plugins/autodiscovery/golang/` and `pkg/plugins/autodiscovery/dockerfile/` seems problematic
Hugo doesn't seem to like single and double quotes


olblak:
No need to run all the tests as this pr only modify code comments

PR #9705, opened 2026-07-23
fix(version): use original version string instead of re-formatted date for lookup

AruneshDwivedi:
Search formats parsed time values back to strings using the layout, then uses those formatted strings as map keys into rawversions. The map is keyed by the original version strings, so after formatting loses precision or changes format, the lookup silently returns an empty string.

Track the original version string alongside each parsed time during Init, so Search can look up the correct key in rawversions.

### DIFF
--- a/pkg/plugins/utils/version/time.go
+++ b/pkg/plugins/utils/version/time.go
@@ -21,6 +21,7 @@ type Time struct {
 	// Day of the month: "2" "_2" "02"
 	layout       string
 	versions     []time.Time
+	originals    []string
 	FoundVersion Version
 }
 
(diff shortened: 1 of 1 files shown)


olblak:
Thanks for the contribution, could I ask you to follow the pulln request template and explain how to reproduce this issue with a release Updatecli manifest? It's meant to help me review the change without spending too much time on each pr because now I quite often receive code contribution generated by AI that do not always make sense.
Especially that here it doesn't contains test

olblak:
I am closing this pr as no activity for more than a month. Feel free to reopen

PR #9706, opened 2026-07-23
fix: add username/password auth to helm/argocd autodiscovery

itsteddyyo:
Hi! First of all, thanks for creating this great tool! 😃 It has become a big cornerstone of my update strategy.

While using it, I stumbled across an issue with some Autodiscovery plugins.

Both the Helm and ArgoCD Autodiscovery documentation mention support for `username/password` authentication in the auth section. However, it looks like only `token` is currently passed through by the code. This pull request adds support for passing both to the generated manifests as well.

I also fixed a small documentation issue where the ArgoCD auth example used `<registry>: my-token` instead of `<registry>.token: my-token` as I was already adding the other options anyway 😄 

## Test

```shell
go test ./pkg/plugins/autodiscovery/helm
go test ./pkg/plugins/autodiscovery/argocd
```

I didn't add any additional tests because passing the `token` field through to the generated manifests was already covered by existing tests, and the new functionality follows the same code path. Adding separate tests for `username/password` felt somewhat redundant, but I'm happy to add them if you think they would be valuable. 🙂

### Checklist

- [ ] <s>I have updated the documentation via pull request in [website](https://github.com/updatecli/website) repository.</s> Not applicable

### Tradeoff

None that i can think of.

### Potential improvement

First of all, it's absolutely possible that I misunderstood part of the code. I only looked at a small portion of the codebase while investigating this issue, so apologies if that's the case. 🙂

I think it could be worth separating the Helm Autodiscovery auth section/handling  for Helm registries and container registries a bit more. The current implementation uses `docker.InlineKeyChain` for both container registries and Helm registries. That makes sense for OCI registries, but it feels a little less intuitive for non-OCI Helm registries (see the [Helm autodiscovery auth handling](https://github.com/updatecli/updatecli/blob/main/pkg/plugins/autodiscovery/helm/main.go#L30), which is used for both container manifests and Helm dependency manifests).

I can also imagine situations where different credentials are needed for container images and Helm charts. One example is GitLab, which provides both a Container Registry and a Package Registry on the same domain. They support fine-grained credentials, so you might want separate credentials with access to only one of those registries.

Additionally, the type itself appears to be primarily intended for OCI registries (see the [`InlineKeyChain` type documentation](https://pkg.go.dev/github.com/updatecli/updatecli/pkg/plugins/utils/docker#InlineKeyChain)). At the same time, it seems to be used for non-OCI registries as well.

I don't think this is a major issue - the `InlineKeyChain` type works well as a catch-all since it supports `username`, `password`, and `token`. It just felt a little confusing while I was tracking down this bug.

### DIFF
--- a/pkg/plugins/autodiscovery/argocd/application.go
+++ b/pkg/plugins/autodiscovery/argocd/application.go
@@ -216,13 +216,17 @@ func (f ArgoCD) generateManifestBySource(data ApplicationSourceSpec, file string
 		}
 	}
 
+	username := ""
+	password := ""
 	token := ""
 	repoURL, err := url.Parse(data.RepoURL) // to validate URL format
 	switch err {
 	case nil:
 		if _, ok := f.spec.Auths[repoURL.Host]; ok {
+			username = f.spec.Auths[repoURL.Host].Username
+			password = f.spec.Auths[repoURL.Host].Password
 			token = f.spec.Auths[repoURL.Host].Token
-			logrus.Debugf("found token for repository %q", data.RepoURL)
+			logrus.Debugf("found credentials for repository %q", data.RepoURL)
 		}
 	default:
 		logrus.Debugf("Ignoring auth configuration due to invalid Helm repository URL: %s", err)
@@ -258,6 +262,8 @@ func (f ArgoCD) generateManifestBySource(data ApplicationSourceSpec, file string
 		TargetID                   string
 		TargetKey                  string
 		TargetYamlDocument         int
+		Username                   string
+		Password                   string
 		Token                      string
 		File                       string
 		ScmID                      string
@@ -280,6 +286,8 @@ func (f ArgoCD) generateManifestBySource(data ApplicationSourceSpec, file string
 		TargetYamlDocument:         yamlDocument,
 		File:                       file,
 		ScmID:                      f.scmID,
+		Username:                   username,
+		Password:                   password,
 		Token:                      token,
 	}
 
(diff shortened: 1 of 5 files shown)


olblak:
Thank you for the pullrequets and the nice words, I'll try to review over the weekend

olblak:
@itsteddyyo Could you fix the linting issue? 
Have you tested this pullrequest?

itsteddyyo:
@olblak Sorry my bad! I did change some variable names and did not format properly 😄 Should be fixed now! Yeah i tested the generated binary locally and it worked as expected. I used both autodiscovery plugins but for the helm plugin i used `ignorecontainer: true`. But the changes should not affect that anyway. Thanks for taking a look! 😄

olblak:
> I think it could be worth separating the Helm Autodiscovery auth section/handling for Helm registries and container registries a bit more. The current implementation uses docker.InlineKeyChain for both container registries and Helm registries. That makes sense for OCI registries, but it feels a little less intuitive for non-OCI Helm registries (see the Helm autodiscovery auth handling, which is used for both container manifests and Helm dependency manifests).I think it could be worth separating the Helm Autodiscovery auth section/handling for Helm registries and container registries a bit more. The current implementation uses docker.InlineKeyChain for both container registries and Helm registries. That makes sense for OCI registries, but it feels a little less intuitive for non-OCI Helm registries (see the Helm autodiscovery auth handling, which is used for both container manifests and Helm dependency manifests).

Yes I agree, it's a good observation.
It's one of those feature that evolve organically because initially we needed Docker credential which were username/password, then we added token support when Dockerhub became an OCI compliant registry.
When we realize that it could be reuse for any OCI compliant registry like helm we didn't refactor at that time and instead used it as is.

olblak (approved):
Thanks for the pullrequest

PR #9970, opened 2026-08-14
fix: hash the full source spec for the cache key

### LINKED ISSUE #9849: Source cache key is incomplete for 11 plugins whose ReportConfig omits result-affecting spec fields
`cache.Key()` (`pkg/core/cache/source.go:69`) hashes `resource.ReportConfig()`. The cache is created unconditionally in `pkg/core/engine/run.go:32` with no opt-out, so any spec field a plugin leaves out of `ReportConfig` is invisible to the cache key. Two distinct sources then share a key and the second silently receives the first one's value.

`ReportConfig` predates the cache (#4898, report export), where omitting a field was harmless.

#9821 is one instance of this (`temurin` omitted `specificversion`). Auditing all 43 implementations found 11 more.

Method: AST diff of each `Spec` against the fields assigned in its `ReportConfig`, then manual verification that the field reaches the `Source()` path.

| Plugin | Missing | Collision |
|---|---|---|
| `shell` | `Shell`, `WorkDir`, `Environments` | Same `command` in two `workdir`s: second source gets the first's stdout |
| `updateclihttp` | `Request` (`Verb`, `Body`, `Headers`, `NoFollowRedirects`) | Same `url`, different POST body or verb (`main.go:62-80`) |
| `yaml` | `DocumentIndex` | Same file and key, different document (`source.go:91`) |
| `csv` | `Comma`, `Comment`, `VersionFilter` | Different separator or version filter |
| `maven` | `VersionFilter` | Same artifact, different filter |
| `cargopackage` | `VersionFilter` | Same |
| `json` | `Engine` | dasel v1/v2/v3 differ in query semantics (`source.go:37+`) |
| `awsami` | `SortBy` | Newest vs oldest AMI from identical filters (`helpers.go:37`) |
| `go/gomod` | `Indirect`, `Replace`, `ReplaceVersion` | Different module version from the same `go.mod` (`version.go:48-134`) |
| `gittag` | `LsRemote`, `Depth` | `lsremote` bypasses the clone; `depth` truncates the tag list |
| `gitbranch` | `Depth` | Clone depth changes the branch list |

Verified as not bugs: credentials (intentional, documented at `pkg/core/cache/source.go:46`); target/condition-only fields (`file.ForceCreate`, `yaml.Comment`, `gittag.Message`, release `Title`/`Commitish`/`Description`/`Draft`/`Prerelease`, `updateclihttp.ResponseAsserts`); `json.Multiple` and `csv.Multiple` (deprecated, normalized into the reported `Query` at `New()`).

## Decision needed

Nine are a one-line addition each. Two are not: `shell.Environments` and `updateclihttp.Request.Headers` routinely carry secrets, and `ReportConfig` now serves two conflicting contracts, redaction for display and completeness for keying. Redacting two distinct values to the same string reintroduces the collision.

1. Add the 9 safe fields, leave `shell` and `updateclihttp`.
2. Split the concerns: a separate cache-key method on the resource interface, `ReportConfig` untouched for display. Touches every plugin, removes the bug class.
3. One method, but hash secret-bearing fields instead of redacting them, so distinct values stay distinct without appearing in the report.

Option 2 makes the contract explicit, which matters because the failure mode is silent: no error, no warning, just a wrong value written to the target.

@olblak preference?

gangadhar-res:
## What

The source cache key (`pkg/core/cache/source.go`) now hashes the full user-provided spec from `ResourceConfig` instead of `resource.ReportConfig()`.

## Why

`ReportConfig()` exists to redact resource configuration for reports and predates the cache (#4898). Eleven plugins omit result-affecting spec fields from it (e.g. `shell` omits `workdir`, `yaml` omits `documentindex`), so two distinct sources could silently share a cache entry and the second received the first one's value — no error, no warning, just a wrong value written to the target.

This implements option 2 from the issue discussion: the cache key gets its own complete input, `ReportConfig` keeps its original reporting purpose, and no per-plugin changes are needed — the whole bug class is removed at the one call site.

Notes:

- Secret-bearing spec fields are safe to include: only the SHA256 digest is ever stored or logged, so distinct secrets stay distinct without appearing anywhere.
- Hashing stays deterministic: specs decode via yaml.v3 into `map[string]interface{}` and `encoding/json` marshals maps with sorted keys.
- The resource is still instantiated first, so unknown kinds or malformed specs keep returning the empty-key cache-miss sentinel.

## Testing

- New regression test: two `shell` sources differing only in `workdir` (a field `ReportConfig` omits) get different keys, while identical specs still share one — both as typed specs and as YAML-decoded map specs, plus hash stability across decodes. Both new tests fail on `main` by construction.
- `go test ./pkg/core/cache/` passes; `go build ./...`, `gofmt`, `go vet`, and `golangci-lint` introduce no new findings.

Fixes #9849


gangadhar-res:
@olblak Could I have a review on this pr ? thanks

olblak:
> @olblak Could I have a review on this pr ? thanks 

Thanks for the pull request but I can't review it at the moment

pkg/core/cache/source.go
```diff
@@ -66,16 +66,25 @@ type cacheKeyInput struct {
...
+	// spec is well-formed before caching anything under this config.
+	if _, err := resource.New(rc); err != nil {
 		logrus.Debugf("source cache: failed to instantiate resource for key: %v", err)
 		return ""
 	}
 
 	data, err := json.Marshal(cacheKeyInput{
 		Kind: rc.Kind,
-		Spec: r.ReportConfig(),
+		Spec: rc.Spec,
```
olblak, line 81: I think it's ok to use the full spec as it's going to be sha256 encoded anyway

pkg/core/cache/source.go
```diff
@@ -66,16 +66,25 @@ type cacheKeyInput struct {
 // the omitempty tag on cacheKeyInput.SCM keeps the hash identical to the
 // pre-SCM form in that case. Returns empty string when hashing fails; callers
 // treat that as a cache miss.
+//
+// The key hashes the full user-provided spec, not resource.ReportConfig():
```
olblak: Could I ask you to clean up the comment? 
It doesn't make sense to document the old behavior.
Those comments are generated by your AI tool to help you understand what it did

gangadhar-res: Fair point, cleaned that up. Trimmed it to what the function does now instead of how it used to work.

PR #9759, opened 2026-07-31
fix(autodiscovery/updatecli): correctly use relative file path

olblak:
## Test

To test this pull request, you can run the following commands:

```shell
cd pkg/plugins/resources/file/
go test
```

### Checklist

- [ ]  I have updated the documentation via pull request in [website](https://github.com/updatecli/website) repository.

PR #10109, opened 2026-08-25
fix(autodiscovery/helmfile): use relative file path in generated manifests

### LINKED ISSUE #10108: Helmfile autodiscovery: condition and target fail with "absolute path not allowed" when scmid is set
## Summary

The Helmfile autodiscovery plugin generates conditions and targets using an
absolute file path, which causes every generated pipeline to fail with:

    init yaml files: absolute path "/tmp/updatecli/.../helmfile.yaml" is not allowed:
    files must stay within the working directory "/tmp/updatecli/..."

## Version

v0.120.1

## Steps to reproduce

1. Configure a `helmfile` crawler in autodiscovery with an `scmid`.
2. Run `updatecli diff`.
3. All helmfile-derived pipelines fail at the condition step with the above error.
   The target is skipped due to `dependsOn conditions`.

## Root cause

In `pkg/plugins/autodiscovery/helmfile/releases.go`, `relativeFoundChartFile`
is correctly computed on line 67 but the `File` field on line 229 is assigned
`foundHelmfile` (the absolute path) instead:

```go
// line 67 — relative path is computed
relativeFoundChartFile, err := filepath.Rel(h.rootDir, foundHelmfile)

// line 229 — absolute path is used instead
File: foundHelmfile, // ← should be relativeFoundChartFile
```

The `ManifestName` and `ConditionName` fields on lines 213 and 217 already use
`relativeFoundChartFile` correctly — `File` is the only field that doesn't.

## Fix

```go
File: relativeFoundChartFile,
```

## Related

This is the same bug fixed for the `updatecli` autodiscovery plugin in #9759.
That PR corrected the identical pattern in
`pkg/plugins/autodiscovery/updatecli/policies.go` but the helmfile plugin was
not updated at the same time.

alecharp:
## Summary

Fixes #10108.

The `File` field in the generated condition and target was set to `foundHelmfile` (absolute path within the SCM clone) instead of `relativeFoundChartFile`. This caused every helmfile-derived pipeline to fail at the condition step when an `scmid` is configured.

`ManifestName` and `ConditionName` already used `relativeFoundChartFile` correctly — `File` was the only field that did not.

This is the same bug fixed for the `updatecli` autodiscovery plugin in #9759; the helmfile plugin was not updated at the same time.

## Before

All helmfile-derived condition steps fail with:

```
✗ Something went wrong:
	init yaml files: absolute path "/tmp/updatecli/.../helmfile.yaml" is not allowed:
	files must stay within the working directory "/tmp/updatecli/..."
```

The target is skipped due to `dependsOn conditions`, silently preventing any helmfile release from being updated.

## After

Conditions resolve correctly against the SCM clone and targets apply as expected:

```
✔ key "$.releases[22].chart" is correctly set to "vaultwarden/vaultwarden"
⚠ - change detected:
	* key "$.releases[22].version" updated from "0.44.0" to "0.46.1", in file "helmfile.yaml"
```

Tested locally by building from this branch and running `updatecli pipeline apply` against a real helmfile repository — 6 version-bump PRs were opened correctly where previously 0 were.

## Checklist

- [ ] I have updated the documentation via pull request in [website](https://github.com/updatecli/website) repository.

### DIFF
--- a/pkg/plugins/autodiscovery/helmfile/main_test.go
+++ b/pkg/plugins/autodiscovery/helmfile/main_test.go
@@ -42,7 +42,7 @@ conditions:
     name: 'Ensure release "datadog" is specified for Helmfile "helmfile.d/cik8s.yaml"'
     kind: 'yaml'
     spec:
-      file: 'testdata/helmfile.d/cik8s.yaml'
+      file: 'helmfile.d/cik8s.yaml'
       key: '$.releases[0].chart'
       value: 'datadog/datadog'
     disablesourceinput: true
@@ -51,7 +51,7 @@ targets:
     name: 'deps(helmfile): update "datadog" Helm Chart version to {{ source "datadog"}}'
     kind: 'yaml'
     spec:
-      file: 'testdata/helmfile.d/cik8s.yaml'
+      file: 'helmfile.d/cik8s.yaml'
       key: '$.releases[0].version'
     sourceid: 'datadog'
 `, `name: 'Bump "docker-registry-secrets" Helm Chart version for Helmfile "helmfile.d/cik8s.yaml"'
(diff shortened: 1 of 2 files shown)


olblak:
Thanks for the fix

olblak:
@alecharp Could you fix the broken unit test :)

olblak (approved):

